PR #138, opened 2024-08-02
chore(scripts): Add support for basic OpenShift to hack/demo-env

jgwest:
This PR:
- Adds support to hack/demo-env for running on a basic OpenShift cluster (must have LoadBalancer support)
    - The existing non-OpenShift case should still work as before, hopefully @jannfis can confirm I've not broken it.
- Update scripts to work on MacOS (pwgen/pwmake, sleep, and sed)
- After setup is complete, a simple sanity test runs that verifies that the Pods are running on all 3 contexts.


hack/demo-env/gen-creds.sh
```diff
@@ -7,11 +7,22 @@
...
+# Use pwmake if it exists, otherwise use pwgen
+which pwmake
+if [[ $? == 0 ]]; then
+	set -e
+	pwmake=$(which pwmake)
+
+else
+	set -e
+	pwmake=$(which pwgen)
+fi
```
jannfis: IMHO, this could be simplified by:

```shell
#!/bin/sh
set -eo pipefail
if ! pwmake=$(which pwmake); then
    pwmake=$(which pwgen)
fi
```

An exit code other than 0 in an if statement doesn't trigger the shell's exit on error.

hack/demo-env/setup-vcluster-env.sh
```diff
@@ -1,15 +1,25 @@
...
+# Kubectl context to restore
+initial_context=`kubectl config current-context`
 
 cleanup() {
-	kubectl config set-context ${context}
+	kubectl config use-context ${initial_context}
+}
+
+on_error() {
+	echo "ERROR: Error occured, terminating."
```
jannfis:
```suggestion
	echo "ERROR: Error occured, terminating." >&2
```

hack/demo-env/setup-vcluster-env.sh
```diff
@@ -20,29 +30,176 @@ namespace() {
...
+check_for_openshift() {
+	set +e
+
+	OPENSHIFT_RESOURCES=`kubectl api-resources | grep "openshift.io" | wc -l`
+
+	if [[ "$OPENSHIFT_RESOURCES" != "0" ]]; then
+		OPENSHIFT=true
+	fi
+
+	set -e
```
jannfis: Simplification:

```shell
if kubectl api-resources | grep -q "openshift.io"; then
    OPENSHIFT=true
fi
```

hack/demo-env/setup-vcluster-env.sh
```diff
@@ -1,15 +1,25 @@
...
+
 SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 VCLUSTERS="control-plane:argocd agent-managed:argocd agent-autonomous:argocd"
+VCLUSTERS_AGENTS="agent-managed:argocd agent-autonomous:argocd"
 action="$1"
 
-# Kubectl context to restore to
-context=default
+# Kubectl context to restore
+initial_context=`kubectl config current-context`
```
jannfis: For sake of consistency:

```suggestion
initial_context=$(kubectl config current-context)
```

hack/demo-env/setup-vcluster-env.sh
```diff
@@ -20,29 +30,176 @@ namespace() {
...
+		EXTRA_VCLUSTER_PARAMS="-f ${SCRIPTPATH}/resources/vcluster.yaml"
+	fi
+
 	echo "-> Creating required vclusters"
 	for c in $VCLUSTERS; do
 		cluster=$(cluster $c)
-		echo "  --> Creating vcluster $cluster"
-		vcluster create --context=default -n vcluster-${cluster} --expose --kube-config-context-name vcluster-${cluster} vcluster-${cluster}
+ 		echo "  --> Creating vcluster $cluster"
+		vcluster create --context=${initial_context} ${EXTRA_VCLUSTER_PARAMS} --switch-context=false -n vcluster-${cluster} --expose --kube-config-context-name vcluster-${cluster} vcluster-${cluster}
```
jannfis, line 194: My vcluster version (0.19.6) doesn't support the `--switch-context` flag. Latest is 0.19.7, but changelog doesn't mention any added switch between .6 and .7.

```
-> Creating required vclusters
  --> Creating vcluster control-plane
15:01:00 fatal unknown flag: --switch-context
ERROR: Error occured, terminating.
```

Which vcluster version are you using?

jannfis (commented):
Awesome! Thanks so much.

I have a couple of suggestions and hit an error below.

jannfis:
Confirmed that scripts are still working with vcluster on microk8s after removing `--switch-context=false` parameter

hack/demo-env/setup-vcluster-env.sh
jgwest: ~The vCluster docs imply one should use the beta versions by default: https://www.vcluster.com/docs/vcluster/#deploy-vcluster~

*EDIT*: Ok, this is slightly less true than I stated. While the vCluster doc URL (above) DOES default to version 0.20, 0.19 is listed as stable in the doc version selector. And 0.19 is still getting recent updates.

In any case, I'm happy either way, just let me know whether you'd prefer the vcluster target to be 0.20 or 0.19.

I'm on:
```
[jgw@localhost-lan argocd-agent]$ vcluster -v
vcluster version 0.20.0-beta.12
```

jannfis: Cool, thanks. I think I will upgrade then, I thought that switch might have been from an older version. But if it's actually in the next stable version, I believe we can just leave it there.

hack/demo-env/setup-vcluster-env.sh
jgwest: @jannfis AFAICT `grep -q` as written will doing the opposite: it will set `OPENSHIFT=true` in the non-openshift cases. 

I tried using `grep -v`, and/or looking at whether any other tweaks could be used, but my bash-fu and google-fu did not reveal any obvious solutions. 

If you have a suggestion here, let me know.

jannfis: `-q` only suppresses the output of matched lines, but doesn't change exit codes:

```shell
$ echo "foo" | grep foo
foo
$ echo $?
0
$ echo "foo" | grep bar
$ echo $?
1
$ echo "foo" | grep -q foo
$ echo $?
0
$ echo "foo" | grep -q bar
$ echo $?
1
```

jgwest: @jannfis Correct, the issue is that the if statement requires an error code of 1 to execute what is inside the if block, and there is no obvious way (to the level of my bash scripting ability) to produce an error code of 1 from `grep` in the openshift=true case.

jgwest: Found a better of doing it based on your original suggestion, ready for re-review.

hack/demo-env/gen-creds.sh
jgwest: Pushed the commit containing this change, ready for re-review.

hack/demo-env/setup-vcluster-env.sh
jannfis: In a shell `if` statement, an exit code of 0 is true and anything else is false :)

hack/demo-env/setup-vcluster-env.sh
```diff
@@ -20,29 +30,173 @@ namespace() {
...
+trap cleanup EXIT
+trap on_error ERR
+
+# check_for_openshift looks for cluster OpenShift API Resources, and if found, sets OPENSHIFT=true
+check_for_openshift() {
+
+	OPENSHIFT=true
+	if kubectl api-resources | grep -q "openshift.io"; then 
+		OPENSHIFT=
+	fi
```
jannfis: The above logic breaks things on non-OpenShift clusters:
```shell
$ ./hack/demo-env/setup-vcluster-env.sh create
namespace/vcluster-agent-managed created
namespace/vcluster-control-plane created
namespace/vcluster-agent-autonomous created
error: resource mapping not found for name: "anyuid-seccomp-netbind" namespace: "" from "/home/vagrant/src/argocd-agent/hack/demo-env/resources/scc-anyuid-seccomp-netbind.yaml": no matches for kind "SecurityContextConstraints" in version "security.openshift.io/v1"
ensure CRDs are installed first
ERROR: Error occurred, terminating.
Switched to context "default".
Switched to context "default".

```
```suggestion
	OPENSHIFT=
	if kubectl api-resources | grep -q "openshift.io"; then 
		OPENSHIFT=true
	fi
```

jannfis (approved):
LGTM, thanks!

Issue #137, opened 2024-08-02
Add E2E framework and initial set of simple tests, to run as part of PRs


jgwest:
Argocd-agent currently has a set of detailed unit tests across the code base, which work well for white-box testing of individual functions within the packages of the project. We also have a what looks like a [simple E2E test](https://github.com/jannfis/argocd-agent/blob/main/test/e2e/e2e_test.go)  of principal code (but from quickly skimming the code it  IMHO looks like a slightly more complex unit test?).

What is missing is fully E2E tests (and a framework to support them) that: setup vclusters, run Argo CD, run principal agent, run one or more of managed/autonomous agents, then test the resulting setup.

#### Criteria:

Initial brainstorm of how the test would be setup and what the test would do:
* Installs and starts k3s (same as we have done in e.g.  [Argo CD](https://github.com/argoproj/argo-cd/blob/0894ddc89e8f9ce26f41d7250ac4fa624b6adf2d/.github/workflows/ci-build.yaml#L386) or [Argo CD operator](https://github.com/argoproj-labs/argocd-operator/blob/51648967810cfed7c3573b80191effeb5c48d65b/.github/workflows/ci-build.yaml#L44) )
* Setup environment, similar to/based on `hack/demo-env`:
    * Creates vclusters vcluster-control-plane/vcluster-agent-managed/vcluster-agent-autonomous on that k3s instance
        * K3s appears to include a built-in LoadBalancer already, ServiceLB
    * Installs Argo CD to vclusters
    * Starts principal
    * Starts autonomous/managed agents (do we want to start both, or should we only start one or the other)?
    * Logic for all of this can be borrowed from https://github.com/jannfis/argocd-agent/tree/main/hack/demo-env
* Runs E2E tests
    * A simple managed-mode E2E test
    * A simple autonomous-mode E2E test
    * I'd imagine these simple E2E test would be something like:
        * Create an Argo CD Application in principal, then verify it is synced/deployed on managed-mode/autonomous agent vcluster
        * Verify the resources exist on those target clusters
    * Cleans up when done
* E2E tests to be written with testify, similar to the other unit tests already in the codebase.
[...]

PR #151, opened 2024-08-14
chore: add additional comments to code/structs,  and sanity tests to verify functions are called within the expected context

jgwest:
This PR:
- Adds comments to important structs/fields within the code.
    - Comments describe the behaviour as it currently exists, but I'm happy to revert/modify any items that are expected to have a different future behaviour.
- Adds some sanity tests to code to ensure that agent-only code is only run in agent-only context, and vice versa for principal.
- We don't have E2E tests setup for this repo yet (see issue #137), so I manually verified I've not broken anything by  setting up the demo-env and ensuring the basic workflow still works.

### DIFF
--- a/agent/connection.go
+++ b/agent/connection.go
@@ -61,6 +61,7 @@ func (a *Agent) sender(stream eventstreamapi.EventStream_SubscribeClient) error 
 		"direction":   "Send",
 		"client_addr": grpcutil.AddressFromContext(stream.Context()),
 	})
+
 	q := a.queues.SendQ(a.remote.ClientID())
 	if q == nil {
 		return fmt.Errorf("no send queue found for the remote principal")
@@ -75,7 +76,7 @@ func (a *Agent) sender(stream eventstreamapi.EventStream_SubscribeClient) error 
 	}
 	logCtx.Tracef("Grabbed an item")
 	if item == nil {
-		// FIXME: Is this really the right thing to do?
+		// TODO: Is this really the right thing to do?
 		return nil
 	}
 
@@ -98,7 +99,7 @@ func (a *Agent) sender(stream eventstreamapi.EventStream_SubscribeClient) error 
 		if grpcutil.NeedReconnectOnError(err) {
 			return err
 		} else {
-			logCtx.Infof("Error while sending: %v", err)
+			logCtx.Errorf("Error while sending: %v", err)
 			return nil
 		}
 	}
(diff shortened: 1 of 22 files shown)


principal/server.go
```diff
@@ -238,29 +254,30 @@ func (s *Server) loadTLSConfig() (*tls.Config, error) {
 	return tlsConfig, nil
 }
 
-// Listener returns the listener of Server s
-func (s *Server) Listener() *Listener {
+// ListenerForE2EOnly returns the listener of Server s
```
jgwest, line 257: For all the `*E2EOnly` field changes here:

When I initially saw these struct fields, I noticed these fields are not exported from the struct, which implies they are for internal package use only. 

BUT, then I noticed we are making the contents of these fields available by exporting these functions. For example, `Queues()` could be called to retrieve the internal queues reference of the principal.

Finally, I noticed that they were only being used for the E2E tests, so I have renamed these functions to allow them to still be used by E2E tests, but to make it obvious they are not for general use.

(If you were expecting these functions WOULD be available for general use, or just generally don't like the name scheme, I'm happy to revert this part of the PR)

agent/connection.go
```diff
@@ -75,7 +76,7 @@ func (a *Agent) sender(stream eventstreamapi.EventStream_SubscribeClient) error
 	}
 	logCtx.Tracef("Grabbed an item")
 	if item == nil {
-		// FIXME: Is this really the right thing to do?
+		// TODO: Is this really the right thing to do?
```
jgwest, line 79: Changed `FIXME` to `TODO`. This seemed to be the only instance of `FIXME` in the code. Happy to revert this part of the PR if `FIXME` was intentional.

agent/connection.go
```diff
@@ -133,26 +134,6 @@ func (a *Agent) receiver(stream eventstreamapi.EventStream_SubscribeClient) erro
 	if err != nil {
 		logCtx.WithError(err).Errorf("Unable to process incoming event")
 	}
-	// switch ev.Type() {
```
jgwest, line 136: This code was moved into its own function which is called above, and is no longer needed.

principal/server.go
jannfis: Good points there, great observations. I think at some point in time we made use of at least the `Listener` function, but since then move forward do something else.

I guess I'm fine with renaming those functions for now.

jannfis (approved):
LGTM. Thanks!

PR #160, opened 2024-08-19
refactor: reduce visibility of appManager/informer fields, rename application field of ApplicationManager

jgwest:
This PR:
- Reduces the visibility of fields that are not currently used by other packages
    - As always, happy to revert any changes for cases where there are expected uses for these fields, or some grander vision
- In `ApplicationManager`, renamed `application` field -> `applicationBackend`, to clarify it is a backend for storing applications, and not a specific application itself (which was already pretty obvious from the context, but still IMHO worth renaming)
- Add an initial makefile target for running the E2E tests in `test/e2e` folder

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,10 @@ test:
 	mkdir -p test/out
 	./hack/test.sh
 
+.PHONY: test-e2e
+test-e2e:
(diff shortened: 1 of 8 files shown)


jannfis (approved):
LGTM, thanks!

We can always revert visibility back to public when it's necessary.

PR #161, opened 2024-08-20
fix agent resources manifest

anandrkskd:
This PR fixes the manifest to use correct configmaps keys in deployment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,10 @@ test:
 	mkdir -p test/out
 	./hack/test.sh
 
+.PHONY: test-e2e
+test-e2e:
+	go test -race -timeout 60s github.com/argoproj-labs/argocd-agent/test/e2e
+
 .PHONY: mod-vendor
 mod-vendor:
 	go mod vendor
--- a/agent/inbound_test.go
+++ b/agent/inbound_test.go
@@ -90,8 +90,7 @@ func Test_UpdateApplication(t *testing.T) {
 
 	t.Run("Update application using patch in managed mode", func(t *testing.T) {
 		a.mode = types.AgentModeManaged
-		a.appManager.Role = manager.ManagerRoleAgent
-		a.appManager.Mode = manager.ManagerModeManaged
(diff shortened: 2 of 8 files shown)


install/kubernetes/agent/agent-tls-secret.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: argocd-agent-tls
+data:
+  ca.crt: ""
+  tls.crt: ""
+  tls.key: ""
+type: Opaque
+
```
jannfis: I don't think this should be part of the installation manifests, since it contains eventually secret data (`tls.key`) and eventually is managed by an external entity such as ESO.

install/kubernetes/agent/argo-agent-agent-userpass.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: argocd-agent-agent-userpass
+stringData:
+  credentials: ""
```
jannfis: Likewise, I don't think this should be part of the installation manifests, since it contains eventually secret data (`credentials`) and eventually is managed by an external entity such as ESO.

jannfis (changes requested):
I have some comments, PTAL. Was there a specific reasons you included the two secrets in the installation manifests?

install/kubernetes/agent/argo-agent-agent-userpass.yaml
anandrkskd: Now I understand, as eventually it will be managed by ESO it is good Idea to remove it. My Idea to have it was to have a list of resources that should be present for both components (agent and principal) and template of the secret. But it can be done by documentation and this can be avoided.
Same for the above comment.

anandrkskd:
Created another PR https://github.com/argoproj-labs/argocd-agent/pull/179

PR #162, opened 2024-08-20
refactor: Reduce dependence on client-go API outside of k8s backend

jgwest:
My understanding of the Application backend abstraction is that it is designed to be agnostic to the particular implementation [1]. And on top of that exists the `ApplicationManager`, which is a generic mechanism for interacting with any backend implementation.

At present we only have the Kubernetes backend, and Kubernetes is our primary interface, so naturally some of the implementation details of that backend will leak out. 

This PR seeks to make a fairly minor change to slightly unleak some of those, as a way to pick some low hanging fruit in this area.

[1] But note that agent role code explicitly states that it requires the Kubernetes-based backend.

**This PR:**

- Agent role code changes:
    - Remove direct kubernetes API objects (k8s client-go, argo application client-go, app informer APIs) from Agent. These should be accessed via ApplicationManager.
        - While it's true that agent role code DOES require K8s backend, IMHO agent code should still access that backend via ApplicationManager, rather than via direct k8s client-go calls.
        - Likewise switches `deleteApplication` in `agent/inbound.go` to call `Delete` from `ApplicationManager`, rather than calling `Delete` on client-go K8s API directly.

- Principal role code changes:
    - Remove appinformer from `Server` struct, `ApplicationManager` should be used instead

- Add `EnsureSynced`, and `deletionPropagationBackground` param to `Delete`, to `Application` backend interface
    - To some extent this might seem like it is INCREASING the abstraction leak, rather than decreasing :)
    - However, my expectation is the concept of 'blocking until initial sync', and 'delete child objects in background' are universal enough that they will be useful for other potential backends (or in the worse case, would be a no-op in those backends)
    - At some point in the future, when we have other backends, we can tweak the specific API names to be more generic

### DIFF
--- a/agent/inbound.go
+++ b/agent/inbound.go
@@ -17,6 +17,7 @@ package agent
 import (
 	"fmt"
 
+	"github.com/argoproj-labs/argocd-agent/internal/backend"
 	"github.com/argoproj-labs/argocd-agent/internal/event"
 	"github.com/argoproj-labs/argocd-agent/pkg/types"
 	"github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
@@ -160,7 +161,8 @@ func (a *Agent) deleteApplication(app *v1alpha1.Application) error {
 
 	logCtx.Infof("Deleting application")
 
-	err := a.appManager.Delete(a.context, a.namespace, app, true)
+	deletionPropagation := backend.DeletePropagationBackground
+	err := a.appManager.Delete(a.context, a.namespace, app, &deletionPropagation)
 	if err != nil {
 		return err
 	}
--- a/agent/inbound.go
+++ b/agent/inbound.go
@@ -21,7 +21,6 @@ import (
 	"github.com/argoproj-labs/argocd-agent/pkg/types"
 	"github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 	"github.com/sirupsen/logrus"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 /*
@@ -161,8 +160,7 @@ func (a *Agent) deleteApplication(app *v1alpha1.Application) error {
 
 	logCtx.Infof("Deleting application")
 
-	delPol := v1.DeletePropagationBackground
-	err := a.appclient.ArgoprojV1alpha1().Applications(a.namespace).Delete(a.context, app.Name, v1.DeleteOptions{PropagationPolicy: &delPol})
+	err := a.appManager.Delete(a.context, a.namespace, app, true)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


internal/manager/application/application.go
```diff
@@ -114,10 +114,18 @@ func NewApplicationManager(be backend.Application, namespace string, opts ...App
 	return m, nil
 }
 
+// StartBackend informs the backend to run startup logic, which usually means beginning to listen for events.
+// For example, in the case of the Kubernetes backend, the shared informer is started, which will listen for Application events from the watch api of the K8s cluster.
 func (m *ApplicationManager) StartBackend(ctx context.Context) {
 	m.applicationBackend.StartInformer(ctx)
 }
 
+// EnsureSynced waits until either the backend indicates it has fully synced, or the
```
jgwest, line 123: See PR description for my thoughts on whether `EnsureSynced` and `deletionPropagationBackground` are reasonable to include in Application backend API.

agent/inbound.go
```diff
@@ -161,8 +160,7 @@ func (a *Agent) deleteApplication(app *v1alpha1.Application) error {
 
 	logCtx.Infof("Deleting application")
 
-	delPol := v1.DeletePropagationBackground
-	err := a.appclient.ArgoprojV1alpha1().Applications(a.namespace).Delete(a.context, app.Name, v1.DeleteOptions{PropagationPolicy: &delPol})
+	err := a.appManager.Delete(a.context, a.namespace, app, true)
```
jgwest: I verified that the likely reason we were using appclient Delete call here (rather than appManager.Delete) is for DeletePropagationBackground. I've thus added it to app manager.

internal/backend/interface.go
```diff
@@ -45,9 +46,10 @@ type Application interface {
 	List(ctx context.Context, selector ApplicationSelector) ([]v1alpha1.Application, error)
 	Create(ctx context.Context, app *v1alpha1.Application) (*v1alpha1.Application, error)
 	Get(ctx context.Context, name string, namespace string) (*v1alpha1.Application, error)
-	Delete(ctx context.Context, name string, namespace string) error
+	Delete(ctx context.Context, name string, namespace string, deletionPropagationBackground bool) error
```
jannfis: How about instead of making this a bool to decide on foreground vs. background deletion, to make this into a dedicated type which follows current Kubernetes notations of background, foreground and orphan?

jannfis (commented):
I'm good with this change, just have a small topic up for discussion.

internal/backend/interface.go
jgwest: @jannfis Done! If you wanted me to further wrap the DeletionPropagation-style type in a DeletionOptions-style struct, let me know (my guess was that you did not).

jannfis (approved):
LGTM, thanks!

PR #177, opened 2024-09-09
feat: Automatic creation of namespaces on control plane for autonomous agents

### LINKED ISSUE #109: Agent namespace should be created on control plane upon agent connection
Right now, an agent's "home" namespace on the control plane cluster must exist before an agent connects. As we control which agents are able to connect (by means of authentication), the principal should create an agent's namespace when the agent connects and does not yet exist.

This feature should come with certain constraints:
* An on/off toggle
* An optional regexp pattern that controls which namespaces are actually allowed to be created, user configurable
* An optional set of labels and annotations to be applied to the created namespace, user configurable

Under no circumstances should a real error (such as, could not create namespace xyz) be propagated to the agent. Instead, when namespace creation fails, the real reason must be logged on the principal's side and the agent will be declined with a generic "Authentication failed" error.

jannfis:
Closes #109

### DIFF
--- a/cmd/principal/main.go
+++ b/cmd/principal/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/argoproj-labs/argocd-agent/internal/auth"
 	"github.com/argoproj-labs/argocd-agent/internal/auth/userpass"
 	"github.com/argoproj-labs/argocd-agent/internal/env"
+	"github.com/argoproj-labs/argocd-agent/internal/labels"
 	"github.com/argoproj-labs/argocd-agent/principal"
 
 	"github.com/sirupsen/logrus"
@@ -53,6 +54,7 @@ func NewPrincipalRunCommand() *cobra.Command {
 		clientCertSubjectMatch bool
 		autoNamespaceAllow     bool
 		autoNamespacePattern   string
+		autoNamespaceLabels    []string
 	)
 	var command = &cobra.Command{
 		Short: "Run the argocd-agent principal component",
@@ -82,7 +84,14 @@ func NewPrincipalRunCommand() *cobra.Command {
 			opts = append(opts, principal.WithListenerAddress(listenHost))
 			opts = append(opts, principal.WithListenerPort(listenPort))
 			opts = append(opts, principal.WithGRPC(true))
-			opts = append(opts, principal.WithAutoNamespaceCreate(autoNamespaceAllow, autoNamespacePattern))
+			nsLabels := make(map[string]string)
+			if len(autoNamespaceLabels) > 0 {
+				nsLabels, err = labels.StringsToMap(autoNamespaceLabels)
+				if err != nil {
+					cmd.Fatal("Could not parse auto namespace labels: %v", err)
+				}
+			}
+			opts = append(opts, principal.WithAutoNamespaceCreate(autoNamespaceAllow, autoNamespacePattern, nsLabels))
 
 			if !disableMetrics {
 				opts = append(opts, principal.WithMetricsPort(metricsPort))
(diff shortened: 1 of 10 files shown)


cmd/principal/main.go
```diff
@@ -173,6 +185,15 @@ func NewPrincipalRunCommand() *cobra.Command {
...
 		env.StringSliceWithDefault("ARGOCD_PRINCIPAL_ALLOWED_NAMESPACES", nil, []string{}),
 		"List of namespaces the server is allowed to operate in")
+	command.Flags().BoolVar(&autoNamespaceAllow, "namespace-create-enabled",
+		env.BoolWithDefault("ARGOCD_PRINCIPAL_NAMESPACE_CREATE_ALLOW", false),
+		"Whether to allow automatic namespace creation for autonomous agents")
+	command.Flags().StringVar(&autoNamespacePattern, "namespace-create-pattern",
+		env.StringWithDefault("ARGOCD_PRINCIPAL_NAMESPACE_CREATE_PATTERN", nil, ""),
+		"Only automatically create namespaces matching pattern")
+	command.Flags().StringSliceVar(&autoNamespaceLabels, "namespace-create-labels",
+		env.StringSliceWithDefault("ARGOCD_PRINCIPAL_NAMESPACE_CREATE_PATTERN", nil, []string{}),
```
ishitasequeira: I think the environment variable needs to be `ARGOCD_PRINCIPAL_NAMESPACE_CREATE_LABELS`

cmd/principal/main.go
```diff
@@ -173,6 +185,15 @@ func NewPrincipalRunCommand() *cobra.Command {
 	command.Flags().StringSliceVar(&allowedNamespaces, "allowed-namespaces",
 		env.StringSliceWithDefault("ARGOCD_PRINCIPAL_ALLOWED_NAMESPACES", nil, []string{}),
 		"List of namespaces the server is allowed to operate in")
+	command.Flags().BoolVar(&autoNamespaceAllow, "namespace-create-enabled",
+		env.BoolWithDefault("ARGOCD_PRINCIPAL_NAMESPACE_CREATE_ALLOW", false),
```
ishitasequeira: Considering the command line param says `enabled` and env variable name says `ALLOW`, I think it would make it easier to use it consistently in both places. WDYT?

principal/event.go
```diff
@@ -230,3 +242,29 @@ func (s *Server) StartEventProcessor(ctx context.Context) error {
...
+		if !kerrors.IsNotFound(err) {
+			return false, err
+		} else {
+			_, err = s.kubeClient.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+				ObjectMeta: v1.ObjectMeta{
+					Name:   name,
+					Labels: s.autoNamespaceLabels,
+				}},
+				v1.CreateOptions{})
+			return true, err
```
ishitasequeira, line 268: Should we be checking if the Namespace is marked for deletion and skip if Yes?

cmd/principal/main.go
jannfis: Fixed, thanks!

cmd/principal/main.go
jannfis: Renamed everything to "enable" instead of "allow". Great feedback!

principal/event.go
jannfis: Interesting idea. I think if the namespace is marked for deletion, the initial `Get` call would not return an  error (i.e. it would return a resource), and thereby skipping the `Create` logic already.

I think this an edge case we should not tackle at this point. However, I put a TODO comment in the code so that we may revisit this later on.

Are you OK with that?

ishitasequeira: That sounds good to me.

PR #179, opened 2024-09-11
fix: fix agent resources manifest

anandrkskd:
This PR fixes the manifest to use correct config maps keys in deployment


jannfis (approved):
LGTM! Thanks

PR #187, opened 2024-09-24
chore: use goreman to start principal and agent

jparsai:
This PR is to make use of `goreman` to run principal and agent in same console, this would make debugging easy as logs of both components would be printed in same console, also it would help to understand the order of execution of agent and principal code blocks.

### DIFF
--- a/hack/demo-env/Makefile
+++ b/hack/demo-env/Makefile
@@ -1,0 +1,10 @@
+AUTONOMOUS_MODE?=false
+GOBIN=$(shell go env GOPATH)/bin
+
+.PHONY: start-local
+start-local:
+	AUTONOMOUS_MODE=$(AUTONOMOUS_MODE) $(GOBIN)/goreman -set-ports=false -f Procfile start
+
+.PHONY: help
+help:
+	@echo "Not yet, sorry."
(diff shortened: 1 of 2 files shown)


jannfis (commented):
Thanks, I do like this idea.

However, I think neither goreman nor make configuration should be part of the top-level for now. 

My suggestion would be to place the Procfile into (a new) `hack/demo-env/Procfile` and Makefile (if required at all) into `hack/demo-env/Makefile` instead.

jannfis (commented):
Also, I think that there really should be two agent targets in `Procfile`, `agent-managed` and `agent-autonomous`. Or some logic that lets you start one or the other.

jparsai:
Thanks for looking into this. I moved `Procfile` inside `hack/demo-env/` directory, but moving `Makefile` requires changes in start up scripts, because they have to be executed from root folder. I was not sure if I should modify start up scripts for this.
To set agent mode I introduced an env variable `AUTONOMOUS_MODE`, which can be used like this

`AUTONOMOUS_MODE=true make start-local
`

jannfis (approved):
LGTM!

PR #188, opened 2024-09-26
chore(scripts): Fix setup script

jopit:
On my Mac, the result of `wc -l` is padded on the left with whitespace, causing the `setup-vcluster-env.sh` script to fail. This PR provides a fix for that.


jannfis (approved):
LGTM, thanks!

PR #190, opened 2024-10-01
feat: Remove hard requirements for HTTP/2

chetan-rns:
gRPC relies on HTTP/2 for multiplexing requests and is required for bi-directional streaming. However, intermediate infra components (load balancers, proxy, etc.) could be incompatible with HTTP/2 and may downgrade to HTTP/1.1. For these cases, we provide an option to use gRPC over WebSocket for bi-directional streaming that doesn't require HTTP/2. https://github.com/stackrox/go-grpc-http1 handles the translation between gRPC and WebSocket.

* Flags to enable WebSocket
* Update the principal and the agent code to use the go-grpc-http1 client if the flag is enabled
* e2e tests with a HTTP/2 incompatible reverse proxy

### DIFF
--- a/test/proxy/proxy.go
+++ b/test/proxy/proxy.go
@@ -1,0 +1,84 @@
+package proxy
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"log"
+	"math/big"
+	"net"
+	"net/http"
+	"net/http/httputil"
+	"path"
+	"testing"
+
+	"github.com/argoproj-labs/argocd-agent/test/fake/testcerts"
+)
+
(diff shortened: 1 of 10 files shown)


test/proxy/proxy.go
```diff
@@ -0,0 +1,84 @@
...
+
+	server := &http.Server{
+		TLSConfig: &tls.Config{
+			InsecureSkipVerify: true,
+		},
+		Handler: downgradeToHTTP1Handler(target),
+		Addr:    lis.Addr().String(),
+	}
+
+	go server.ServeTLS(lis, basePath+".crt", basePath+".key")
```
jannfis, line 38:
```suggestion
    //nolint:errcheck
	go server.ServeTLS(lis, basePath+".crt", basePath+".key")
```

PR #199, opened 2024-10-09
fix: principal install missing argocd-agent-principal-userpass secret

mikeshng:
I am following the [quickstart](https://github.com/argoproj-labs/argocd-agent/blob/main/docs/hack/quickstart.md) and after the step `kubectl apply -n argocd -k install/kubernetes/principal`, the principal agent pod is not coming up.

The error was:
```
MountVolume.SetUp failed for volume "userpass-passwd" : secret "argocd-agent-principal-userpass" not found
```

This fix is to ensure the kustomization.yaml properly contains the required secret for install.

### DIFF
--- a/install/kubernetes/principal/kustomization.yaml
+++ b/install/kubernetes/principal/kustomization.yaml
@@ -7,3 +7,4 @@ resources:
 - principal-deployment.yaml
 - principal-service.yaml
 - principal-params-cm.yaml
+- principal-userpass-secret.yaml


jannfis:
Thanks! However, I believe this is a documentation issue.

IMHO, the secret should not be installed automatically, but users should create the secret manually - so the manifests can be used out of the box even if you use a secret manager to create/manage that secret.

So from my PoV, the real fix would be to add the manual step to the docs :)

PR #200, opened 2024-10-09
fix: casing for Dockerfiles and default make image repo url

mikeshng:
Small fix to Dockerfiles to avoid getting a warning when running `make image-...`

```                                                                                                                                                                                                                                          
 1 warning found (use docker --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)
```

Adjust the default image repo from a personal repo to the default deployment repo.


jannfis (approved):
Good catch, thanks!

LGTM.

PR #202, opened 2024-10-09
fix: principal agent not starting up with the default autoNamespaceLabels config

mikeshng:
The principal agent pod kept crashlooping with the error:

```
[FATAL]: Could not parse auto namespace labels: invalid label '': has no value
```

I am using the default values so the auto namespace labels is an empty string.
This fix is to ensure that the principal can properly start with the default case.

Also contains a minor logging fix to address the following:
```
level=info msg="Listing AppProjects &{0xc0002579a0 }" module=AppProjectInformer
```


jannfis (approved):
LGTM, thanks!

PR #204, opened 2024-10-11
fix-demo: misc demo related fixes

mikeshng:
I've tried to setup the demo/dev env and encountered a few issues. This PR is to address those issues:

1) When I ran the `setup-vcluster-env.sh` script, got the following error:
```
  --> Creating vcluster control-plane
18:40:36 fatal unknown flag: --switch-context
ERROR: Error occurred, terminating.
```
`switch-context` is no longer a valid flag for `vcluster create` in vcluster version 0.20.1 stable and there are no similar replacement flags. After I removed the flag, the script ran fine and ultimately the demo/dev env setup was successful so for now I think it's safe to remove this flag.

2) When I ran the `start-*` scripts, I got the following error:
```
set: Illegal option -o pipefail
```
The pipefail option seems to be a [bash specific feature](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html). I've added the bash shebang in this PR.
Also added the set x to `start-agent-autonomous.sh` to be consistent with the other start scripts.

3) misc demo README fixes for consistency.

### DIFF
--- a/hack/demo-env/README.md
+++ b/hack/demo-env/README.md
@@ -40,7 +40,7 @@ sudo microk8s.config > ~/.kube/config
 
 What's left is to install the metallb configuration. You should be good now.
 
-```
+```shell
 kubectl apply -n metallb-system -f hack/demo-env/resources/metallb-ipaddresspool.yaml
 ```
 
@@ -50,7 +50,7 @@ Make sure you have administrative access to a cluster and [vcluster](https://git
 
 To setup all required virtual clusters, run
 
-```
+```shell
 ./hack/demo-env/setup-vcluster-env.sh create
 ```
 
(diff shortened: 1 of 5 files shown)


jannfis:
It's weird that the `--switch-context`  is not available in v0.20.1, because it's been introduced in v0.20.0 :thinking: 

I just upgraded from v0.20.0 to v0.20.1 to see the switch being gone.

mikeshng:
@jannfis if you have some cycles, could you PTAL again? I want to add some ocm-io addon steps to this doc and I want to avoid having to deal with conflicts.

jannfis:
Sure, sorry for the delay @mikeshng. I was wondering about the impact of removing `--switch-context` without any replacement. I think we had to either switch back context manually, or use `--connect=false`.

Let me run a quick test.

jannfis:
OK, it seems to still work as expected.

jannfis (approved):
LGTM

PR #206, opened 2024-10-23
fix: fix gen-creds.sh

jopit:
The gen-creds.sh script sets up an environment variable alias for pwmake in case the system has pwgen instead, but then doesn't use it.


jannfis (approved):
Good catch!

LGTM.

PR #209, opened 2024-10-30
feat: Add informer metrics

jannfis:
This PR adds metrics to the generic informer, conveniently handled within the callbacks.

Additionally, it adds some unit tests for the informer package.

The next step will move the existing application informer to use the generic informer, too.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -83,11 +83,6 @@ jobs:
         uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
         with:
           go-version: ${{ env.GOLANG_VERSION }}
-      - name: Restore go build cache
-        uses: actions/cache@13aacd865c20de90d75de3b17ebe84f7a17d57d2 # v4.0.0
-        with:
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yaml
```diff
@@ -83,11 +83,6 @@ jobs:
         uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
         with:
           go-version: ${{ env.GOLANG_VERSION }}
-      - name: Restore go build cache
-        uses: actions/cache@13aacd865c20de90d75de3b17ebe84f7a17d57d2 # v4.0.0
-        with:
-          path: ~/.cache/go-build
-          key: ${{ runner.os }}-go-build-v1-${{ github.run_id }}
```
jannfis, line 90: Cache was not used anyway.

.github/workflows/ci.yaml
```diff
@@ -104,11 +99,6 @@ jobs:
...
```
jannfis, line 111: Cache was not used anyway

ishitasequeira (approved):
Lgtm!!

PR #193, opened 2024-10-04
chore(scripts): Let demo scripts run from anywhere in the tree

jannfis:
This removes the requirements to start the scripts in `hack/demo-env` from the root of the repository tree.

PR #211, opened 2024-10-30
feat: Initial E2E framework and simple tests

jopit:
Initial E2E framework and simple tests

Resolves issue #137


jannfis (approved):
LGTM

Makefile
```diff
@@ -24,6 +24,19 @@ MOCKERY_V2_VERSION?=v2.43.0
 .PHONY: build
 build: agent principal
 
+.PHONY: setup-e2e2
+setup-e2e2:
+	test/e2e2/test-env/setup-vcluster-env.sh create
+
+.PHONY: start-argocd-agent
```
jgwest, line 31: Re: the `Makefile` target of `start-argocd-agent`. Rather than `start-argocd-agent`, which might imply this would start Argo agent for development purposes (similar to `make start` from Argo CD), WDYT about:
- `start-argocd-agent-for-e2e`  (maybe too long?) 
- `start-e2e` / `start-e2e2` (shorter, this is what we use in argo cd, and in rollouts operator)
- Other options?

test/e2e2/fixture/kubeclient.go
```diff
@@ -0,0 +1,271 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
+)
+
+type KubeObject interface {
```
jgwest, line 48: Could add a comment here for `KubeObject` and `KubeObjectList` indicating these are similar to `Object` and `ObjectList` from controller-runtime, or even more generally that this package aims to provide a similar style interface to controller-runtime (which we are avoiding adding a dependency on here).

IMHO this might be useful comment to add (and presuming my assumption is true), but it's up to you!

test/e2e2/basic_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	key := fixture.ToNamespacedName(&app)
+
+	// Ensure the app has been pushed to the managed-agent
+	requires.Eventually(func() bool {
+		app := argoapp.Application{}
+		err := suite.ManagedAgentClient.Get(suite.Ctx, key, &app, metav1.GetOptions{})
+		return err == nil
+	}, 30*time.Second, 1*time.Second)
+
```
jgwest, line 71: Can we check that the content (`.spec` field)  is the same between the managed-agent and principal, as well? It's very likely, but it can't hurt to check.

test/e2e2/basic_test.go
```diff
@@ -0,0 +1,140 @@
...
+	// Delete the app from the principal
```
jgwest, line 103: Let me know if you think this is scope creep from the original requirements, but before we delete the Application in the next step, can we test the update case, as well? 

E.g. just making a small change to one of the fields of the Application on principal, and making sure it percolates down to managed-agent.

test/e2e2/basic_test.go
```diff
@@ -0,0 +1,140 @@
...
+	requires.Eventually(func() bool {
+		app := argoapp.Application{}
+		err := suite.PrincipalClient.Get(suite.Ctx, key, &app, metav1.GetOptions{})
+		return err == nil
+	}, 30*time.Second, 1*time.Second)
+
+	// Delete the app from the autonomous-agent
+	err = suite.AutonomousAgentClient.Delete(suite.Ctx, &app, metav1.DeleteOptions{})
+	requires.NoError(err)
+
```
jgwest, line 192: Same as above, can be useful to check `.spec` is in sync, and to test the update case before deleting.

test/e2e2/fixture/fixture.go
```diff
@@ -0,0 +1,153 @@
...
+
+	config, err = GetSystemKubeConfig("vcluster-agent-autonomous")
+	requires.Nil(err)
+	suite.AutonomousAgentClient, err = NewKubeClient(config)
+	requires.Nil(err)
+}
+
+func (suite *BaseSuite) SetupTest() {
+	err := CleanUp(suite.Ctx, suite.PrincipalClient, suite.ManagedAgentClient, suite.AutonomousAgentClient)
+	suite.Assert().Nil(err)
```
jgwest: Perhaps, `Require` rather than `Assert`? I presume we would want to fail fast if initial cleanup didn't succeed.

test/e2e2/fixture/kubeconfig.go
```diff
@@ -0,0 +1,58 @@
...
+
+	restConfig, err := clientConfig.ClientConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	err = setRateLimitOnRestConfig(restConfig)
+	if err != nil {
+		return nil, err
+	}
```
jgwest: I think we can remove this `setRateLimitOnRestConfig` part of the function below: if I'm not mistaken the `QPS`/`Burst` values came from managed-gitops, where it made slightly more sense to have a generous rate limit (since managed-gitops was a managed service). But as part of an E2E test, I don't think it's needed (and probably could also have been removed from managed-gitops E2E test setup code as well, for the same reason)

test/e2e2/fixture_test.go
```diff
@@ -0,0 +1,543 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/restmapper"
+)
+
+type FixtureTestSuite struct {
```
jgwest, line 41: Could add a comment like `'FixtureTestSuite' is a set of tests to verify that the E2E test fixture code itself is working as designed, rather than testing the project's components.` Perhaps the comment is too obvious, and thus unnecessary, up to you if you want to add.

test/e2e2/sync_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+	key := fixture.ToNamespacedName(&app)
+
+	// Ensure the app has been pushed to the managed-agent
+	requires.Eventually(func() bool {
+		app := argoapp.Application{}
+		err := suite.ManagedAgentClient.Get(suite.Ctx, key, &app, metav1.GetOptions{})
+		return err == nil
+	}, 30*time.Second, 1*time.Second)
+
```
jgwest, line 129: As mentioned elsewhere, can be useful to verify that the `.Spec` fields are the same in both places.

test/e2e2/sync_test.go
```diff
@@ -0,0 +1,256 @@
...
+	agentKey := types.NamespacedName{Name: app.Name, Namespace: app.Namespace}
+	principalKey := types.NamespacedName{Name: app.Name, Namespace: "agent-autonomous"}
+
+	// Ensure the app has been pushed to the principal
+	requires.Eventually(func() bool {
+		app := argoapp.Application{}
+		err := suite.PrincipalClient.Get(suite.Ctx, principalKey, &app, metav1.GetOptions{})
+		return err == nil
+	}, 30*time.Second, 1*time.Second)
+
```
jgwest, line 240: As above, couldn't hurt to verify `.spec`s are the same.

test/e2e2/test-env/start-agent-autonomous.sh
```diff
@@ -0,0 +1,24 @@
...
+# limitations under the License.
+
+set -e -o pipefail
+ARGS=$*
+if ! kubectl config get-contexts | tail -n +2 | awk '{ print $2 }' | grep -qE '^vcluster-agent-autonomous$'; then
+    echo "kube context vcluster-agent-autonomous is not configured; missing setup?" >&2
+    exit 1
+fi
+SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+test -f cmd/agent/main.go || (echo "Script should be run from argocd-agent's root path" >&2; exit 1)
```
jgwest: Rather than testing that the script is run from the correct place, would it be preferable to just add `cd $SCRIPTPATH/../../..` before `go run` to ensure that this is always true?

test/e2e2/test-env/start-agent-managed.sh
```diff
@@ -0,0 +1,24 @@
...
+# limitations under the License.
+
+set -ex -o pipefail
+ARGS=$*
+if ! kubectl config get-contexts | tail -n +2 | awk '{ print $2 }' | grep -qE '^vcluster-agent-managed$'; then
+    echo "kube context vcluster-agent-autonomous is not configured; missing setup?" >&2
+    exit 1
+fi
+SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+test -f cmd/agent/main.go || (echo "Script should be run from argocd-agent's root path" >&2; exit 1)
```
jgwest: Same as automonous above, can use SCRIPTPATH if it makes sense.

test/e2e2/test-env/start-principal.sh
```diff
@@ -0,0 +1,24 @@
...
+# limitations under the License.
+
+set -ex -o pipefail
+ARGS=$*
+if ! kubectl config get-contexts | tail -n +2 | awk '{ print $2 }' | grep -qE '^vcluster-control-plane$'; then
+    echo "kube context vcluster-agent-autonomous is not configured; missing setup?" >&2
+    exit 1
+fi
+SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+test -f cmd/principal/main.go || (echo "Script should be run from argocd-agent's root path" >&2; exit 1)
```
jgwest: Same as autonomous above, can use SCRIPTPATH if it makes sense.

jgwest (changes requested):
Looks great, very nice, thanks @jopit! I added some minor comments I noticed while reading through the code, let me know what you think.

I had a couple of questions but they are more general project questions for everyone, so I will loop in Jann et al.

One other question from me:			

Any changes in 'argocd-agent/test/e2e2' that you would suggest we may want to port back to 'argocd-agent/hack/demo-env'?
- For example, extra timeout logic in 'setup-vcluster-env.sh' looked useful for both
- Removal of `.data` from Secrets defined in `argocd-secret.yaml` et al

Makefile
jopit: Done. Changed it to `start-e2e2`

test/e2e2/fixture/kubeclient.go
jopit: Done.

test/e2e2/basic_test.go
jopit: Done.

test/e2e2/basic_test.go
jopit: Done.

test/e2e2/basic_test.go
jopit: Done.

test/e2e2/fixture/fixture.go
jopit: Done.

test/e2e2/fixture/kubeconfig.go
jopit: Done.

test/e2e2/fixture_test.go
jopit: Done.

test/e2e2/sync_test.go
jopit: Done.

test/e2e2/sync_test.go
jopit: Done.

test/e2e2/test-env/start-agent-autonomous.sh
jopit: Resolved this by making the same changes that Jann made to the demo-env scripts to allow running from anywhere in the repo tree, PR #193

test/e2e2/test-env/start-agent-managed.sh
jopit: Resolved this by making the same changes that Jann made to the demo-env scripts to allow running from anywhere in the repo tree, PR #193

test/e2e2/test-env/start-principal.sh
jopit: Resolved this by making the same changes that Jann made to the demo-env scripts to allow running from anywhere in the repo tree, PR #193

jgwest (approved):
LGTM too, thanks @jopit!

PR #214, opened 2024-11-04
chore(scripts): fix argocd deploy release version

mikeshng:
For the demo/dev env script, the ArgoCD deployments installed on the hub and spoke are using the `latest` tag instead of the release tag. This is due to the kustomization yamls are all using `?ref=stable` but the argocd github repo still references latest (bug with argocd repo?). For example, https://github.com/argoproj/argo-cd/blob/stable/manifests/base/server/argocd-server-deployment.yaml#L21

This fix is to replace the latest tag with the latest stable release tag.

### DIFF
--- a/hack/demo-env/agent-autonomous/kustomization.yaml
+++ b/hack/demo-env/agent-autonomous/kustomization.yaml
@@ -7,5 +7,9 @@ resources:
 - https://github.com/argoproj/argo-cd/manifests/cluster-rbac/application-controller?ref=stable
 - ../common
 
+images:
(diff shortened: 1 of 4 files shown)


jannfis (commented):
Good catch.

However, wouldn't that trigger a diff in Git once you run the setup?

mikeshng:
All the files are copied to a tmp folder first before the sed modifications so it should be ok.

jannfis (approved):
You're absolutely right!

LGTM.

PR #216, opened 2024-11-05
fix: AppProject informer should be single namespace

### LINKED ISSUE #213: AppProject informer shouldn't be cluster scoped when namespace is not set
The AppProject informer should never be cluster scoped, even if the namespace is not set.

It should behave as documented [here](https://github.com/argoproj-labs/argocd-agent/blob/61613f0178528c77949a129cbbaa933b53bbe803/install/kubernetes/principal/principal-params-cm.yaml#L22-L25) and never require cluster scoped permissions to list or watch AppProjects.

jannfis:
The AppProject informer should not be cluster-scoped and instead operate in a single namespace.

Fixes #213

### DIFF
--- a/internal/informer/appproject/projectinformer.go
+++ b/internal/informer/appproject/projectinformer.go
@@ -16,7 +16,6 @@ package appproject
 
 import (
 	"context"
-	"strings"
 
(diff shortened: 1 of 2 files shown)


ishitasequeira (approved):
LGTM!!

mikeshng (approved):
/lgtm

PR #221, opened 2024-11-08
refactor: AppInformer now uses GenericInformer as foundation

jannfis:
This refactor makes the AppInformer use GenericInformer instead of rolling its own.


jannfis:
Closing this in favor of a new, truly generic informer.

PR #222, opened 2024-11-14
feat: New checkpoint module

jannfis:
This adds a new utility module for measuring time by defining checkpoints.

I plan to use this throughout the code for logging and metrics purposes.

### DIFF
--- a/internal/checkpoint/checkpoint.go
+++ b/internal/checkpoint/checkpoint.go
@@ -1,3 +1,17 @@
+// Copyright 2024 The argocd-agent Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


internal/checkpoint/checkpoint.go
```diff
@@ -0,0 +1,134 @@
...
+	cp.mutex.RLock()
+	defer cp.mutex.RUnlock()
+	return len(cp.steps)
+}
+
+// Steps returns the steps within the checkpoint cp
+func (cp *Checkpoint) Steps() []Step {
+	cp.mutex.RLock()
+	defer cp.mutex.RUnlock()
+	return cp.steps
```
jgwest: Rather than returning `cp.steps`, should we instead clone the slice (inside the mutex) and return that? 

e.g.
```go
cp.mutex.RLock()
defer cp.mutex.RUnlock()
return append([]Step{}, cp.steps)
```

My concern is that modifications (addition/removal) to the list returned by `Steps()` by the caller (unprotected by mutex), may modify the original list (protected by mutex).

jgwest (changes requested):

internal/checkpoint/checkpoint.go
jannfis: Great catch!

jannfis:
```suggestion
	return append([]Step{}, cp.steps...)
```

ishitasequeira (approved):
LGTM!!

jgwest (approved):
LGTM, thanks @jannfis!

Issue #117, opened 2024-07-10
Agents should keeping trying to send events to the principal, until the principal acknowledges that the event has been processed and persisted to the backend.


jgwest:
### **At present, there are a couple issues with event communication reliability between principal and agent:**

- If the agent is unable to send a message to the principal, the agent will skip the message (not retry it)
	- Thus, principal can get out of sync with agents for significant periods of time (depending on how often new events are sent for that Application)
	- For example, for a very quiet Application (manual sync, with not a lot of K8s resource changes), it might just appear invisible to principal if principal misses the last creation/update event
- Even if the principal successfully receives a message from agent, there is no guarantee that message will be processed.
	- For example, if the principal container restarts, any unprocessed messages in the queue will be permanently lost.
	
In GitOps Service, we solved this using a queue stored in RDBMS (the 'Operation's table). The general algorithm is the same, here, although the specifics are slightly different (because principal/agent does not persist queue entries to disk, unlike RDBMS).

### **To solve both these problems, agent <-> principal communication can work as follows:**
- A) When an agent event occurs, queue it to be sent to principal, replacing any previously waiting events for that resource:
	- If an older event for the same resource (e.g. Application) is already waiting, then replace the old resource event with the new resource event, in the queue:
		- Example: for a resource A, if 3 update events have occurred, 1, 2, 3, and none of them have been sent, we only need to send 3: because update 3 will include any updates that were made in 1 and 2.
	- That is, since each event is self-contained, and newer events necessarily include all changes from older events, there is no point is sending (or even storing in memory) old events.
	- Since we only ever need to keep track of the last event that was sent for each resource, there is minimal memory cost.

- B) Do not remove an event from the agent queue until the principal indicates that it  has been processed AND stored in the principal backend:
	- When an agent sends a message to the principal, the principal must acknowledge that it has been processed and persisted.
	- The agent must wait for the principal to acknowledge that the event has been processed/persisted, before the agent removes it from the agent queue.
	- Why? This solves this problem:
		- Agent sends events 1, 2, 3, 4, 5 to principal
		- Principal receives all events: 1, 2, 3, 4, 5
		- Principal processes events 1, 2, but then OOMs and the container is restarted.
			- Because principal does not use persistent storage for queue events, event 3, 4, and 5 are lost.
		- Principal restarts, but the event queue is empty due to restart.
		- Agent does not resend events 3, 4, 5, because it already sent them.
		- Thus principal permanently misses events 3/4/5.

[...]

Issue #94, opened 2024-06-21
Agent should synchronize existing resources with principal upon connect


jannfis:
The agent should synchronize all resources (apps, appprojects, etc) with the principal on the time it connects to ensure information on the principal is up-to-date as soon as possible.

Right now, the trigger is informer events on either side. While information will be eventually synchronized, it can take `appResync` time to do so.

When the agent connects in managed mode, it should request updates for all applications from the principal and submit back status information.

When the agent connects in autonomous mode, it should send updates for all applications to the principal.

jgwest:
Jayendra is available and taking a look at this one.

PR #225, opened 2024-11-18
feat: resend events with exponential backoff and wait for ACK

chetan-rns:
* Added a new event type `ACK` that is sent by the principal/agent when an event is processed successfully.
* `EventWriter` keeps track of the latest events and sends them on the gRPC stream. Both the agent and the principal will add events to the `EventWriter` instead of directly sending them to the gRPC stream. The `EventWriter` will send the events with exponential backoff until we receive an ACK from the other side. It continuously loops over the waiting events and sends them to the other side with minimal delay. The goal was to propagate the events as soon as possible with minimal delay and without blocking the other events.
* Added `UID` (resourceID) and `resourceVersion` (eventID) to the cloudevent as extensions. We do this to map an event to a particular state of a resource. For example, when we receive an ACK for an event we want to know if the ACK is for the latest version of a resource. 

**Note**: I think there are multiple ways of designing this and I'm happy to update the approach based on the feedback.

Other choices that were considered:

1. We send events on a stream and immediately wait for an ACK. Use apimachinery's wait.ExponentialBackOff to resend events until we receive an ACK. However, the events from other resources will be blocked until the current event is processed.
2. We can improvise the above approach by running the wait.ExponentialBackOff in a separate goroutine for each app. However, each goroutine is shortlived (they only exist until we receive an ACK) and there might be a scenario where the runtime spends more time context switching if there are a large number of apps.
3. We could potentially solve this using a pool of goroutines (like a boss-worker pattern). But I still went with the current approach (continuously loop and resend) since there is minimal delay and we don't have to wait for the ACK sequentially.

Fixes: https://github.com/argoproj-labs/argocd-agent/issues/117

### DIFF
--- a/agent/connection.go
+++ b/agent/connection.go
@@ -79,7 +79,7 @@ func (a *Agent) sender(stream eventstreamapi.EventStream_SubscribeClient) error 
 		return nil
 	}
 
-	logCtx.Trace("Adding an item to the event writer", "resourceID", event.ResourceID(ev), "eventID", event.EventID(ev))
+	logCtx.Tracef("Adding an item to the event writer: resourceID %s eventID %s", event.ResourceID(ev), event.EventID(ev))
 	a.eventWriter.Add(ev)
 
 	return nil
@@ -110,13 +110,13 @@ func (a *Agent) receiver(stream eventstreamapi.EventStream_SubscribeClient) erro
 	logCtx.Debugf("Received a new event from stream")
 
 	if ev.Target() == event.TargetEventAck {
-		logCtx.Trace("Received an ACK for an event", "resourceID", ev.ResourceID(), "eventID", ev.EventID())
+		logCtx.Tracef("Received an ACK for an event: resourceID %s eventID %s", ev.ResourceID(), ev.EventID())
 		rawEvent, err := format.FromProto(rcvd.Event)
 		if err != nil {
 			return err
 		}
 		a.eventWriter.Remove(rawEvent)
-		logCtx.Trace("Removed an event from the EventWriter", "resourceID", ev.ResourceID(), "eventID", ev.EventID())
+		logCtx.Tracef("Removed an event from the EventWriter: resourceID %s eventID %s", ev.ResourceID(), ev.EventID())
 		return nil
 	}
 
@@ -130,7 +130,7 @@ func (a *Agent) receiver(stream eventstreamapi.EventStream_SubscribeClient) erro
 			return fmt.Errorf("no send queue found for the remote principal")
 		}
 		sendQ.Add(a.emitter.ProcessedEvent(event.EventProcessed, ev))
-		logCtx.Info("Sent an ACK for an event", "resourceID", ev.ResourceID(), "eventID", ev.EventID())
+		logCtx.Tracef("Sent an ACK for an event: resourceID %s eventID %s", ev.ResourceID(), ev.EventID())
 	}
 	return nil
 }
--- a/agent/connection.go
+++ b/agent/connection.go
@@ -79,7 +79,7 @@ func (a *Agent) sender(stream eventstreamapi.EventStream_SubscribeClient) error 
 		return nil
 	}
 
-	logCtx.Tracef("Adding an item to the event writer: resourceID %s eventID %s", event.ResourceID(ev), event.EventID(ev))
+	logCtx.WithField("resource_id", event.ResourceID(ev)).WithField("event_id", event.EventID(ev)).Trace("Adding an event to the event writer")
 	a.eventWriter.Add(ev)
 
 	return nil
@@ -107,16 +107,22 @@ func (a *Agent) receiver(stream eventstreamapi.EventStream_SubscribeClient) erro
 		logCtx.Errorf("Could not unwrap event: %v", err)
 		return nil
 	}
+
+	logCtx = logCtx.WithFields(logrus.Fields{
+		"resource_id": ev.ResourceID(),
+		"event_id":    ev.EventID(),
+	})
+
 	logCtx.Debugf("Received a new event from stream")
 
 	if ev.Target() == event.TargetEventAck {
-		logCtx.Tracef("Received an ACK for an event: resourceID %s eventID %s", ev.ResourceID(), ev.EventID())
+		logCtx.Trace("Received an ACK for an event")
 		rawEvent, err := format.FromProto(rcvd.Event)
 		if err != nil {
 			return err
 		}
 		a.eventWriter.Remove(rawEvent)
-		logCtx.Tracef("Removed an event from the EventWriter: resourceID %s eventID %s", ev.ResourceID(), ev.EventID())
+		logCtx.Trace("Removed an event from the event writer")
 		return nil
 	}
 
@@ -130,7 +136,7 @@ func (a *Agent) receiver(stream eventstreamapi.EventStream_SubscribeClient) erro
 			return fmt.Errorf("no send queue found for the remote principal")
 		}
 		sendQ.Add(a.emitter.ProcessedEvent(event.EventProcessed, ev))
-		logCtx.Tracef("Sent an ACK for an event: resourceID %s eventID %s", ev.ResourceID(), ev.EventID())
+		logCtx.Trace("Sent an ACK for an event")
 	}
 	return nil
 }
--- a/agent/connection.go
+++ b/agent/connection.go
@@ -79,23 +79,8 @@ func (a *Agent) sender(stream eventstreamapi.EventStream_SubscribeClient) error 
 		return nil
 	}
 
-	logCtx.Tracef("Sending an item to the event stream")
-
-	pev, err := format.ToProto(ev)
-	if err != nil {
-		logCtx.Warnf("Could not wire event: %v", err)
-		return nil
-	}
-
-	err = stream.Send(&eventstreamapi.Event{Event: pev})
-	if err != nil {
-		if grpcutil.NeedReconnectOnError(err) {
-			return err
-		} else {
-			logCtx.Errorf("Error while sending: %v", err)
-			return nil
-		}
-	}
+	logCtx.Trace("Adding an item to the event writer", "resourceID", event.ResourceID(ev), "eventID", event.EventID(ev))
+	a.eventWriter.Add(ev)
 
 	return nil
 }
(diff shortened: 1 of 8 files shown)


agent/connection.go
```diff
@@ -86,23 +86,8 @@ func (a *Agent) sender(stream eventstreamapi.EventStream_SubscribeClient) error
...
-	err = stream.Send(&eventstreamapi.Event{Event: pev})
-	if err != nil {
-		if grpcutil.NeedReconnectOnError(err) {
-			return err
-		} else {
-			logCtx.Errorf("Error while sending: %v", err)
-			return nil
-		}
-	}
+	logCtx.Tracef("Adding an item to the event writer: resourceID %s eventID %s", event.ResourceID(ev), event.EventID(ev))
```
jannfis: Can you please use structured fields for `resourceID` and `eventID` here, e.g. 

```go 
logCtx.WithField("resource_id", event.resourceID(ev)).WithField("event_id", event.EventID(ev)).Trace(...)
```

Same goes for the other calls to the logger where these data goes. I think it will greatly help correlating the logs later on, especially if you ingest the logs into a tokenizer or use JSON logging.

jannfis:
Great stuff, @chetan-rns!

I'm starting a review cycle. Please bear with me, as my reviews are dripping in :)

principal/apis/eventstream/eventstream.go
```diff
@@ -47,6 +48,10 @@ type Server struct {
 
 	options *ServerOptions
 	queues  queue.QueuePair
+
+	// key: AgentName
+	// value: EventWriter for that agent
+	eventWriters map[string]*event.EventWriter
```
jgwest, line 61: From reading through the code, it appears `eventWriters` requires a mutex for when it is read/modified, for example, in `Subscribe` func in this file. AFAICT the existing code flow does not prevent concurrent modification.

internal/event/event.go
```diff
@@ -148,8 +210,204 @@ func (ev Event) Application() (*v1alpha1.Application, error) {
...
+	}()
+
+	// Resend the event since it is not ACK'd.
+	pev, err := format.ToProto(eventMsg.event)
+	if err != nil {
+		ew.log.Errorf("Could not wire event: %v\n", err)
+		return
+	}
+
+	err = ew.target.Send(&eventstreamapi.Event{Event: pev})
```
jgwest, line 424: Can add this comment which was moved from elsewhere in the code:
```
	// A Send() on the stream is actually not blocking.
```

principal/apis/eventstream/eventstream.go
```diff
@@ -189,6 +195,17 @@ func (s *Server) recvFunc(c *client, subs eventstreamapi.EventStream_SubscribeSe
 		return fmt.Errorf("panic: no recvq for agent %s", c.agentName)
 	}
 
+	if event.Target(incomingEvent) == event.TargetEventAck {
+		logCtx.Tracef("Received an ACK: resourceID %s eventID %s", event.ResourceID(incomingEvent), event.EventID(incomingEvent))
+		eventWriter, exists := s.eventWriters[c.agentName]
+		if !exists {
+			return fmt.Errorf("panic: event writer not found for agent %s", c.agentName)
```
jgwest, line 247: Errors returned from this function will cause the entire client <-> agent connection/event processing logic to terminate. Is that what we want in this case?

principal/apis/eventstream/eventstream.go
```diff
@@ -189,6 +195,17 @@ func (s *Server) recvFunc(c *client, subs eventstreamapi.EventStream_SubscribeSe
 		return fmt.Errorf("panic: no recvq for agent %s", c.agentName)
 	}
 
+	if event.Target(incomingEvent) == event.TargetEventAck {
+		logCtx.Tracef("Received an ACK: resourceID %s eventID %s", event.ResourceID(incomingEvent), event.EventID(incomingEvent))
+		eventWriter, exists := s.eventWriters[c.agentName]
```
jgwest: As above, think a mutex is needed for this.

internal/event/event.go
```diff
@@ -148,8 +210,204 @@ func (ev Event) Application() (*v1alpha1.Application, error) {
...
+	ew.mu.Lock()
+	defer ew.mu.Unlock()
+
+	// Remove the event only if it matches both the resourceID and eventID.
+	resourceID := ResourceID(ev)
+	latestEvent, exists := ew.latestEvents[resourceID]
+	if !exists {
+		return
+	}
+
```
jgwest, line 343: Need to acquire `latestEvent`'s mutex here, since we are reading from it

internal/event/event.go
```diff
@@ -148,8 +210,204 @@ func (ev Event) Application() (*v1alpha1.Application, error) {
...
+	if EventID(latestEvent.event) == EventID(ev) {
+		delete(ew.latestEvents, resourceID)
+	}
+}
+
+func (ew *EventWriter) SendWaitingEvents(ctx context.Context) {
```
jannfis, line 356: Should probably be noted that this function will never return unless the context is done, and therefore should be started in a go routine.

jannfis (commented):
Looks pretty good to me, except for the logging.

Going forward, we may want to augment the event writer with metrics and potentially rate limiting or a size cap. But that's something for different PRs.

principal/apis/eventstream/eventstream.go
```diff
@@ -225,26 +242,15 @@ func (s *Server) sendFunc(c *client, subs eventstreamapi.EventStream_SubscribeSe
 		return fmt.Errorf("panic: invalid data in sendqueue: want: %T, have %T", cloudevents.Event{}, item)
 	}
 
-	prEv, err := format.ToProto(ev)
-	if err != nil {
-		return fmt.Errorf("panic: could not serialize event to wire format: %v", err)
+	eventWriter, exists := s.eventWriters[c.agentName]
+	if !exists {
+		return fmt.Errorf("panic: event writer not found for agent %s", c.agentName)
```
jgwest, line 287: At the moment, it appears that if principal attempts to send a message to agent, before agent has connected, the message will be discarded.

This causes a problem in this case, for example:
1) Principal process stops.
2) User makes changes to one or more Argo CD Application/AppProjects/etc.
3) Principal process starts.
4) Principal process starts calls 'StartInformer' to setup K8s watch and start processing K8s events
5) Principal receives events for all watched resources from informer, e.g. Applications, and attempts to send the new Application state (from step 2) to agent
    - But agent hasn't connected yet, so the events are discarded.
6) Agent connects.
    - But agent missed the initial Application updates, above, so no new messages are sent.

You can see an example of this in `principal/apis/eventstream/eventstream.go` in `sendFunc`, where:
* If there is no send queue (outbox) for an agent (by name), the event is discarded and we stop processing events
* If there IS a send queue, but not an eventWriter for that agent (by name), we likewise discard the event and stop processing events.

AFAICT the solution to this is either:
* A) we keep agent events in the queue, even if the agent hasn't connected yet (don't discard them).
    * The only issue is how to know what the name of the agent is before it connects, and I presume we can get this from the namespace on principal side (in the managed agent case)
* or, B) when an agent first connects (after principal OS process startup), we manually re-queue all events for all resources in the agent namespace on principal (e.g. we cause all the events to be re-reconciled by the informer, which cause messages to be sent containing the contents).

My personal preference is A).

internal/event/event.go
```diff
@@ -85,28 +107,66 @@ func IsEventDiscarded(err error) bool {
...
+	cev.SetExtension(eventID, createEventID(app.ObjectMeta))
+	cev.SetExtension(resourceID, createResourceID(app.ObjectMeta))
 	cev.SetDataSchema(TargetApplication.String())
 	// TODO: Handle this error situation?
 	_ = cev.SetData(cloudevents.ApplicationJSON, app)
 	return &cev
 }
 
+func createResourceID(res v1.ObjectMeta) string {
+	return fmt.Sprintf("%s_%s", res.Name, res.UID)
```
jgwest, line 136: Currently, in this PR, the UID is taken from `.metadata`. In order to ensure consistency between source of truth (e.g. principal) and peer (e.g. managed agent), we need to use the UID from the source of truth on both sides:
- Resources events are sent between principal <-> agent
- The UID of the resource (application/appproject) is now included as part of the event, as part of this PR.
- However, the uid value MUST be the same on both the principal and agent sides.
- Our logic would be: when the agent/principal sees an event for a resource (e.g. application) with a matching name, but without have a matching uid, our logic should be to assume that the object with the old uuid was deleted.
- At present, (correct me if wrong!) this PR is using `.metadata.UID` for both principal and agent sides. This is a problem (likely that we will hit later), as this value will be different on both sides.
    - The `.metadata.uid` value is generated by K8s when the object is created.
    - However this means it is generated twice: one on the principal side when the Application is created in the principal namespace, and again on the agent side, when the agent completes the Application sync from principal.
    - Since they are being created on separate clusters, and the UID value is not used on creation, the values will be different.
- The canonical UID for a resource should come from the source of truth for that resource:
    - For managed agent, the `.metadata.UID` on principal is the source of truth
    - For autonomous agent, the `.metadata.UID` on agent is the source of truth.
- So, we can still use `.metadata.UID` when sending events from source of truth (principal), but the non-source of truth peer needs a way to know what UUID to use.

### Proposed Solution:
- For this, I recommend we add an annotation like `application-uid: (uid from principal)` to the resource on the agent (e.g. non source of truth) side.
- Then, when the agent is processing an event, it can use this uuid judge whether or not the event from principal still applies (and vice versa)
    - If the UID matches, the event still applies. If the UID doesn't match, the event no longer applies.

jgwest (changes requested):
Looks great @chetan-rns, nice work! 

In addition to some suggestions re: logic/mutex, I've added some feedback around 1) that we need to use UID value from source of truth, and 2)  not discarding messages if agent hasn't connected yet.
- IMHO these are both within scope of this PR/issue, but if you feel that these are out of scope for the issue, or that you would prefer to work on them as part of a separate PR, that's fine with me!

agent/connection.go
chetan-rns: Updated.

principal/apis/eventstream/eventstream.go
chetan-rns: Added a mutex.

principal/apis/eventstream/eventstream.go
chetan-rns: We don't expect eventWriter to be nil at this stage since we always initialize it when the agent invokes `Subscribe()`. If it is nil for some unexpected reason, we error and terminate the existing connection, forcing the agent to reconnect and thereby initializing a new eventWriter.  But this should never happen since we remove the eventWriter only when the stream is closed.

principal/apis/eventstream/eventstream.go
chetan-rns: Thanks for the detailed explanation. Since there are multiple ways of implementing this, we should probably address it in a separate PR to ensure that we don't miss any corner cases. I think we are also tracking this issue here https://github.com/argoproj-labs/argocd-agent/issues/94. WDYT?

internal/event/event.go
chetan-rns: Currently, we use UIDs in this PR to handle ACKs. We add the `.metadata.UID` of the resources from the source of truth, and the same UID is used on the other side. For example, in the case of managed mode:
1. When an event occurs on the Principal, we attach a resource's `metadata.UID` to an event and send it to the other side.
2. The agent receives this event and processes the event without updating the UID. It sends back an ACK event with the same UID.
4. The principal receives the ACK and verifies if the UID matches.

I think we will still face issues when handling `delete` events and I like the annotation approach for verification. We already keep track of all the managed apps and their resourceVersions using [ApplicationManager](https://github.com/argoproj-labs/argocd-agent/blob/a959b0e4246ac428afd92a8159d00effb96cf27e/internal/manager/application/application.go#L58). Including UID there could be one more option. If it's okay can we address it in a separate PR?

jannfis (approved):
LGTM

jgwest (approved):
LGTM, thanks again @chetan-rns! Outstanding issues will be addressed separately.

PR #227, opened 2024-11-18
fix: write trace level logs to stdout

chetan-rns:
Trace level logs don't show up when running the agent/principal with `--log-level=trace`


chetan-rns:
Fixed in https://github.com/argoproj-labs/argocd-agent/pull/226

PR #229, opened 2024-11-19
Minor fixes for documentation to install principal/agent

anandf:
When installing on an OpenShift clusters, it would be easy if customers can directly install the `argocd-agent` components directly from github without cloning.


jannfis (approved):
LGTM!

PR #230, opened 2024-11-19
chore(build): Upgrade Go to version 1.23

jannfis:


go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/argoproj-labs/argocd-agent
 
-go 1.22.0
-
-toolchain go1.22.3
+go 1.23.0
```
mikeshng, line 3:
```suggestion
go 1.23
```

I am not a go mod expert but is it better to not lock into a `.0` version?

jannfis: It's not a lock-in, it's a minimum version constraint required to build the code.

So in this case `go 1.23` and `go 1.23.0` mean semantically the same thing (i.e. go version 1.23.0 or higher required for build).

ishitasequeira (approved):
LGTM!!

PR #232, opened 2024-11-20
refactor: New generic informer

jannfis:
This is a major refactor of the informer used by the agent and the principal.

Instead of having specialized informers that use a common base, we now have a single Informer that uses Go generics and can be instantiated for arbitrary resource kinds.

The result is much less code to maintain, and way less repetitions. It allows us to implement informers for other resource types fairly easily. On the flip side, the generics syntax is ugly.

It's not yet fully feature complete, as we'll work adding metrics and profiling to the code. But I think it's a good foundation so far.


internal/metrics/metrics.go
```diff
@@ -30,6 +30,14 @@ type AppProjectMetrics struct {
 	AppProjectWatcherMetrics
 }
 
+type InformerMetrics struct {
+	ResourcesListed *prometheus.GaugeVec
+	ListDuration    *prometheus.GaugeVec
+	AddDuration     *prometheus.GaugeVec
+	UpdateDuration  *prometheus.GaugeVec
+	DeleteDuration  *prometheus.GaugeVec
+}
```
jannfis, line 39: This slipped in, but is not used right now (except in some tests I think). It's not yet finished, either.

internal/metrics/metrics.go
```diff
@@ -61,6 +69,20 @@ type AppProjectClientMetrics struct {
...
+			Name: "argocd_agent_informer_list_num_resources",
+			Help: "The number of resources seen by this informer",
+		}, []string{label}),
+		ListDuration: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "argocd_agent_informer_list_duration",
+			Help: "The time it took to list resources (in seconds)",
+		}, []string{label}),
+	}
+	return im
+}
```
jannfis, line 84: This slipped in, but is not used right now (except in some tests I think). It's not yet finished, either.

agent/agent.go
```diff
@@ -173,14 +203,25 @@ func (a *Agent) Start(ctx context.Context) error {
...
+			log().Info("Application backend has exited")
+		}
 	}()
+
+	// Start the AppProject backend in the background
 	go func() {
-		a.projectManager.StartBackend(a.context)
-		log().Warnf("Project Informer has exited")
+		if err := a.projectManager.StartBackend(a.context); err != nil {
+			log().WithError(err).Error("Application backend has exited non-successfully")
```
chetan-rns:
```suggestion
			log().WithError(err).Error("AppProject backend has exited non-successfully")
```

agent/agent.go
```diff
@@ -120,30 +123,55 @@ func NewAgent(ctx context.Context, appclient appclientset.Interface, namespace s
...
-	appInformer := appinformer.NewAppInformer(ctx, appclient, a.namespace,
-		appinformer.WithListAppCallback(a.listAppCallback),
-		appinformer.WithNewAppCallback(a.addAppCreationToQueue),
-		appinformer.WithUpdateAppCallback(a.addAppUpdateToQueue),
-		appinformer.WithDeleteAppCallback(a.addAppDeletionToQueue),
-		appinformer.WithFilterChain(a.DefaultFilterChain()),
-	)
+	// appListFunc and watchFunc are anonymous functions for the informer
+	appListFunc := func(ctx context.Context, opts v1.ListOptions) (runtime.Object, error) {
+		return appclient.ArgoprojV1alpha1().Applications(a.namespace).List(ctx, opts)
```
chetan-rns, line 128: Should we call `Manage()` and `IgnoreChange()` on the listed apps?  I noticed that we invoke them after creating the resource in `ApplicationManager.Create()`. But I wonder if we need the `managedApps` and `observedApps` populated when the informer is initialized.

agent/outbound.go
```diff
@@ -21,23 +21,23 @@ import (
...
-		if err := a.appManager.Manage(app.QualifiedName()); err != nil {
-			log().Warnf("Could not manage app %s: %v", app.QualifiedName(), err)
-		}
-		if err := a.appManager.IgnoreChange(app.QualifiedName(), app.ResourceVersion); err != nil {
-			log().Warnf("Could not ignore change %s for app %s: %v", app.ResourceVersion, app.QualifiedName(), err)
-		}
-	}
-	return apps
-}
+// func (a *Agent) listAppCallback(apps []v1alpha1.Application) []v1alpha1.Application {
```
chetan-rns: Do we plan to use this function again? or should we remove it?

principal/server.go
```diff
@@ -131,56 +137,70 @@ func NewServer(ctx context.Context, kubeClient *kube.KubernetesClient, namespace
...
 	}
 
 	if s.options.metricsPort > 0 {
-		appInformerOptions = append(appInformerOptions, appinformer.WithMetrics(metrics.NewApplicationWatcherMetrics()))
-		managerOpts = append(managerOpts, application.WithMetrics(metrics.NewApplicationClientMetrics()))
-		appProjectInformerOptions = append(appProjectInformerOptions, appprojectinformer.WithMetrics(metrics.NewAppProjectWatcherMetrics()))
-		appProjectManagerOption = append(appProjectManagerOption, appproject.WithMetrics(metrics.NewAppProjectClientMetrics()))
+		appInformerOpts = append(appInformerOpts, informer.WithMetrics[*v1alpha1.Application](prometheus.NewRegistry(), metrics.NewInformerMetrics("applications")))
+		appManagerOpts = append(appManagerOpts, application.WithMetrics(metrics.NewApplicationClientMetrics()))
+		projInformerOpts = append(projInformerOpts, informer.WithMetrics[*v1alpha1.AppProject](prometheus.NewRegistry(), metrics.NewInformerMetrics("applications")))
```
chetan-rns: Should the label be `appProjects` instead of `applications`?

chetan-rns (commented):
Looks good to me. Added a couple of small comments. It's good that we are removing repetitive code. It'll take some time to get used to the generic code style. Hehe :)

agent/outbound.go
jannfis: Oops. This should be gone indeed :)

principal/server.go
jannfis: Good catch!

agent/agent.go
jannfis: Actually a good thought, but I think the previous iteration, i.e. having these calls in the lister, was a bug.

All items that are returned by the list function and pass the (new) informer's filter will implicitly go through the handler called for adding a resource, and the code for claiming an app to be managed is already in that handler. Those that do not pass the filter will also not be managed.

So I think the list callback is correct as is.

ishitasequeira (approved):
Overall, the changes look good. Removing the redundant code will significantly improve maintainability. However, like Chetan said, it will take some time in getting accustomed to the use of generics.

chetan-rns (approved):

PR #235, opened 2024-11-20
chore(deps): Bump Argo CD to v2.13.1

jannfis:
Replaces #233, because updates of replaces in the go.mod are required.


.golangci.yaml
```diff
@@ -9,3 +9,5 @@ linters:
 issues:
   max-issues-per-linter: 0
   max-same-issues: 0
+  exclude:
+  - SA1019 # temporarily ignore deprecation warnings
```
jannfis, line 13: Temporarily silence the linter for deprecations, because some k8s workqueue interfaces are deprecated.

Will re-enabled once we switched to the new interfaces, but this needs to be a separate change.

anandf (approved):
LGTM

ishitasequeira (approved):

PR #236, opened 2024-11-21
chore: Move to new typed K8s workqueues

jannfis:
Move from the deprecated `RateLimitingInterface` to `TypedRateLimitingInterface`.

We also re-enable deprecation warnings for the linter that have been temporarily disabled with https://github.com/argoproj-labs/argocd-agent/pull/235


ishitasequeira (approved):
LGTM!!

PR #239, opened 2024-11-30
chore(demo-env): Use local argo-cd manifests for speeding up kustomize

jannfis:
This change shallow clones the Argo CD repository to a temp path and points the kustomize bases to use those, instead of always pulling from github.com.


chetan-rns (approved):
LGTM!

ishitasequeira (approved):
LGTM!!

PR #242, opened 2024-12-03
feat: handle UID mismatch between incoming and existing resources

chetan-rns:

Currently, we don't handle a case where the incoming app/app project could have a different source UID than the resource on the cluster with the same name and namespace. This indicates that the existing resource in the agent was created from a different resource in principal which no longer exists. We annotate the resources in the agent with the source UID to track their source of truth.
For more details: https://github.com/argoproj-labs/argocd-agent/pull/225#discussion_r1851965152
1. An app is present on both the principal and the agent
2. Principal process stops
3. User deletes the app 
4. Principal starts again and the user creates a new app
5. The agent couldn't differentiate between the new and old apps.

Fixes https://github.com/argoproj-labs/argocd-agent/pull/225#discussion_r1851965152

**Checklist**

### DIFF
--- a/agent/inbound.go
+++ b/agent/inbound.go
@@ -55,13 +55,36 @@ func (a *Agent) processIncomingApplication(ev *event.Event) error {
 		return err
 	}
 
+	sourceUIDMatch, err := a.appManager.CompareSourceUID(a.context, incomingApp)
+	if err != nil {
+		return fmt.Errorf("failed to compare the source UID of app: %w", err)
+	}
+
 	switch ev.Type() {
 	case event.Create:
+		if !sourceUIDMatch {
+			logCtx.Debug("An app already exists with a different source UID. Deleting the existing app")
+			if err := a.deleteApplication(incomingApp); err != nil {
+				return err
+			}
+		}
+
 		_, err = a.createApplication(incomingApp)
 		if err != nil {
 			logCtx.Errorf("Error creating application: %v", err)
 		}
 	case event.SpecUpdate:
+		if !sourceUIDMatch {
+			logCtx.Debug("Source UID mismatch between the incoming app and existing app. Deleting the existing app")
+			if err := a.deleteApplication(incomingApp); err != nil {
+				return err
+			}
+
+			logCtx.Debug("Creating the incoming app after deleting the existing app")
+			_, err := a.createApplication(incomingApp)
+			return err
+		}
+
 		_, err = a.updateApplication(incomingApp)
 		if err != nil {
 			logCtx.Errorf("Error updating application: %v", err)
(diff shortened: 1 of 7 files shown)


agent/inbound.go
```diff
@@ -55,13 +55,36 @@ func (a *Agent) processIncomingApplication(ev *event.Event) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to compare the source UID of app: %w", err)
+	}
+
 	switch ev.Type() {
 	case event.Create:
+		if !sourceUIDMatch {
+			logCtx.Debug("An app already exists with a different source UID. Deleting the existing app")
+			if err := a.deleteApplication(incomingApp); err != nil {
+				return err
```
jannfis: I think we should give more context to the returned error here.

The caller will expect an application to be created, and probably reflect that in any error message logged, but could potentially receive an error from the delete operation. 

For example, the following case would be confusing (admittedly, the chance for this happening is very slim, but it should ):

1. A create event is sent for an application where `sourceUIDMatch` is false
2. The application is deleted on the cluster by a third party
3. `a.deleteApplication` fails because the resource already is deleted. The error would be "the requested resource could not be found"
4. That error is passed back to the caller
5. The error log for the caller would look similar to "Could not create app foo: requested resource not found"

Or think about RBAC - agent might have RBAC to create applications, but not delete applications, but the create call would return a permission denied. Troubleshooting that would be fun :)

So it would make more sense to augment the returned error by more information, e.g. `fmt.Errorf("could not delete existing resource prior to creation: %w", err)` or something similar.

agent/inbound.go
```diff
@@ -55,13 +55,36 @@ func (a *Agent) processIncomingApplication(ev *event.Event) error {
...
+
 		_, err = a.createApplication(incomingApp)
 		if err != nil {
 			logCtx.Errorf("Error creating application: %v", err)
 		}
 	case event.SpecUpdate:
+		if !sourceUIDMatch {
+			logCtx.Debug("Source UID mismatch between the incoming app and existing app. Deleting the existing app")
+			if err := a.deleteApplication(incomingApp); err != nil {
+				return err
```
jannfis: Same about augmenting the error as above.

agent/inbound.go
```diff
@@ -87,13 +110,36 @@ func (a *Agent) processIncomingAppProject(ev *event.Event) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to validate source UID of appProject: %w", err)
+	}
+
 	switch ev.Type() {
 	case event.Create:
+		if !sourceUIDMatch {
+			logCtx.Debug("An appProject already exists with a different source UID. Deleting the existing appProject")
+			if err := a.deleteAppProject(incomingAppProject); err != nil {
+				return err
```
jannfis: Same about augmenting the error as above.

agent/inbound.go
```diff
@@ -87,13 +110,36 @@ func (a *Agent) processIncomingAppProject(ev *event.Event) error {
...
+
 		_, err = a.createAppProject(incomingAppProject)
 		if err != nil {
 			logCtx.Errorf("Error creating appproject: %v", err)
 		}
 	case event.SpecUpdate:
+		if !sourceUIDMatch {
+			logCtx.Debug("Source UID mismatch between the incoming and existing appProject. Deleting the existing appProject")
+			if err := a.deleteAppProject(incomingAppProject); err != nil {
+				return err
```
jannfis: Same about augmenting the error as above.

agent/inbound.go
```diff
@@ -287,7 +334,7 @@ func (a *Agent) deleteAppProject(project *v1alpha1.AppProject) error {
 	// means that we're out-of-sync from the control plane.
 	//
 	// TODO(jannfis): Handle this situation properly instead of throwing an error.
-	if !a.appManager.IsManaged(project.Name) {
+	if !a.projectManager.IsManaged(project.Name) {
```
jannfis, line 341: :+1:

jannfis (commented):
Thanks!

I have some minor comments about error handling, otherwise this looks good to go.

agent/inbound.go
chetan-rns: Makes sense. Added more context to errors.

agent/inbound.go
chetan-rns: Updated

agent/inbound.go
chetan-rns: Updated

agent/inbound.go
chetan-rns: Updated

jannfis (approved):
LGTM

Issue #243, opened 2024-12-03
Repeated deliveries of invalid events


jannfis:
**Describe the bug**

When an unknown or invalid event is transmitted between agent and principal, it is never acknowledged. For example, if the principal sends an event to the agent that the agent cannot process, the agent does not acknowledge it. That leads the principal to keep sending the event endlessly.

**Steps to reproduce the behaviour**

1. Send an event of unknown target or invalid type to the receiver
2. Observe error message on the receiving side
3. Observe repeated tries to deliver the event

**Expected behavior**

On error, the receiving side should let the sending side know that it has received the event but that it deemed it invalid and could not process it. The sender should then stop sending the event.

**Additional context**

PR #254, opened 2024-12-12
fix: resend events only for retryable errors

chetan-rns:

* Check if the error is retryable/transient.
* Don't send an ACK for retryable errors so that the principal can resend events.

Fixes https://github.com/argoproj-labs/argocd-agent/issues/243

**Checklist**


jannfis (approved):
LGTM! Thanks.

PR #271, opened 2025-01-07
feat: store the informer events even if the agent is not connected

chetan-rns:

See https://github.com/argoproj-labs/argocd-agent/pull/225#discussion_r1851955435 for more details.

Currently, the principal removes the queue when the agent disconnects and drops all the incoming events from the informer. When the agent connects back it may miss the updates and be out of sync from the principal.  For example:

1. The principal and the agent are in sync
2. Agent disconnects
3. The user updates the application. But the principal discards them since there is no agent
4. The agent connects back and is now out of sync with the principal.

There are two solutions to this problem:
1. The principal shouldn't remove the queue and reuse it when the agent connects again. The principal will just send the events from the queue when the agent reconnects.
2. The principal could list all the latest resources from the cluster and send them to the agent.

This PR explores solution 1. However, it doesn't handle orphaned resources. For example, if a resource is deleted when the principal is down it cannot transmit this event to the agent and the resource on the agent will be orphaned.

Changes introduced in this PR:
1. Keep the queue open even if the agent is not connected and store the informer events
2. Since the namespace name matches the agent ID, we remove the queue when the namespace associated with the agent is deleted.

Fixes https://github.com/argoproj-labs/argocd-agent/pull/225#discussion_r1851955435

**Checklist**

### DIFF
--- a/internal/queue/queue_test.go
+++ b/internal/queue/queue_test.go
@@ -15,6 +15,7 @@
 package queue
 
 import (
+	"strconv"
 	"testing"
 
 	"github.com/cloudevents/sdk-go/v2/event"
@@ -66,11 +67,16 @@ func Test_Queue(t *testing.T) {
 
 		for i := 1; i <= defaultMaxQueueSize; i++ {
 			ev := event.New()
+			ev.SetID(strconv.Itoa(i))
 			queue.Add(&ev)
 		}
 
-		// Since the queue is full, check if it is emptied before adding a new item.
-		queue.Add(&event.Event{})
-		assert.Equal(t, 1, queue.Len())
+		// Since the queue is full, check if the oldest item is popped before adding a new item.
+		ev := event.New()
+		ev.SetID("1001")
+		queue.Add(&ev)
+		assert.Equal(t, defaultMaxQueueSize, queue.Len())
+		front, _ := queue.Get()
+		assert.Equal(t, "2", front.ID())
 	})
 }
--- a/internal/queue/queue_test.go
+++ b/internal/queue/queue_test.go
@@ -17,6 +17,7 @@ package queue
 import (
 	"testing"
 
+	"github.com/cloudevents/sdk-go/v2/event"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -55,5 +56,19 @@ func Test_Queue(t *testing.T) {
 		assert.NoError(t, err)
 		err = q.Delete("agent1", true)
 		assert.Error(t, err)
+	})
+
+	t.Run("Ensure that the max queue size is respected", func(t *testing.T) {
+		q := NewSendRecvQueues()
+		err := q.Create("agent1")
+		assert.NoError(t, err)
+		queue := q.RecvQ("agent1")
+
+		for i := 1; i <= defaultMaxQueueSize+1; i++ {
+			ev := event.New()
+			queue.Add(&ev)
+		}
+
+		assert.Equal(t, 1, queue.Len())
 	})
 }
--- a/internal/queue/queue_test.go
+++ b/internal/queue/queue_test.go
@@ -64,11 +64,13 @@ func Test_Queue(t *testing.T) {
 		assert.NoError(t, err)
 		queue := q.RecvQ("agent1")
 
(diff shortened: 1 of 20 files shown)


chetan-rns:
> I have not yet looked at the code, but can this lead to potential unlimited queue growth when there is no consumer for the particular queue?

That's a valid concern. The logic in this PR assumes that the queue is required as long as an agent namespace (same as the agent ID) is present on the control plane. We remove the queue if the user has deleted the agent namespace thereby deleting apps on the principal. AFAIU the unlimited queue growth may only happen when the agent is permanently gone and the user being unaware is continuously updating the apps on the control plane. I'm assuming this happens rarely because if there is no agent the user might also delete the agent namespace or wouldn't update the apps or the agent is down for a short period.

principal/server.go
```diff
@@ -274,6 +277,37 @@ func (s *Server) Start(ctx context.Context, errch chan error) error {
...
+		case <-ticker.C:
+			queues := s.queues.Names()
+			for _, queueName := range queues {
+				s.removeQueueIfUnused(ctx, queueName)
+			}
+		}
+	}
+}
+
+func (s *Server) removeQueueIfUnused(ctx context.Context, agentName string) {
```
jannfis: I think the better solution would be to remove the queue when the agent (in case of Kubernetes backend: namespace) is deleted., instead of executing on an interval.

internal/queue/queue_test.go
```diff
@@ -56,4 +57,18 @@ func Test_Queue(t *testing.T) {
...
+		err := q.Create("agent1")
+		assert.NoError(t, err)
+		queue := q.RecvQ("agent1")
+
+		for i := 1; i <= defaultMaxQueueSize+1; i++ {
+			ev := event.New()
+			queue.Add(&ev)
+		}
+
+		assert.Equal(t, 1, queue.Len())
```
jannfis: Shouldn't `queue.Len()` be `defaultMaxQueueSize` here?

principal/server.go
chetan-rns: Updated. I have included a simple namespace informer using the generic informer in the backend module.

principal/apis/eventstream/eventstream.go
```diff
@@ -271,6 +283,19 @@ func (s *Server) sendFunc(c *client, subs eventstreamapi.EventStream_SubscribeSe
 		return fmt.Errorf("panic: nil item in queue")
 	}
 
+	mode, err := agentMode(c.ctx)
+	if err != nil {
+		return fmt.Errorf("unable to extract the agent mode from context")
```
jannfis: I think a more specific error message would make more sense here, e.g.

```suggestion
		return fmt.Errorf("unable to determine agent mode: %w", err)
```

principal/auth.go
```diff
@@ -118,7 +118,8 @@ func (s *Server) authenticate(ctx context.Context) (context.Context, error) {
 
 	// claims at this point is validated and we can propagate values to the
 	// context.
-	authCtx := session.ClientIdToContext(ctx, agentInfo.ClientID)
+	authCtx := context.WithValue(session.ClientIdToContext(ctx, agentInfo.ClientID),
+		types.ContextAgentMode, agentInfo.Mode)
```
jannfis: This IMHO screams for a slight adaption to `session.ClientIdToContext()`, e.g. it should be changed to `ClientInfoToContext` and take both, the agent's ID and mode as arguments. Then this code would merely be

```golang
authCtx := session.ClientInfoToContext(ctx, agentInfo.ClientID, agentInfo.Mode)
```

And it makes it more clear that both, the agent ID and the agent mode are mandatory for the context.

internal/queue/queue_test.go
chetan-rns: We added `defaultMaxQueueSize + 1` items in the loop above to check if the queue would get drained before adding a new item. I've updated the code to make it clear.

principal/auth.go
chetan-rns: I've refactored the code.

jannfis (approved):
LGTM

Issue #155, opened 2024-08-14
Export "average connection time for agents" as metric


jannfis:
The average time of connection time to the principal, exported as a metric.

This should be calculated as an average across all agent connections.

Issue #154, opened 2024-08-14
Export "number of connected agents" as metric


jannfis:
On the principal, it would be great to export a metric that has the number of currently connected agents.

Issue #153, opened 2024-08-14
Make metrics actually work


jannfis:
We do have a metrics server and some stub metrics [here](https://github.com/argoproj-labs/argocd-agent/tree/main/internal/metrics) which are used by both, agent and principal.

However, the implementation right now is very bare bones and this functionality needs to be revisited.

The current available metrics are:

* Number of apps created
* Number of apps updated
* Number of apps deleted
* Number of errors encountered

We must ensure that these metrics are counted and exported correctly on both, agent and principal.

PR #278, opened 2025-01-16
feat: Improve Argo Agent Metrics

jparsai:
This PR is to improve metrics. The metrics could be seen locally using below links:

- Principal metrics :- http://0.0.0.0:8000/metrics
- Managed agent metrics :- http://0.0.0.0:8181/metrics
- Autonomous agent metrics :- http://0.0.0.0:8182/metrics

Fixes #?
https://github.com/argoproj-labs/argocd-agent/issues/153
https://github.com/argoproj-labs/argocd-agent/issues/154
https://github.com/argoproj-labs/argocd-agent/issues/155


jgwest:
@jparsai  Two things I noticed:
1) The `metrics` field of the `Informer` struct in  `internal/informer/informer.go`, is not used. Your code is calling `WithMetrics`, but this isn't actually doing anything, because the value of that field is not used anywhere else.
2) I noticed that `StartMetricsServer` is only called from client, not from agent.

jparsai:
> @jparsai Two things I noticed:
> 
> 1. The `metrics` field of the `Informer` struct in  `internal/informer/informer.go`, is not used. Your code is calling `WithMetrics`, but this isn't actually doing anything, because the value of that field is not used anywhere else.
> 2. I noticed that `StartMetricsServer` is only called from client, not from agent.

The `metrics` field of the `Informer` will be used for informer related data collection, it is initialized, but as of now informer data is not being collected. It was like that in existing metrics implementation for server, so I used it in agent as well, I can remove that and at the time of informer data collection we will have to put it back.

cmd/agent/main.go
```diff
@@ -166,6 +174,12 @@ func NewAgentRunCommand() *cobra.Command {
 	command.Flags().BoolVar(&enableWebSocket, "enable-websocket",
 		env.BoolWithDefault("ARGOCD_AGENT_ENABLE_WEBSOCKET", false),
 		"Agent will rely on gRPC over WebSocket to stream events to the Principal")
+	command.Flags().IntVar(&metricsPort, "metrics-port",
+		env.NumWithDefault("ARGOCD_AGENT_METRICS_PORT", cmd.ValidPort, 8181),
+		"Port the metrics server will listen on")
+	command.Flags().BoolVar(&disableMetrics, "disable-metrics",
+		env.BoolWithDefault("ARGOCD_AGENT_DISABLE_METRICS", false),
+		"Disable metrics collection and metrics server")
```
jannfis: I'm not in favor of a dedicated `--disable-metrics` option. IMO, it'd be better to disable metrics if metrics port is set to `0`.

agent/inbound.go
```diff
@@ -42,6 +42,10 @@ were received from a server.
 const defaultResourceRequestTimeout = 5 * time.Second
 
 func (a *Agent) processIncomingEvent(ev *event.Event) error {
+
+	start := time.Now()
+	status := "success"
```
jannfis: I was wondering if we should use constants for the outcome of an operation. Potentially define them in the metrics package.

agent/inbound.go
```diff
@@ -42,6 +42,10 @@ were received from a server.
 const defaultResourceRequestTimeout = 5 * time.Second
 
 func (a *Agent) processIncomingEvent(ev *event.Event) error {
+
+	start := time.Now()
```
jannfis: PTAL at the checkpoint package for measuring time.

principal/event.go
```diff
@@ -39,6 +39,8 @@ import (
 // events received by agents. It will trigger updates of resources in the
 // server's backend.
 func (s *Server) processRecvQueue(ctx context.Context, agentName string, q workqueue.TypedRateLimitingInterface[*cloudevents.Event]) (*cloudevents.Event, error) {
+	start := time.Now()
+	status := "success"
```
jannfis: See previous comment about constants.

internal/metrics/metrics.go
```diff
@@ -15,20 +15,14 @@
...
-	ApplicationWatcherMetrics
-	ApplicationClientMetrics
-}
-
-type AppProjectMetrics struct {
-	AppProjectClientMetrics
-	AppProjectWatcherMetrics
-}
+// ConnectionTimeMap is an utility for AvgAgentConnectionTime
+var ConnectionTimeMap = make(map[string]time.Time)
```
jannfis: This map will need a lock, because it is used by concurrent threads.

jannfis (commented):
First round of reviews. Have a couple of comments. PTAL.

jannfis:
BTW, I do not think this PR should be classified a `chore`. It really is a `feat` :)

jparsai:
Thanks @jannfis for review, addressed your comments PTAL.

internal/metrics/metrics.go
```diff
@@ -83,111 +82,139 @@ func NewInformerMetrics(label string) *InformerMetrics {
...
+	ConnectionTimeMap: make(map[string]time.Time),
+}
 
-// func (am *ApplicationWatcherMetrics) AppUpdated() {
-// 	am.AppsUpdated.Inc()
-// }
+// SetAgentConnectionTime inserts connection time of new agent
+func SetAgentConnectionTime(agentName string, start time.Time) {
+	agentConnectionTime.Lock.RLock()
+	defer agentConnectionTime.Lock.RUnlock()
```
jannfis: Should use a write lock, not a read lock, when accessing maps for writing

internal/metrics/metrics.go
```diff
@@ -83,111 +82,139 @@ func NewInformerMetrics(label string) *InformerMetrics {
...
+	agentConnectionTime.ConnectionTimeMap[agentName] = start
+}
 
-// func (am *ApplicationWatcherMetrics) Error() {
-// 	am.Errors.Inc()
-// }
+// DeleteAgentConnectionTime removed connection time of existing agent
+func DeleteAgentConnectionTime(agentName string) {
+	agentConnectionTime.Lock.RLock()
+	defer agentConnectionTime.Lock.RUnlock()
```
jannfis: Same here, should be a write lock.

jannfis (commented):
Nice, thanks!

Got two more comments.

internal/metrics/metrics.go
jparsai: Ah :(, fixed it.

jannfis (commented):
LGTM, thanks!

@jgwest / @chetan-rns PTAL as well. Gonna merge this PR on Friday otherwise.

jannfis (approved):

PR #282, opened 2025-01-28
feat: CLI to manipulate agent configuration

jannfis:

This PR introduces a new CLI for argocd-agent, which can be used to manipulate and inspect agent configuration on the principal. 

Right now, it will be able to:

* perform CRUD operations on a TLS CA used by the principal to maintain TLS certificates for clients of the resource proxy,
* manipulate and inspect Argo CD cluster secrets so the Argo CD UI can make use of the resource proxy

In the future, we may want to extend this functionality beyond the limited use cases we support initially. I will create feature issues for these things.

It also adds some e2e tests for the resource proxy.

Fixes #?

The new CLI is located in `cmd/ctl`. The new commands introduced by this CLI for now are `ca` (for manipulating and inspecting the CA) and `agent` for manipulating and inspecting cluster configuration.

The names of these commands are not finalized, so if you have suggestions, let me know!

Agent configuration for now is an Argo CD cluster secret that points to the resource proxy, instead of to a real cluster.

With this change, the resource proxy can be easily tested. First, create the CA and an agent configuration like so:

```shell
# build the CLI
make cli
# set the kube context to use
export ARGOCD_AGENT_CONTEXT=vcluster-control-plane`
# create a new CA
./dist/argocd-agentctl ca generate
# create a new agent configuration
./dist/argocd-agentctl agent create agent-managed --resource-proxy-server <endpoint_addr_of_rp>:9090
```

The `endpoint_addr_of_rp` must be reachable by Argo CD running in your cluster, so it must NOT be 127.0.0.1 or similar. Since the resource proxy will listen on all addresses, it's best to use the address of one of your interfaces.

Then, create a new application on the principal and set the destination cluster to the one created by the previous command (i.e. `agent-managed`).

The kubeconfig context to use with the agent can be specified by either setting the environment variable `ARGOCD_AGENT_CONTEXT` or the command line switch `--context` to the name of the context to use. The command line has precedence over the environment variable.

**agentctl - CA related commands**

* `ca generate` - Generate a new CA for use with the agent
* `ca inspect` - Inspect existing CA
* `ca delete` - Delete the CA
* `ca dump` - Dump cert or key of CA to stdout
* `ca issue` - Issue certificates for particular components from the CA

**agentctl - Agent related commands**

* `agent create` - Create a new agent configuration (cluster secret for now)
* `agent list` - List all configured agents (cluster secrets for now)
* `agent inspect` - Inspect a particular agent's configuration
* `agent print-tls` - Print TLS configuration of agent
* `agent reconfigure` - Reconfigure certain aspects of an agent

**Checklist**


hack/dev-env/create-agent-config.sh
```diff
@@ -0,0 +1,54 @@
...
+set -eo pipefail
+
+RECREATE="$1"
+
+SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+BASEPATH="$( cd -- "$(dirname "$0")/../.." >/dev/null 2>&1 ; pwd -P )"
+AGENTCTL=${BASEPATH}/dist/argocd-agentctl
+export ARGOCD_AGENT_CONTEXT=vcluster-control-plane
+
+IPADDR=$(ip r show default | sed -e 's,.*\ src\ ,,' | sed -e 's,\ metric.*$,,')
```
jannfis, line 26: This will be incompatible with MacOS environments. I do not have access to a Mac, so I'm not sure how to come up with a portable solution.

internal/tlsutil/generate.go
```diff
@@ -0,0 +1,199 @@
...
+// configured to be usable as a Certificate Authority (CA). It returns both,
+// the public certificate and the private key in PEM format.
+//
+// The certificate will be valid for 10 days.
+//
+// DO NOT USE THE RESULTING CERTIFICATE OR KEY OR ANY CERTIFICATES SIGNED BY
+// THIS CA FOR PRODUCTION PURPOSES. NEVER. YOU HAVE BEEN WARNED.
+//
+// And sorry for shouting.
+func GenerateCaCertificate(commonName string) (string, string, error) {
```
jgwest, line 43: Nit: `CA` or `Ca`, e.g. `GenerateCaCertificate` or `GenerateCACertificate`?

Not a blocker, just FYI.

(I don't have a preference, my only preference is consistency, but tend to use `CA`-style. I only ask because elsewhere in the package we are ALSO using CA)

jgwest (approved):
Looks perfect, thanks @jannfis! No notes :smile:

Issue #171, opened 2024-09-05
Provide configurable keep-alive interval for event stream


jannfis:
Many network infrastructure components, such as load balancers, ingresses, firewalls etc have a TCP idle timeout that will shut down the connection after some time of network inactivity. This can lead to the agent having to reconnect to the principal when there was no events during that timeframe.

The agent should be able to send a keep-alive packet on the event stream, with a configurable interval, to prevent the connection from being dropped. By default, the keep-alive should be disabled for now.

PR #288, opened 2025-02-04
feat: support periodic ping to principal for keeping connection alive

jparsai:
This PR is to introduce a configurable periodic ping to keep connection alive between agent and principal using [grpc keepalive feature](https://grpc.io/docs/guides/keepalive/). 

By default it is disabled and to enable it you need to set value of `keep-alive-ping-interval` in startup script or set env variable `KEEP_ALIVE_PING_INTERVAL`  at agent side and at principal side you have to set value of `wait-duration-for-agent-ping` in startup script or set value of env variable `WAIT_DURATION_FOR_AGENT_PING` .

Fixes https://github.com/argoproj-labs/argocd-agent/issues/171

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/argoproj-labs/argocd-agent/agent"
 	"github.com/argoproj-labs/argocd-agent/cmd/cmd"
@@ -53,6 +54,10 @@ func NewAgentRunCommand() *cobra.Command {
 		tlsClientKey    string
 		enableWebSocket bool
 		metricsPort     int
+
+		// Time interval for agent to principal ping
+		// Ex: "30m", "1h" or "1h20m10s". Valid time units are "s", "m", "h".
+		keepAlivePingInterval time.Duration
 	)
 	command := &cobra.Command{
 		Short: "Run the argocd-agent agent component",
@@ -101,6 +106,8 @@ func NewAgentRunCommand() *cobra.Command {
 			}
 			remoteOpts = append(remoteOpts, client.WithWebSocket(enableWebSocket))
 			remoteOpts = append(remoteOpts, client.WithClientMode(types.AgentModeFromString(agentMode)))
+			remoteOpts = append(remoteOpts, client.WithKeepAlivePingInterval(keepAlivePingInterval))
+
 			if serverAddress != "" && serverPort > 0 && serverPort < 65536 {
 				remote, err = client.NewRemote(serverAddress, serverPort, remoteOpts...)
 				if err != nil {
@@ -176,6 +183,9 @@ func NewAgentRunCommand() *cobra.Command {
 	command.Flags().IntVar(&metricsPort, "metrics-port",
 		env.NumWithDefault("ARGOCD_AGENT_METRICS_PORT", cmd.ValidPort, 8181),
 		"Port the metrics server will listen on")
+	command.Flags().DurationVar(&keepAlivePingInterval, "keep-alive-ping-interval",
+		env.DurationWithDefault("ARGOCD_AGENT_KEEP_ALIVE_PING_INTERVAL", nil, 0),
+		"Ping interval to keep connection alive with Principal")
 
 	command.Flags().StringVar(&kubeConfig, "kubeconfig", "", "Path to a kubeconfig file to use")
 	command.Flags().StringVar(&kubeContext, "kubecontext", "", "Override the default kube context")
--- a/cmd/principal/main.go
+++ b/cmd/principal/main.go
@@ -67,6 +67,11 @@ func NewPrincipalRunCommand() *cobra.Command {
 		resourceProxyCertPath  string
 		resourceProxyKeyPath   string
 		resourceProxyCAPath    string
+
+		// Minimum time duration for agent to wait before sending next keepalive ping to principal
+		// if agent sends ping more often than specified interval then connection will be dropped
+		// Ex: "30m", "1h" or "1h20m10s". Valid time units are "s", "m", "h".
+		keepAliveMinimumInterval time.Duration
 	)
 	var command = &cobra.Command{
 		Short: "Run the argocd-agent principal component",
@@ -181,6 +186,7 @@ func NewPrincipalRunCommand() *cobra.Command {
 			}
 
 			opts = append(opts, principal.WithWebSocket(enableWebSocket))
+			opts = append(opts, principal.WithKeepAliveMinimumInterval(keepAliveMinimumInterval))
 
 			s, err := principal.NewServer(ctx, kubeConfig, namespace, opts...)
 			if err != nil {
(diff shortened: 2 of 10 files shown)


chetan-rns:
gRPC has support for configuring keepalives. Do you think we should use it?

https://grpc.io/docs/guides/keepalive/

jparsai:
> gRPC has support for configuring keepalives. Do you think we should use it?
> 
> https://grpc.io/docs/guides/keepalive/

@chetan-rns addressed your comment PTAL

cmd/agent/main.go
```diff
@@ -176,6 +182,9 @@ func NewAgentRunCommand() *cobra.Command {
 	command.Flags().IntVar(&metricsPort, "metrics-port",
 		env.NumWithDefault("ARGOCD_AGENT_METRICS_PORT", cmd.ValidPort, 8181),
 		"Port the metrics server will listen on")
+	command.Flags().StringVar(&keepAlivePingInterval, "keep-alive-ping-interval",
```
jannfis: The cobra framework has flags that take duration strings as arguments (`Duration`, `DurationVar`, etc). Using these might remove the need for parsing the duration string yourself.

jparsai: Updated it, but we have to parse for env variable. PTAL

cmd/cmdutil/utils.go
```diff
@@ -0,0 +1,35 @@
...
+package cmdutil
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+func ParseKeepAlivePingInterval(interval string) time.Duration {
```
jannfis: It might make sense to make this function more generic and part of the [internal/env](https://github.com/argoproj-labs/argocd-agent/blob/main/internal/env/env.go) package, so that it streamlines with the rest of the code base.

cmd/principal/main.go
```diff
@@ -275,6 +280,9 @@ func NewPrincipalRunCommand() *cobra.Command {
 	command.Flags().BoolVar(&enableResourceProxy, "enable-resource-proxy",
 		env.BoolWithDefault("ARGOCD_PRINCIPAL_ENABLE_RESOURCE_PROXY", true),
 		"Whether to enable the resource proxy")
+	command.Flags().DurationVar(&waitDurationForAgentPing, "wait-duration-for-agent-ping",
+		cmdutil.ParseKeepAlivePingInterval(env.StringWithDefault("WAIT_DURATION_FOR_AGENT_PING", nil, "")),
```
jannfis: I think `WAIT_DURATION_FOR_AGENT_PING` is not a good name for the environment variable. PTAL at how the other environment variables that we use as arguments are named, and make this one align.

cmd/agent/main.go
```diff
@@ -176,6 +183,9 @@ func NewAgentRunCommand() *cobra.Command {
 	command.Flags().IntVar(&metricsPort, "metrics-port",
 		env.NumWithDefault("ARGOCD_AGENT_METRICS_PORT", cmd.ValidPort, 8181),
 		"Port the metrics server will listen on")
+	command.Flags().DurationVar(&keepAlivePingInterval, "keep-alive-ping-interval",
+		cmdutil.ParseKeepAlivePingInterval(env.StringWithDefault("KEEP_ALIVE_PING_INTERVAL", nil, "")),
```
jannfis: I think `KEEP_ALIVE_PING_INTERVAL` is not a good name for the environment variable. PTAL at how the other environment variables that we use as arguments are named, and make this one align.

jannfis (commented):
I have some comments, and a question: What exactly is the `wait-duration-for-agent-ping` doing, why does it need to be set?

jparsai:
> I have some comments, and a question: What exactly is the `wait-duration-for-agent-ping` doing, why does it need to be set?

It is the value user wants to set for `keepalive.EnforcementPolicy.MinTime`, it is a minimum amount of time a client should wait before sending next keepalive ping. if client sends more than one pings in this window then connection will be closed by server. By default it is set to 5 minutes. It could be ignored/removed, but in case a user configures agent's `keep-alive-ping-interval` flag to less than 5 minutes (which I guess would be rare) then connection will keep on closing, so to be on safe side I thought of asking users to configure this flag as well.

Addressed your review comments PTAL.

internal/env/env.go
```diff
@@ -139,3 +142,16 @@ func StringSlice(key string, validator func(string) error) ([]string, error) {
 	}
 	return ret, nil
 }
+
+func ParseDuration(dr string) time.Duration {
```
jannfis: This should directly parse a duration from the environment, much like the other functions in this package do it for other data types. 

We need two functions: One that returns a default value, and one that doesn't. So the function prototypes should be:

```go
Duration(key string, validator func(time.Duration) error) (time.Duration, error)
```
and
```go
DurationWithDefault(key string, validator func(time.Duration) error, def time.Duration) time.Duration
```

cmd/principal/main.go
```diff
@@ -275,6 +280,9 @@ func NewPrincipalRunCommand() *cobra.Command {
 	command.Flags().BoolVar(&enableResourceProxy, "enable-resource-proxy",
 		env.BoolWithDefault("ARGOCD_PRINCIPAL_ENABLE_RESOURCE_PROXY", true),
 		"Whether to enable the resource proxy")
+	command.Flags().DurationVar(&waitDurationForAgentPing, "wait-duration-for-agent-ping",
+		env.ParseDuration(env.StringWithDefault("ARGOCD_PRINCIPAL_WAIT_DURATION_FOR_AGENT_PING", nil, "")),
```
jannfis: Please take a look at my comment for the new function in the `env` package to see what should be changed here.

cmd/principal/main.go
```diff
@@ -275,6 +280,9 @@ func NewPrincipalRunCommand() *cobra.Command {
 	command.Flags().BoolVar(&enableResourceProxy, "enable-resource-proxy",
 		env.BoolWithDefault("ARGOCD_PRINCIPAL_ENABLE_RESOURCE_PROXY", true),
 		"Whether to enable the resource proxy")
+	command.Flags().DurationVar(&waitDurationForAgentPing, "wait-duration-for-agent-ping",
```
jannfis: With the explanation that you gave me for the purpose of this switch, I think the phrase "wait duration" might be a bit misleading here. To me, this phrase implies that the principal waits some time for the agent ping, when in reality it enforces agent pings to not being sent at a shorter interval than the value given. Good command line option names should be more or less self-explaining.

So for this, `enforce-min-keepalive` or `keepalive-min-interval` would be two options that come to my mind. I'm open for suggestions of other names.

cmd/principal/main.go
```diff
@@ -275,6 +280,9 @@ func NewPrincipalRunCommand() *cobra.Command {
...
+		env.ParseDuration(env.StringWithDefault("ARGOCD_PRINCIPAL_WAIT_DURATION_FOR_AGENT_PING", nil, "")),
+		"Minimum time duration agent should wait before next ping to principal") // It should be less than "keep-alive-ping-interval" of agent
```
jannfis: Since this option will have dire consequences for agents (e.g. their connections will be terminated if they don't abide by this enforcement rule), we should call that out in the option's description. For example, `Drop agent connections that send keepalives more often than the specified interval`

principal/options.go
```diff
@@ -377,3 +377,10 @@ func WithResourceProxyTLS(tlsConfig *tls.Config) ServerOption {
 		return nil
 	}
 }
+
+func WithWaitDurationForAgentPing(interval time.Duration) ServerOption {
```
jannfis: Similar comment as for the command line. This should be renamed to be more in line with what it does.

jannfis (commented):
We're coming close :)

internal/env/env.go
jparsai: oh, my bad, I misunderstood previous comment and didn't check implementation of other functions for reference.
Addressed all your comments, PTAL

internal/env/env.go
```diff
@@ -139,3 +140,32 @@ func StringSlice(key string, validator func(string) error) ([]string, error) {
 	}
 	return ret, nil
 }
+
+func Duration(key string) (time.Duration, error) {
```
jannfis: The `validator` argument is missing from this func, as well as its execution.

jannfis (commented):
Cool. I have one additional thing, and also unit tests for the new code in the env package is missing (it's perfectly testable).

internal/env/env.go
jparsai: ah, in hurry I missed it :(

PR #301, opened 2025-02-12
fix: Unmanage app on agent on deletion event

jannfis:

When an app is deleted on the agent, it was not unmanaged. That prevented the agent from creating a new app with the same name again.

In addition, this PR adds unit tests for the agent's informer callbacks.

Fixes #?

**Checklist**

### DIFF
--- a/agent/outbound.go
+++ b/agent/outbound.go
@@ -97,7 +97,9 @@ func (a *Agent) addAppDeletionToQueue(app *v1alpha1.Application) {
 	logCtx.Debugf("Delete app event")
 
 	if !a.appManager.IsManaged(app.QualifiedName()) {
-		logCtx.Tracef("App is not managed")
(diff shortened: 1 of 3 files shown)


anandf (approved):
LGTM

agent/outbound.go
```diff
@@ -97,7 +97,9 @@ func (a *Agent) addAppDeletionToQueue(app *v1alpha1.Application) {
 	logCtx.Debugf("Delete app event")
 
 	if !a.appManager.IsManaged(app.QualifiedName()) {
-		logCtx.Tracef("App is not managed")
+		logCtx.Tracef("App is not managed, proceeding anyways")
```
anandf, line 100: nitpick: `logCtx.Trace` can be used as there are no format specifiers.

chetan-rns (approved):
LGTM

PR #309, opened 2025-02-20
feat: agent and principal should resync on start up

chetan-rns:

Currently, the agent and the principal might get out of sync when they miss events that occur when they are down. Jonathan proposed an [algorithm](https://docs.google.com/document/d/1NwQaYxmRva8irx_clPTqRu_YQ8MNeFNXSTJoOgv7ZRs/edit?tab=t.0#heading=h.u46lpf9a77zp) with message types that would handle these scenarios. This PR implements the proposed logic to resync the agent/principal when they restart.

Changes introduced:

- New event types that are exchanged when the agent/principal restarts. These events are exchanged only ONCE when the process restarts
- Use checksums to avoid redundant events. If the checksums match, we can assume that the components are in sync
- Support for both managed and autonomous mode
- Unit tests

Simple sequence diagram for the logic proposed in the [doc](https://docs.google.com/document/d/1NwQaYxmRva8irx_clPTqRu_YQ8MNeFNXSTJoOgv7ZRs/edit?tab=t.0#heading=h.u46lpf9a77zp):
[Managed Mode](https://github.com/user-attachments/assets/5b234896-ced8-4e01-ac9e-bec74199b37c)
[Autonomous Mode](https://github.com/user-attachments/assets/035347d4-d4a8-48b4-8e9f-fb308c68d409)

Fixes https://github.com/argoproj-labs/argocd-agent/issues/94

- Create an app and ensure that it is synced to the agent
- Stop the principal process
- Delete/Update the app on the control plane
- The agent and the principal are now out of sync
- Restart the principal process
- The agent should be in sync with the principal

**Checklist**

### DIFF
--- a/agent/connection_test.go
+++ b/agent/connection_test.go
@@ -1,0 +1,73 @@
+// Copyright 2024 The argocd-agent Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package agent
+
+import (
+	"testing"
+
+	"github.com/argoproj-labs/argocd-agent/internal/event"
+	"github.com/argoproj-labs/argocd-agent/pkg/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestResyncOnStart(t *testing.T) {
+	a := newAgent(t)
+	a.emitter = event.NewEventSource("test")
+	logCtx := log()
+
+	err := a.queues.Create(a.remote.ClientID())
+	assert.Nil(t, err)
+
+	t.Run("should return if the agent has already been synced", func(t *testing.T) {
+		a.resyncedOnStart = true
+		err := a.resyncOnStart(logCtx)
+		assert.Nil(t, err)
+
+		sendQ := a.queues.SendQ(a.remote.ClientID())
+		assert.Zero(t, sendQ.Len())
+	})
+
+	t.Run("send request entity resync in autonomous mode", func(t *testing.T) {
+		a.resyncedOnStart = false
+		a.mode = types.AgentModeAutonomous
+		err := a.resyncOnStart(logCtx)
+		assert.Nil(t, err)
+
+		sendQ := a.queues.SendQ(a.remote.ClientID())
+		assert.Equal(t, 1, sendQ.Len())
+
+		ev, shutdown := sendQ.Get()
+		assert.False(t, shutdown)
+
+		assert.Equal(t, event.EventRequestEntityResync.String(), ev.Type())
+		assert.True(t, a.resyncedOnStart)
+	})
+
+	t.Run("send basic entity list in managed mode", func(t *testing.T) {
+		a.resyncedOnStart = false
+		a.mode = types.AgentModeManaged
+		err := a.resyncOnStart(logCtx)
+		assert.Nil(t, err)
+
+		sendQ := a.queues.SendQ(a.remote.ClientID())
+		assert.Equal(t, 1, sendQ.Len())
+
+		ev, shutdown := sendQ.Get()
+		assert.False(t, shutdown)
+
+		assert.Equal(t, event.RequestBasicEntity.String(), ev.Type())
+		assert.True(t, a.resyncedOnStart)
+	})
+}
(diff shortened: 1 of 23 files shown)


chetan-rns:
Sorry for the big PR. These event types are interconnected making it difficult to break them into multiple PRs.

internal/resources/resources.go
```diff
@@ -0,0 +1,234 @@
...
+
+	agentResources, ok := r.resources[agent]
+	if !ok {
+		return []byte{}
+	}
+
+	return agentResources.Checksum()
+}
+
+// Resources is a map of all the resources that are currently in the cluster.
```
jannfis: I think that comment is a bit misleading, or at least I hope so. It's not that we pull in all the resources from the cluster, but only the ones that are being synced between principal and agents (such as, Applications and AppProjects), right?

agent/connection_test.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2024 The argocd-agent Authors
```
jannfis:
```suggestion
// Copyright 2025 The argocd-agent Authors
```

internal/resources/resources.go
```diff
@@ -0,0 +1,234 @@
+// Copyright 2024 The argocd-agent Authors
```
jannfis:
```suggestion
// Copyright 2025 The argocd-agent Authors
```

internal/resources/resources_test.go
```diff
@@ -0,0 +1,168 @@
+// Copyright 2024 The argocd-agent Authors
```
jannfis:
```suggestion
// Copyright 2025 The argocd-agent Authors
```

internal/resync/resync.go
```diff
@@ -0,0 +1,361 @@
+// Copyright 2024 The argocd-agent Authors
```
jannfis:
```suggestion
// Copyright 2025 The argocd-agent Authors
```

internal/resync/resync.go
```diff
@@ -0,0 +1,361 @@
...
+	// The destination field will be modified to "in-cluster" by the agent.
+	// Principal and the agent checksums will differ if we don't remove the destination field.
+	delete(specMap, "destination")
+
+	specBytes, err := json.Marshal(resSpec)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal application spec: %v", err)
+	}
+
+	checksum := sha256.Sum256(specBytes)
```
jannfis, line 358: I was wondering if we'd be able to use hash algorithms other than SHA-256 for this purpose? Just asking because I think that we do not need the cryptographic promises of SHA-256 here, and SHA-256 is really slow.

internal/resync/resync_test.go
```diff
@@ -0,0 +1,348 @@
+// Copyright 2024 The argocd-agent Authors
```
jannfis:
```suggestion
// Copyright 2025 The argocd-agent Authors
```

jannfis (commented):
Oops. Forgot to hit the submit button.

Change looks awesome at first glance, have a minor nits and a few questions in the first round of reviews.

internal/resources/resources.go
chetan-rns: Yeah. Only the resources (apps,appProjects) that the agent/principal informers handle. Updated the comment

internal/resync/resync.go
chetan-rns: I've replaced SHA-256 with md5 which is faster without the cryptographic promises. There's also blake but I think md5 is much more simpler for this purpose.

jannfis: Not sure about the claim of speed with MD5, to be honest. 

Seems with very small data chunks (16 and 64 bytes), MD5 and SHA256 are about equal on my CPU (11th gen mobile i7) , but then SHA256 definitely becomes faster than MD5 up to almost twice. 

```
$ openssl speed md5 sha1 sha256
...
The 'numbers' are in 1000s of bytes per second processed.
type             16 bytes     64 bytes    256 bytes   1024 bytes   8192 bytes  16384 bytes
md5             114839.44k   307933.58k   624454.57k   832024.66k   880315.05k   913021.69k
sha1            131719.37k   415097.96k  1031060.72k  1638272.94k  2033814.19k  2036344.89k
sha256          118151.61k   283071.98k   883516.53k  1404555.73k  1716873.33k  1732094.63k
```

And I think there are other, non-cryptographic checksum algorithms there that perform much faster (both, md5 and sha-1 claimed cryptographic promises before they were broken). 

So the ultimate question to me is, what are we trying to achieve with this checksum? From my understanding, we want to detect whether something changed. But not for the purpose of verifying its integrity, but for the purpose of deciding whether we need to take action (e.g. sync that particular resource). Is that the right understanding? 

If the answer to that is yes, the next question would be: What happens if the same checksum is calculated for a different input by accident, and is there an attack vector where somebody would gain an advantage by tricking us to calculate the same checksum for a different input under their control.

So maybe we should rather stay at SHA256 for now. And start thinking about using a non-cryptographic, much faster non-cryptographic hashing algorithm (such as FNV, heck maybe even crc32) in the future.

chetan-rns: I've reverted it to SHA256 for now. The checksum is used to detect if a resource has changed and if there is a collision some of the resources could remain stale until the next event.

jannfis (approved):
LGTM

PR #312, opened 2025-02-20
fix: Only remap cluster when label changed

jannfis:

The informer callback would always remap a cluster on resource update, regardless of whether that's necessary.

This PR fixes that.

Fixes #?

**Checklist**


jannfis:
Turns out not be a bug :)

Issue #316, opened 2025-02-26
When the resource proxy is disabled, the principal is still looking for the resource proxy tls causing the pod to crash


mikeshng:
**Describe the bug**
I have the resource proxy is disabled via setting `ARGOCD_PRINCIPAL_ENABLE_RESOURCE_PROXY` to false in the principal deployment. The principal pod is crashing because it's still looks for the resource proxy tls secret.

```
[FATAL]: Error reading TLS config for resource proxy: error getting proxy certificate: could not read TLS secret open-cluster-management-hub/resource-proxy-tls: secrets "resource-proxy-tls" not found
```

**Steps to reproduce the behaviour**
1. Set the principal deployment `ARGOCD_PRINCIPAL_ENABLE_RESOURCE_PROXY` to false
2. Apply the principal deployment.
3. See the pod crash with the above error.

**Expected behavior**
When the resource proxy is disabled, principal should not look for the resource proxy tls because it's not needed.

**Additional context**

PR #317, opened 2025-02-26
fix: When ResourceProxy is disabled, the principal should not look for the ResourceProxyTLS

mikeshng:
The principal pod is crashing when ResourceProxy is disabled because it needs the ResourceProxyTLS.
This fix is to make sure the principal skip looking for the ResourceProxyTLS when ResourceProxy is disabled.

Fixes https://github.com/argoproj-labs/argocd-agent/issues/316

Built and tested locally.
Set principal deployment env `ARGOCD_PRINCIPAL_ENABLE_RESOURCE_PROXY` to false.
Apply the deployment and it did not crash with the changes.

**Checklist**


jannfis (approved):
LGTM, thanks!

PR #337, opened 2025-03-17
feat: Add mTLS as an optional authentication method.

### LINKED ISSUE #335: Add an optional mTLS authentication method with customizable agent identity extraction
Currently, the argocd-agent project uses mTLS for secure gRPC communication between the principal and agent, but it also requires an additional authentication layer (by default, username/password) to identify and authorize each agent.
In env already leveraging auto rotation short lived client certificates such as [Open Cluster Management (OCM)](https://open-cluster-management.io/), this additional authentication might not be necessary.

Introduce an optional mTLS authentication method, which leverages client certificate identities directly for authentication and authorization. This new method should support customizable extraction of agent identities such as cluster/agent names from certificate subjects using configurable regular expressions (or other pattern definition).

In the principal config, introduce:
```
# optional - by default the value is "userpass"
principal.auth.method: "mtls"

# optional - used with mtls auth method
principal.auth.mtls.agent-id-regex: "_regex_here_"
```

In the agent config, introduce:
```
# optional - by default the value is "userpass"
agent.creds: "mtls"
```

Existing TLS configs (agent.tls.* and principal.tls.*) continue to be used without changes. Those are what the principal and agent will use for the new mTLS authentication.

Introduce an OCM specific authentication method with hardcoded subject extraction specifically to the OCM certificate subject format. This method would still rely on mTLS under the hood. However, this approach is less flexible compared to the generic solution proposed, as the generic method would also support non OCM use cases if there are other non OCM platforms that also leverages mTLS client cert authentication. 

**Additional context**
An example of an OCM subject inside a short live auto rotate certificate used by the argocd-agent components for existing mTLS:
```
Subject: O = system:authenticated + O = system:open-cluster-management:addon:argocd-agent + O = system:open-cluster-management:cluster:cluster1:addon:argocd-agent, CN = system:open-cluster-management:cluster:cluster1:addon:argocd-agent:agent:argocd-agent-agent
```

We should consider extend TLS code for ability to check CRL but this could be done later as recommended by Jann.

mikeshng:
- Add mTLS as an optional authentication method.
- The agent identity can be extracted from the client cert subject  via user defined regex pattern.

Allow the argocd-agent project to use mTLS for secure gRPC communication between the principal and agent without requiring an additional authentication layer (by default username/password) to identify and authorize each agent.

Fixes #335 

- The regex used for testing `open-cluster-management:cluster:([^:]+):addon:argocd-agent`
- Tested using the new mtls auth method.
- Tested using the existing userpass auth method.

**Checklist**


jannfis:
Nice, I think this is the way to go :+1:

mikeshng:
> Nice, I think this is the way to go 👍

Thank you Jann for all the initial feedback in Slack. I've completed some local testings so I am setting this PR for ready for review. CC @jnpacker @ch-stark

cmd/principal/main.go
```diff
@@ -167,10 +170,27 @@ func NewPrincipalRunCommand() *cobra.Command {
...
+				if authConfig != "" {
+					regex, err = regexp.Compile(authConfig)
+					if err != nil {
+						cmdutil.Fatal("Error compiling mtls agent id regex: %v", err)
+					}
+				}
+				mtlsauth := mtls.NewMTLSAuthentication(regex)
+				err := authMethods.RegisterMethod("mtls", mtlsauth)
+				if err != nil {
+					cmdutil.Fatal("Could not register mtls auth method")
```
jannfis:
```suggestion
					cmdutil.Fatal("Could not register mtls auth method: %v", err)
```

hack/vscode/launch.json
```diff
@@ -14,7 +14,7 @@
                 "--insecure-jwt-generate",
                 "--kubecontext=vcluster-control-plane",
                 "--log-level=trace",
-                "--passwd=${workspaceFolder}/hack/demo-env/creds/users.control-plane"
+                "--auth=userpass:${workspaceFolder}/hack/demo-env/creds/users.control-plane"
```
jannfis, line 17: :+1:

internal/auth/mtls/mtls.go
```diff
@@ -0,0 +1,71 @@
...
+		matches := m.AgentIDRegex.FindStringSubmatch(subject)
+		if len(matches) < 2 {
+			return "", fmt.Errorf("the TLS subject '%s' does not match the agent ID regex pattern", subject)
+		}
+		agentID = matches[1]
+	}
+	if agentID == "" {
+		return "", fmt.Errorf("agent ID is empty")
+	}
+	return agentID, nil
```
jannfis: Can we please ensure that `agentID` validates against a DNS label here?

jannfis (commented):
Just have a couple of minor things, PTAL. Otherwise, great stuff!

cmd/principal/main.go
mikeshng: Done as suggested. Thanks.

internal/auth/mtls/mtls.go
mikeshng: Done as suggested. Thanks.

jannfis (approved):
LGTM, thanks!

PR #357, opened 2025-04-04
chore: fix flaky unit test

jparsai:
This PR is to fix a flaky unit test, where `Test_UpdateClusterInfo` test fails sometimes because getClusterInfo() takes more than 1 second to fetch data from Redis cache. It is fixed by increasing time range. It can be tested by running below command to confirm.

`go test -run Test_UpdateClusterInfo -count 100 -failfast`


jannfis (approved):
LGTM, thanks

PR #358, opened 2025-04-06
feat: Add version flag to principal command

### LINKED ISSUE #354: Principal component should print version on request

The agent component prints version information if one uses `--version` (and optionally, `--version-format`). The principal does not, but it should.

Similar to the agent, the principal component should be able to output its own version.

**Additional context**

ATGardner:
Introduce a `--version` flag to the principal command, allowing users to display version information for the argocd-agent. An optional `--version-format` flag specifies the output format (text, json, or json-indent).

Fixed #354

i implemented the `--version` flag in the same way as it is currently implemented in the agent. is there a reason this is not a separate sub command in both cases? i can make that change as well, if it is preferred.

**Checklist**

* [] Documentation update is required by this PR (and has been updated) OR no documentation update is required.

### DIFF
--- a/cmd/principal/main.go
+++ b/cmd/principal/main.go
@@ -35,6 +35,7 @@ import (
 	"github.com/argoproj-labs/argocd-agent/internal/kube"
 	"github.com/argoproj-labs/argocd-agent/internal/labels"
 	"github.com/argoproj-labs/argocd-agent/internal/tlsutil"
+	"github.com/argoproj-labs/argocd-agent/internal/version"
 	"github.com/argoproj-labs/argocd-agent/principal"
 	cacheutil "github.com/argoproj/argo-cd/v2/util/cache"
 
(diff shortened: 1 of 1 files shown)


chetan-rns:
```
++ kubectl --context vcluster-control-plane -n argocd get secret cluster-agent-autonomous -o 'jsonpath={.data.server}'
Error from server (NotFound): secrets "cluster-agent-autonomous" not found
```

I see a similar error on other PRs as well. Perhaps a race condition where we try to access the secret before it is created.

jannfis:
> I see a similar error on other PRs as well. Perhaps a race condition where we try to access the secret before it is created.

Yes. The value of this environment variable is not used anywhere in the code, so we may as well remove setting the environment variable from the script. @jparsai Did you intend anything to do with that in the future?

jannfis:
> is there a reason this is not a separate sub command in both cases?

@ATGardner I think the main reason is, that we have not implemented any sub commands as of now. For the CLI (`argocd-agentctl`), it makes more sense to have it a separate sub command though.

jannfis (approved):
LGTM, thank you so much @ATGardner!

PR #359, opened 2025-04-07
fix: agent should use the default queue name instead of client ID

chetan-rns:

Since each agent connects to a single principal, the queue name can be a constant. We can keep the queue open even if the agent is disconnected from the principal.

Fixes #?

**Checklist**


jannfis (approved):
LGTM

PR #361, opened 2025-04-07
chore(e2e): Fix race condition for agent configuration

jannfis:

Agent configuration is created in the `start-e2e` target. However, we run the `start-e2e` target in the background, and GH actions will progress quickly to the next step in the workflow.

Sometimes, when the runner is a little slow, the agent configuration does not yet exist when required.

This PR moves the agent configuration creation to `setup-e2e` target.

Fixes #?

**Checklist**


chetan-rns (approved):
LGTM!

PR #369, opened 2025-04-16
fix: don't remove the queue/eventwriter on disconnect

chetan-rns:

Currently, we remove the queue/eventwriter in the Subscribe() when an agent disconnects with the principal. We recreate the queue if there are new events from the informer. However, the queue/eventwriter could have some pending events yet to be transmitted. Removing them could lead to losing these pending events. We want the queue to keep the events, even if there is a transient disconnect between the agent and the principal.

1. Don't remove the queue and the eventwriter on disconnect
2. sendFunc() should exit when the client context is cancelled
3. Remove the queue/eventwriter when the namespace is deleted.

Fixes #?

**Checklist**

### DIFF
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -540,6 +540,12 @@ func NewEventWriter(target streamWriter) *EventWriter {
 	}
 }
 
+func (ew *EventWriter) UpdateTarget(target streamWriter) {
+	ew.mu.Lock()
+	defer ew.mu.Unlock()
+	ew.target = target
+}
+
 func (ew *EventWriter) Add(ev *cloudevents.Event) {
 	resID := ResourceID(ev)
 	logCtx := ew.log.WithFields(logrus.Fields{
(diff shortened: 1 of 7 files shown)


internal/event/event.go
```diff
@@ -692,3 +698,45 @@ func (ew *EventWriter) sendEvent(resID string) {
 		isACKRemoved = true
 	}
 }
+
+// eventWritersMap provides a thread-safe way to manage event writers.
+type EventWritersMap struct {
```
chetan-rns, line 703: This is unchanged. Moved from the eventstream package.

principal/apis/eventstream/eventstream.go
```diff
@@ -295,10 +247,14 @@ func (s *Server) sendFunc(c *client, subs eventstreamapi.EventStream_SubscribeSe
 	// Get() is blocking until there is at least one item in the
 	// queue.
 	logCtx.Tracef("Waiting to grab an item from the queue")
-	ev, shutdown := q.Get()
+	ev, shutdown := queue.GetWithContext(q, c.ctx)
 	if shutdown {
```
chetan-rns, line 251: Get() is a blocking call, and it suspends until a new item is added to the queue or if the queue shuts down. Since we don't remove the queue on disconnect, Get will never return. So, I've added a wrapper around Get() that checks if the context is cancelled. So, on disconnect, the client context is cancelled, and the Get should no longer block.

internal/queue/queue.go
```diff
@@ -168,3 +176,45 @@ func (q *SendRecvQueues) Delete(name string, shutdown bool) error {
...
+			bq.notify.L.Unlock()
+		case <-done:
+			// If function exits early, prevent leak
+		}
+	}()
+
+	// Suspend until an item is available or context is cancelled
+	for q.Len() == 0 && ctx.Err() == nil {
+		bq.notify.Wait()
+	}
```
chetan-rns, line 204: I think using a `for with select` to check if a context is done is more simple. I'm using a condition variable so that we don't spin continuously in a for loop and poll if the queue has new items.

```
for {
 select {
    case <-ctx.Done:
         // return from function
    default:
       if q.Len() == 0:
            continue
       time.Sleep(100 * millisecond)
  }
}
```

I don't think polling would be a big issue now. So, I'm fine with replacing the condition variable with the simple for select logic. Or even better, if there is another easy way to handle this case.

jannfis: Have you considered using a simple channel instead of a condition to synchronize, as pointed out in the [docs](https://pkg.go.dev/sync#Cond)?

chetan-rns: I was trying to extend the [condition variable logic](https://github.com/kubernetes/client-go/blob/master/util/workqueue/queue.go#L277) used by workqueue's Get() to be aware of ctx. But I agree that channels are more idiomatic. We could use a buffered channel (non-blocking) to signal when an item is added to the queue.

PR #372, opened 2025-04-23
test: e2e tests to verify resync after connection loss using toxiproxy

chetan-rns:

* Use toxiproxy to simulate connection issues between the agent and the principal.
* Verify if the components resync after the connection is re-enabled.

Fixes #?

**Checklist**


test/e2e2/fixture/toxyproxy.go
```diff
@@ -0,0 +1,107 @@
...
+
+		if err = proxyServer.Close(); err != nil {
+			t.Errorf("failed to close proxy server: %v", err)
+		}
+
+		// Restart the agent process
+		RestartAgent(t, agentName)
+
+		// Give some time for the agent to be ready
+		time.Sleep(10 * time.Second)
```
chetan-rns: I'm using sleep to wait until the agent is up and running. Since it's not a server, we can't use `net.Dial()`. Perhaps we should add custom health check logic to determine if the agent is running.

jannfis: Good point. We would need such a health check anyway for readiness/liveness probes.

chetan-rns: Included a simple health check server and removed the call to sleep.

jannfis: Thanks! Would be a big effort to submit that as a separate PR -- it's a new feature, after all and I think it should be treated as such.

We'll merge the health PR first, then we can merge the test PR afterwards.

chetan-rns: Makes sense. Raised https://github.com/argoproj-labs/argocd-agent/pull/374 for health check

jannfis: Thank you! Merged it.

PR #375, opened 2025-04-28
docs: Add docs for metrics and profiling

jparsai:
This PR is to add documentation for ArgoCD agent metrics, also there is a minor change for pprof implementation.

### DIFF
--- a/docs/operations/metrics.md
+++ b/docs/operations/metrics.md
@@ -1,0 +1,43 @@
+# Metrics
+The argocd-agent exposes different sets of Prometheus metrics for agent and principal components.
+
+Metrics are by default enabled in both principal and agent but they will be disabled if metrics port is set to `0`.
+
+Metrics for principal are exposed at `http://0.0.0.0:8000/metrics` endpoint. Users can overwrite metrics port by setting `--metrics-port` flag in CLI or `ARGOCD_PRINCIPAL_METRICS_PORT` environment variable.
+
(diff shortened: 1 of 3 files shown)


docs/concepts/metrics.md
```diff
@@ -0,0 +1,50 @@
...
+|   Label Name  |   Example Value   |   Description |
+|--------------------|---------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+|   call_status | success   |   Status of event processing. Possible values are: success, failure, discarded, not-allowed.  |
+|   agent_name  |   agent-managed   |   Name of Agent. Possible values are: agent-managed, agent-autonomous.    |
+|   resource_type   |   application |   Type of resource. Possible values are: application, app project, resource, resourceResync.   |
+
+
+# Profiling
+The argocd-agent uses `pprof` for collecting profiling data.
+Profiling is by default disabled in both control-plane and agent and it will be enabled if `pprof-port` flag is set or environment variable `ARGOCD_PRINCIPAL_PPROF_PORT` in control-plane and `ARGOCD_AGENT_PPROF_PORT` in agent are set.
```
jannfis: Should be _principal and agent_ instead of _control-plane and agent_

jannfis (commented):
Thanks. This should really be two PRs: One adding the pprof for the agent and pprof related changes to the principal, the second one for adding the metrics docs.

As for the docs, I think they are at the top-level category. We might want to introduce a new top level category, for example named "Operations", where this would better fit. You will find this kind of configuration in mkdocs.yaml, and every page in the docs need to be referred in this configuration. Please make sure the docs are reachable and looking good in a rendering preview. You can either use `make serve-docs` locally for that, or use the link from the PR's readthedocs check.

jannfis (approved):
LGTM

PR #376, opened 2025-04-28
feat: update pprof implementation

jparsai:
This PR is to modify pprof implementation.


jannfis:
Thought exercise: As for the semantic PR title - do you think it is a chore, or how would you classify the addition of something to one of the components?

PR #380, opened 2025-05-01
refactor: Reduce cognitive complexity of informer package

jannfis:

Preparations for complexity linters.

Fixes #?

**Checklist**


internal/informer/informer.go
```diff
@@ -205,10 +223,9 @@ func NewInformer[T runtime.Object](ctx context.Context, opts ...InformerOption[T
 		},
 	})
 	if err != nil {
-		return nil, err
+		return err
 	}
-	i.ctlCh = make(chan struct{})
-	return i, nil
+	return nil
```
chetan-rns, line 228: Nit: We can return the error directly instead of `if err != nil { return err }`

chetan-rns (approved):
LGTM!

internal/informer/informer.go
jannfis: Next time :)

PR #390, opened 2025-05-15
feat: Only allow managed resources through resource proxy

### LINKED ISSUE #291: Agents should check if requested resource from resource proxy belongs to a managed Application

Right now, the resource proxy can request any resource on the agent systems that the agent has RBAC permissions for. Authorization is done solely by the Argo CD API running on the principal, and the subsequent requests trust this request. As of now, there is no sensible way to carry this authorization to the agent.

When the agent receives a request for resource retrieval, it should check whether that resource belongs to an Application managed by Argo CD. Since some resources are not directly managed (for example, pods are indirectly managed through a ReplicaSet, which in turn is usually indirectly managed by something like a Deployment or StatefulSet), the agent should be able to resolve such a relationship (i.e. by following ownerReferences).

If the requested resource is not managed by Argo CD, a 403 should be returned.

In a next step, a sanity check against AppProject permissions can be performed.

n/a

**Additional context**
n/a

jannfis:

This PR makes sure that the agent only ever returns resources to resource proxy requests that are deemed managed by Argo CD.

Fixes #291 

**Checklist**

### DIFF
--- a/agent/resource.go
+++ b/agent/resource.go
@@ -81,7 +81,7 @@ func (a *Agent) processIncomingResourceRequest(ev *event.Event) error {
 		logCtx.Error("Remote queue disappeared")
 		return nil
 	}
-	q.Add(a.emitter.NewResourceResponseEvent(rreq.UUID, event.HttpStatusFromError(status), string(jsonres)))
+	q.Add(a.emitter.NewResourceResponseEvent(rreq.UUID, event.HTTPStatusFromError(status), string(jsonres)))
 	logCtx.Tracef("Emitted resource response")
 
 	return nil
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -230,10 +230,10 @@ type ResourceResponse struct {
 	Resource string `json:"resource,omitempty"`
 }
 
-// HttpStatusFromError tries to derive a HTTP status code from the error err.
+// HTTPStatusFromError tries to derive a HTTP status code from the error err.
 // If err is non-nil and the code cannot be derived, it will be set to HTTP 500
 // InternalServerError. If err is nil, HTTP 200 OK will be returned.
-func HttpStatusFromError(err error) int {
+func HTTPStatusFromError(err error) int {
 	if err != nil {
 		if status, ok := err.(apierrors.APIStatus); ok || errors.As(err, &status) {
 			return int(status.Status().Code)


internal/kube/client.go
```diff
@@ -87,6 +91,11 @@ func NewKubernetesClientFromConfig(ctx context.Context, namespace string, kubeco
 
 	cl := NewKubernetesClient(ctx, clientset, applicationsClientset, namespace)
 	cl.RestConfig = config
+	dyn, err := dynamic.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	cl.DynamicClient = dyn
```
chetan-rns, line 96: I think we missed initializing the DiscoveryClient here. Or we can directly use the `Discovery()` method of Clientset.

internal/event/event.go
```diff
@@ -184,6 +188,36 @@ type ResourceRequest struct {
...
+	return r.Namespace == ""
+}
+
+// IsNamespaced returns true if the resource request is a namespaced request
+func (r *ResourceRequest) IsNamespaced() bool {
+	return r.Namespace != ""
+}
+
+// IsValid returns true if the resource request is valid
+func (r *ResourceRequest) IsValid() bool {
```
chetan-rns, line 217: Some of these methods are not used. Do we plan to use them in the future?

agent/resource.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+
+		// This should not happen, as it means we couldn't resolve the owner's
+		// Kind to a resource available in the cluster.
+		if gvr.Resource == "" {
+			continue
+		}
+
+		// Get the owning object from the cluster.
+		owner, err := kube.DynamicClient.Resource(gvr).Namespace(res.GetNamespace()).Get(context.Background(), ref.Name, v1.GetOptions{})
```
chetan-rns, line 207: Do we need to handle the case of a namespace-scoped resource having a cluster-scoped owner? Since we use the namespace of `res` I guess the client would be restricted to access resources only in that namespace.

chetan-rns (commented):
Added some questions. Everything else looks good to me!

agent/resource.go
jannfis: Ah, good catch. I missed the fact that namespace-scoped dependants may specify cluster-scoped parents. Will fix it.

internal/kube/client.go
jannfis: Good catch, I think we should always derive the discovery client from the clientset, using the `Discovery()` method. Updated the code.

internal/event/event.go
jannfis: Yep, I thought they'd be good to have. I now have added unit tests for them, at least :)

jannfis:
@chetan-rns Thanks for your review. Addresses all your questions. Please take another look.

chetan-rns (approved):
Thanks. LGTM!

PR #391, opened 2025-05-16
build: Add service manifest to expose metrics server

jparsai:
This PR is to add kube manifests to expose metrics servers running on principal and agent via a Service.


jannfis:
Are the manifests being left out intentionally from the Kustomize build?

jparsai:
I updated it, PTAL.

jannfis (approved):

PR #399, opened 2025-05-29
chore: Add support for developing on non-local K8s via reverse tunnel

jgwest:

At present, some of the E2E tests (and product features) assume that K8s is running in a specific microk8s-based configuration with a set IP range.

This script adds support for developing on remote K8s clusters (as long as they support LoadBalancers), via a rathole-based reverse tunnel for argocd-agent's resource-proxy.
- This PR should be considered an optional configuration as an alternative to existing microk8s-based configuration.

There are [many many many ways to reverse tunnel](https://github.com/anderspitman/awesome-tunneling) traffic these days. In this PR I went with [Rathole](https://github.com/yujqiao/rathole?tab=readme-ov-file), because I have used it previously, and thus could enable it quickly. Unfortunately, rathole is no longer being developed (AFAICT), so in the medium/long term a tool like 'telepresence' or 'frp' would be better.

See [README.md](https://github.com/jgwest/argocd-agent/blob/proxy-load-balance-resource-proxy-may-2025/hack/dev-env/reverse-tunnel/README.md) for details.

### DIFF
--- a/hack/dev-env/reverse-tunnel/README.md
+++ b/hack/dev-env/reverse-tunnel/README.md
@@ -23,7 +23,7 @@ To enable the configuration:
 make setup-e2e2
 
 # Installs Rathole Deployment/Services on vcluster, update 'managed-agent' cluster Secret to point to Rathole Deployment, and start local rathole container
-hack/dev-env/reverse-tunnel/setup.sh
+./hack/dev-env/reverse-tunnel/setup.sh
 
 ```
 


hack/dev-env/reverse-tunnel/README.md
```diff
@@ -0,0 +1,62 @@
...
+### How to enable
+
+To enable the configuration:
+
+```
+# Setup environment as usual
+make setup-e2e2
+
+# Installs Rathole Deployment/Services on vcluster, update 'managed-agent' cluster Secret to point to Rathole Deployment, and start local rathole container
+hack/dev-env/reverse-tunnel/setup.sh
```
jannfis: Just very much nitpicking.

```suggestion
./hack/dev-env/reverse-tunnel/setup.sh
```

jannfis (approved):

Issue #238, opened 2024-11-25
If a user modifies .spec field of Application on Managed Agent (or any non-source of truth peer), it is not reconciled back


jgwest:
If a user modifies the `.spec` field of an Application on Managed Agent (or any non-source of truth peer), it is not reconciled back to the canonical value from principal.

Arguable whether this is an enhancement or a bug, it's right on the line between the two :smile:. Feel free to re-classify.

**Consider the following scenario**:
1) Principal and managed agent are connected. They share one Application `appA`.
2) A user on managed agent cluster modifies the `.spec` field of Application.
- (Same scenario but for principal and autonomous agent)

**Expected Behaviour**:
- Managed-agent detects that the `.spec` field has been modified, and reconciles the Application's `.spec` field back to the canonical value from principal.

**Actual Behaviour**:
- Managed-agent does not reconcile `.spec` field back (in fact, I think it treats it as a .status update instead?) My expectation is that it will only be reconciled back when the principal is next reconciled.

At least two possible solutions:
- A) When managed agent detects a `.spec` change, it should send `request-update` message with checksum. Principal will reply with canonical version of the `.spec` if the checksum differs. Agent will update back to that.
- B) Agent maintains an in-memory copy of all applications, and reconciles them back to that value on detected spec change.

~My personal preference is A, since it is simpler (given that it fits with other logic we already implementing). However A) would not cover all cases: for example, what if managed agent is temporarily disconnected from principal (e.g. the managed agent on a cruise ship scenario).~

I think B) is a better solution, as it will better handle all scenarios, for example, what if managed agent is temporarily disconnected from principal (e.g. the managed agent on a cruise ship scenario). This solution will be more complex, as [there are two hard problems in computer science, cache invalidation and naming things](https://martinfowler.com/bliki/TwoHardThings.html).

**Issue Criteria**:
* Peer (e.g. managed agent) detects and revert changes to the `.spec` field of resource (e.g. Application) back to canonical value.
* E2E test which verifies the issue is resolved.

Horiodino:
yeah i think option two make more sense the reconciliation happens locally on the Managed Agent without needing to communicate with the Principal for every .spec field change. 

hey I’d like to work on this. Is it okay if no one else is currently working on it?

PR #400, opened 2025-06-02
fix: Direct changes in app on managed-cluster are not reverted by agent

jparsai:
This PR is to fix a bug when user makes any changes in application on managed-cluster instead of principal, but managed-agent does not revert those changes. As principal is source of truth in managed mode changes in managed cluster should not be allowed.

Fixes https://github.com/argoproj-labs/argocd-agent/issues/238

### DIFF
--- a/agent/inbound.go
+++ b/agent/inbound.go
@@ -340,16 +340,17 @@ func (a *Agent) createApplication(incoming *v1alpha1.Application) (*v1alpha1.App
 	incoming.Spec.Destination.Server = ""
 	incoming.Spec.Destination.Name = "in-cluster"
 
+	if a.mode == types.AgentModeManaged {
+		// Store app spec in cache
+		appCache.SetApplicationSpec(incoming.UID, incoming.Spec, logCtx)
+	}
+
 	created, err := a.appManager.Create(a.context, incoming)
 	if apierrors.IsAlreadyExists(err) {
 		logCtx.Debug("application already exists")
 		return created, nil
 	}
 
-	if a.mode == types.AgentModeManaged && err == nil {
-		// Store app spec in cache
-		appCache.SetApplicationSpec(incoming.UID, incoming.Spec)
-	}
 	return created, err
 }
 
@@ -379,12 +380,12 @@ func (a *Agent) updateApplication(incoming *v1alpha1.Application) (*v1alpha1.App
 	var napp *v1alpha1.Application
 	switch a.mode {
 	case types.AgentModeManaged:
+
+		// Update app spec in cache
 		logCtx.Tracef("Calling update spec for this event")
+		appCache.SetApplicationSpec(incoming.UID, incoming.Spec, logCtx)
+
 		napp, err = a.appManager.UpdateManagedApp(a.context, incoming)
-		if err == nil {
-			// Update app spec in cache
-			appCache.SetApplicationSpec(incoming.UID, napp.Spec)
-		}
 	case types.AgentModeAutonomous:
 		logCtx.Tracef("Calling update operation for this event")
 		napp, err = a.appManager.UpdateOperation(a.context, incoming)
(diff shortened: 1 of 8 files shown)


agent/inbound.go
```diff
@@ -346,6 +346,10 @@ func (a *Agent) createApplication(incoming *v1alpha1.Application) (*v1alpha1.App
 		return created, nil
 	}
 
+	if a.mode == types.AgentModeManaged && err == nil {
```
chetan-rns: Nit: We have check if the agent is in managed mode a couple of lines above.

agent/outbound.go
```diff
@@ -74,6 +79,26 @@ func (a *Agent) addAppUpdateToQueue(old *v1alpha1.Application, new *v1alpha1.App
 		return
 	}
 
+	// Revert any direct modifications done in application on managed-cluster
+	// because for managed-agent all changes should be done through principal
+	sourceUID, exists := new.Annotations[manager.SourceUIDAnnotation]
```
chetan-rns: Can we move this block to a different function? We can also reduce nested conditions.

internal/cache/appcache.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"sync"
+
+	"github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// ApplicationSpecCache type is for caching Argo CD application spec to keep
+// application in sync with last known state of principal application
+type ApplicationSpecCache struct {
```
chetan-rns, line 27: Similarly, we might have to revert the changes made to AppProjects. Do we plan to handle that case in a different PR?

internal/cache/appcache.go
```diff
@@ -0,0 +1,63 @@
...
+// SetApplicationSpecCache inserts/updates app spec in cache
+func SetApplicationSpecCache(sourceUID types.UID, app v1alpha1.ApplicationSpec) {
+	appSpecCache.lock.Lock()
+	defer appSpecCache.lock.Unlock()
+
+	appSpecCache.appSpec[sourceUID] = app
+}
+
+// GetApplicationSpecCache returns cached app spec
+func GetApplicationSpecCache(sourceUID types.UID) (v1alpha1.ApplicationSpec, bool) {
```
chetan-rns: Nit: We could remove the cache suffix from the function names since it is part of the cache package. 
`cache.SetApplicationSpec` instead of `cache.SetApplicationSpecCache`

internal/cache/appcache.go
chetan-rns: We agreed to handle this issue in a different PR.

chetan-rns (dismissed):
LGTM!

internal/manager/application/application.go
```diff
@@ -590,6 +592,38 @@ func (m *ApplicationManager) RemoveFinalizers(ctx context.Context, incoming *v1a
...
+	})
+
+	sourceUID, exists := app.Annotations[manager.SourceUIDAnnotation]
+	if exists && m.mode == manager.ManagerModeManaged {
+		if cachedAppSpec, ok := appCache.GetApplicationSpec(ty.UID(sourceUID), logCtx); ok {
+			logCtx.Debugf("Application: %s is available in agent cache", app.Name)
+
+			if diff := reflect.DeepEqual(cachedAppSpec, app.Spec); !diff {
+				app.Spec = cachedAppSpec
+				if _, err := m.UpdateManagedApp(ctx, app); err != nil {
```
jgwest, line 618: Can you add a log statement (at Info level) just before the call to `m.UpdateManagedApp`, indicating that we are about to revert the Application?

jgwest (changes requested):
A tiny change then good to go!

internal/manager/application/application.go
jparsai: it is done, PTAL

PR #401, opened 2025-06-02
feat: add support for resource actions in the agent

chetan-rns:

Currently, resource actions are not supported because the resource proxy can only process read requests (GET/LIST). Executing resource actions involves creating a new resource or patching an existing resource on the agent. This PR includes:

- Add support for processing POST and PATCH requests from the argocd server
- Add unit tests and e2e tests

Fixes #?

**Checklist**

### DIFF
--- a/test/e2e2/fixture/argoclient.go
+++ b/test/e2e2/fixture/argoclient.go
@@ -169,7 +169,7 @@ func (c *ArgoRestClient) GetResource(app *v1alpha1.Application, group, version, 
 }
 
 func (c *ArgoRestClient) RunResourceAction(app *v1alpha1.Application, action, group, version, kind, namespace, name string) error {
-	reqUrl := c.url(
+	reqURL := c.url(
 		"appNamespace", app.Namespace,
 		"project", app.Spec.Project,
 		"namespace", namespace,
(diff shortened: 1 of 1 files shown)


agent/resource.go
```diff
@@ -87,6 +101,55 @@ func (a *Agent) processIncomingResourceRequest(ev *event.Event) error {
 	return nil
 }
 
+func (a *Agent) processIncomingPostResourceRequest(ctx context.Context, req *event.ResourceRequest, gvr schema.GroupVersionResource) (*unstructured.Unstructured, error) {
```
jannfis, line 104: Is Argo CD actually using POST to create new resources?

chetan-rns: Not directly. But Argo CD calls [Create()](https://github.com/argoproj/argo-cd/blob/2c3a452fd7514777deb98679f68586242b9feaa7/server/application/application.go#L2509) while executing the resource action that gets translated to a [POST request](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#create-deployment-v1-apps).

jannfis: Ah, it can create jobs - then support for POST (Create) makes sense.

jannfis (approved):
LGTM!

PR #419, opened 2025-06-18
feat: Allow fetching API groups via resource proxy

### LINKED ISSUE #412: Cannot create apps through Argo CD API anymore
**Describe the bug**

I believe https://github.com/argoproj-labs/argocd-agent/pull/390 has broken the ability to create apps through the Argo CD API. The API performs a couple of validating Kubernetes API calls that will not work anymore with the recent changes.

It does not affect the ability to create an app in a declarative manner.

1. Try to create an app through the Argo CD UI with a destination cluster being an agent.

**Expected behavior**

The app to be created.

**Additional context**

jannfis:

This allows fetching the list of available APIs through the resource-proxy.

Fixes #412 

**Checklist**

### DIFF
--- a/agent/resource.go
+++ b/agent/resource.go
@@ -59,7 +59,8 @@ func (a *Agent) processIncomingResourceRequest(ev *event.Event) error {
 	switch rreq.Method {
 	case http.MethodGet:
 		// If we have a request for a named resource, we fetch that particular
-		// resource. If the name is empty, we fetch a list of resources instead.
+		// resource. If the name is empty, we fetch either a list of resources
+		// or a list of APIs instead.
 		if name != "" {
 			if gvr.Resource != "" {
 				logCtx.Debugf("Fetching managed resource %s/%s/%s", gvr.Group, gvr.Version, gvr.Resource)
--- a/agent/resource.go
+++ b/agent/resource.go
@@ -28,8 +28,10 @@ var ErrUnmanaged = errors.New("resource not managed by app")
 //
 // There can be multiple forms of requests. Currently supported are:
 //
-// - Request for a particular resource, both namespace and cluster scoped
-// - Request for a list of available APIs
+//   - Request for a particular resource, both namespace and cluster scoped
+//   - Request for a limited list of resources, both namespace and cluster
+//     scoped.
+//   - Request for a list of available APIs
 func (a *Agent) processIncomingResourceRequest(ev *event.Event) error {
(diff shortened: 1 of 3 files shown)


agent/resource.go
```diff
@@ -57,14 +54,24 @@ func (a *Agent) processIncomingResourceRequest(ev *event.Event) error {
 	// Extract required information from the resource request
 	gvr, name, namespace := rreqEventToGvk(rreq)
 
+	logCtx.Infof("Processing resource request for resource of type %s named %s/%s", gvr.String(), namespace, name)
+
 	switch rreq.Method {
 	case http.MethodGet:
 		// If we have a request for a named resource, we fetch that particular
 		// resource. If the name is empty, we fetch a list of resources instead.
```
ranakan19: This comment should be updated.

ranakan19 (commented):
Looks good overall, could use to update the comment highlighted, and docString comment for function at line 31-32 agent/resource.go

Looks like the endpoint is not changing but probably needs documentation to be updated? Not familiar with the codebase enough to know if this exists and where to point.

agent/resource.go
```diff
@@ -217,6 +229,44 @@ func (a *Agent) getAvailableResources(ctx context.Context, gvr schema.GroupVersi
...
 }
 
+// getAvailableAPIs retrieves a list of available APIs for a given group and version.
+// If group and version are empty, all available APIs are returned.
+// If group is empty and version is not, all APIs for the given version are returned.
+// If group and version are not empty, all APIs for the given group and version are returned.
+//
+// It does not yet support the new aggregated API.
+//
+//nolint:unused
```
chetan-rns: Do we need the `nolint:unused` directive? Or is it a False positive by golangci-lint?

jannfis: Good question. I put it there because the `ctx` parameter is actually unused in the method, but I decided to keep it so that method signatures stay similar. Actually, I have removed the nolint directive and ran the linter, and it did not complain.

agent/resource.go
jannfis: Good catch, thanks!

jannfis:
> Looks like the endpoint is not changing but probably needs documentation to be updated? Not familiar with the codebase enough to know if this exists and where to point.

There is no documentation for this endpoint itself, and the endpoint itself is basically a protocol-breaking reverse proxy for access to the Kubernetes API.

chetan-rns (approved):
LGTM. Thanks!

PR #420, opened 2025-06-18
chore: Add script to convert Argo CD cluster secrets to kubeconf

jannfis:

This script allows you to create a kubeconfig file from Argo CD's cluster secrets. It's intended for development purposes, e.g. to debug the resource proxy.

It's been vibe coded using Cursor.

Fixes #?

**Checklist**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ test/profile
 vendor/
 .vscode
 build/
+__pycache__
--- a/hack/secret-to-kubeconf/README.md
+++ b/hack/secret-to-kubeconf/README.md
@@ -1,0 +1,212 @@
+# Argo CD Cluster Secret to Kubeconfig Converter
+
+This utility converts Argo CD cluster secrets stored in a Kubernetes cluster into kubectl-compatible kubeconfig contexts.
+
+## Overview
+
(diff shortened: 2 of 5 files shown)


aali309:
Is it a good idea to for example to add a `--no-creds` or `--redact-creds` flag to avoid writing client certs or basic auth credentials?
Or may be emit a warning when credentials are written by default?

aali309:
> > Is it a good idea for example to add a `--no-creds` or `--redact-creds` flag to avoid writing client certs or basic auth credentials? Or may be emit a warning when credentials are written by default?
> 
> Good thinking there, but the credentials are an integral part of the resulting kubeconf. The purpose is to read the cluster secrets and stuff them into a kubeconfig to enable use of kubectl against the resource proxy.
> 
> Why shouldn't the credentials being written? Did I miss something?

Not critical, its just a thought. I was mainly thinking about cases where someone might want to preview the clusters without writing sensitive data, like in CI environments or for audit purposes.

jannfis:
> I was mainly thinking about cases where someone might want to preview the clusters without writing sensitive data, like in CI environments or for audit purposes.

You can always inspect the cluster secrets manually :)

aali309 (approved):
LGTM! 🚀

PR #425, opened 2025-06-25
feat: sync appProjects to the appropriate agents

chetan-rns:

Fixes #?

**Checklist**


principal/event.go
```diff
@@ -500,3 +529,11 @@ func (s *Server) createNamespaceIfNotExist(ctx context.Context, name string) (bo
 	}
 	return false, nil
 }
+
+func agentPrefixedProjectName(project, agent string) string {
+	project = agent + "-" + project
+	if len(project) > validation.DNS1123SubdomainMaxLength {
+		project = project[:validation.DNS1123SubdomainMaxLength]
+	}
```
jannfis: I'm wondering whether it's a good idea to cut the name at maximum length, or whether it's better to error out when the name exceeds a given length.

I can imagine all kinds of weird side-effects when there's two or more projects with too long names, all ending up to be "the same" after the length-cut.

agent/inbound.go
```diff
@@ -475,14 +474,17 @@ func (a *Agent) createAppProject(incoming *v1alpha1.AppProject) (*v1alpha1.AppPr
 }
 
 func (a *Agent) updateAppProject(incoming *v1alpha1.AppProject) (*v1alpha1.AppProject, error) {
-	//
-	incoming.SetNamespace(a.namespace)
 	logCtx := log().WithFields(logrus.Fields{
 		"method":          "UpdateAppProject",
 		"app":             incoming.Name,
```
jannfis: Unrelated, by I think `app` should be `proj` here :)

agent/outbound.go
```diff
@@ -150,3 +151,115 @@ func (a *Agent) deleteNamespaceCallback(outbound *corev1.Namespace) {
...
+
+	if err := a.projectManager.Manage(appProject.Name); err != nil {
+		logCtx.Errorf("Could not manage appProject: %v", err)
+		return
+	}
+
+	// Only send the creation event when we're in autonomous mode
+	if !a.mode.IsAutonomous() {
+		return
+	}
```
jannfis: Should this go to earlier in the execution workflow?

agent/outbound.go
```diff
@@ -150,3 +151,115 @@ func (a *Agent) deleteNamespaceCallback(outbound *corev1.Namespace) {
...
+	// If the appProject is not managed, we ignore this event.
+	if !a.projectManager.IsManaged(new.Name) {
+		logCtx.Errorf("Received update event for unmanaged appProject")
+		return
+	}
+
+	// Only send the update event when we're in autonomous mode
+	if !a.mode.IsAutonomous() {
+		return
+	}
```
jannfis: Same, should this go earlier in execution workflow?

agent/outbound.go
```diff
@@ -150,3 +151,115 @@ func (a *Agent) deleteNamespaceCallback(outbound *corev1.Namespace) {
...
+	if !a.projectManager.IsManaged(appProject.Name) {
+		logCtx.Warn("AppProject is not managed, proceeding anyways")
+	} else {
+		_ = a.projectManager.Unmanage(appProject.Name)
+	}
+
+	// Only send the deletion event when we're in autonomous mode
+	if !a.mode.IsAutonomous() {
+		return
+	}
```
jannfis: Same.

principal/callbacks.go
```diff
@@ -143,6 +146,11 @@ func (s *Server) newAppProjectCallback(outbound *v1alpha1.AppProject) {
 
 	s.resources.Add(outbound.Namespace, resources.NewResourceKeyFromAppProject(outbound))
 
+	if outbound.Annotations[appproject.AppProjectAgentModeAnnotation] == string(types.AgentModeAutonomous) {
+		logCtx.Debugf("Discarding event, because the appProject is managed by an autonomous agent")
+		return
+	}
```
jannfis: Is there another, non user-modifiable way to find out whether an AppProject is destined for a managed or autonomous agent?

principal/event.go
chetan-rns: I've updated the function to return an error if the name exceeds the maximum length.

principal/callbacks.go
chetan-rns: I've updated the code to check if the project name is prefixed with the agent's name. This approach is much simpler. If a user tries to create a project prefixed with the autonomous agent's name, the event will be discarded

I also thought of storing a map of projects to the corresponding agent's mode. But we will lose that information when the principal restarts.

jannfis (approved):
LGTM

Fantastic job, @chetan-rns

PR #432, opened 2025-06-26
chore: Remove-mounted-tls-secrets

### LINKED ISSUE #431: agent and principal try to read crt/key/ca values from volume path by default
**Describe the bug**
in a clean installation of both the agent and the principal, the params-cm contain values that cause the Pod to try to load the certificate, key and ca data from local volume by default.
the deployment mounts those files from a hard-coded Secret (`argocd-agent-tls`).
only the agent has that volume marked as optional.

with the cm param values set to `""`, the code falls back into reading the secrets from k8s api.
removing the volume + volumeMounts from both deployment allows the Pods to start even when the secrets are not in the cluster.

1. install the default principal using kubernetes
2. the Pod fails to start because `argocd-agent-tls` Secret doesn't exist

**Expected behavior**
the Pod should start, and then try to access the `argocd-agent-principal-tls` and `argocd-agent-ca` Secrets from the code.

ATGardner:
Remove hardcoded TLS certificate paths in agent and principal
ConfigMaps to enhance security and avoid potential misconfigurations.
Set default values to empty strings for TLS-related fields.

Fixes #431

a clean installation of either the principal or an agent should only require `argocd-agent-principal-tls` or `argocd-agent-client-tls` with crt/key fields, and `argocd-agent-ca` with crt field.

**Checklist**

### DIFF
--- a/install/kubernetes/agent/agent-deployment.yaml
+++ b/install/kubernetes/agent/agent-deployment.yaml
@@ -103,8 +103,6 @@ spec:
             seccompProfile:
               type: RuntimeDefault
           volumeMounts:
-            - name: tls-secret
-              mountPath: /app/config/tls
             - name: userpass-passwd
               mountPath: /app/config/creds
       serviceAccountName: argocd-agent-agent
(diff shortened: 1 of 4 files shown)


jannfis (dismissed):
LGTM! Thanks!

jannfis (approved):

PR #436, opened 2025-06-27
feat: proxy redis traffic from principal to agents

### LINKED ISSUE #344: Desired manifests view to be viewable on the control plane

Right now, the Argo CD UI can only display desired manifests and render diff if it has access to the Redis cache. This means, the agents need to store their Redis data on a server running on the control plane. However, this subverts some of the design principles and makes the control plane a single point of failure.

We need the control plane to access data stored in Redis on the workload clusters. Initial ideas are to use a similar method as we use for the resource proxy to access live resources, just for the Redis component.

**Additional context**

jgwest:

Implementation TL;DR:
- Principal agent now contains a redis proxy, which acts as an intermediary/broker for redis traffic.
- Principal Argo CD connects to principal redis proxy, rather than directly to Principal Argo CD Redis.
- Principal agent listens and reads for traffic on redis port:
	- If redis commands `subscribe`, or `get` are received from principal Argo CD:
		- AND those commands are subscribe/get specifically for an argo agent, then redirect message to agent via event message
		- Otherwise, forward to principal redis (and return response to Argo CD)
	- For all other redis commands, forward it to principal redis (and return response to Argo CD)
- Agents running on agent clusters will receive redis messages from principal, see above.
	- These will be either get or subscribe messages.
- These messages are forwarded to agent redis, and the response returns back through agent event machinery

Notes:
- I cloned `principal/resourceproxy/tracking.go` into a new package, `principal/tracking`.
	- This tracking functionality is useful for tracking redis sync calls as well.
	- I think we should refactor resourceproxy to use `principal/tracking` (since it's the same API), and delete the original file, but I've left that out of this PR to reduce merge friction.
	- Happy to do it as part of a separate PR, presuming you all agree.
- More info on Redis wire protocol is [here](https://redis.io/docs/latest/develop/reference/protocol-spec/#sets)
- I had to disable parallel running of go tests for `make test`: 
    - Why? There are now multiple separate packages that call `(principal) Server.Start()`, which listens on Redis port, which causes 'port already in use' errors. (`pkg/client/remote_test.go` and `principal/listen_test.go`).

Fixes #344

**Checklist**

### DIFF
--- a/agent/inbound.go
+++ b/agent/inbound.go
@@ -60,8 +60,16 @@ func (a *Agent) processIncomingEvent(ev *event.Event) error {
 		err = a.processIncomingResourceRequest(ev)
 	case event.TargetResourceResync:
 		err = a.processIncomingResourceResyncEvent(ev)
+	case event.TargetRedis:
+		go func() {
+			// Process request in a separate go routine, to avoid blocking the event thread on redis I/O
+			err := a.processIncomingRedisRequest(ev)
+			if err != nil {
+				log().WithError(err).Errorf("Unable to process incoming redis event")
+			}
+		}()
 	default:
-		err = fmt.Errorf("unknown event target: %s", ev.Target())
+		err = fmt.Errorf("unknown event target - processIncomingEvent: %s", ev.Target())
 	}
 
 	cp.End()
--- a/agent/inbound_redis.go
+++ b/agent/inbound_redis.go
@@ -1,3 +1,17 @@
+// Copyright 2025 The argocd-agent Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package agent
 
 import (
@@ -47,6 +61,11 @@ type connectionEntry struct {
 	// lastPing the time at which the last ping message was received from principal
 	lastPing time.Time
 }
+
+const (
+	// principalRedisConnectionTimeout is the length of time the agent will wait to receive redis communication from principal, before it will close the connection. e.g. if we receive no requests for 10 minutes, we receive no heartbeats (pings) the connection is no longer required.
+	principalRedisConnectionTimeout = 10 * time.Minute
+)
 
 // processIncomingRedisRequest handles incoming redis-specific events from principal, including get/subscribe (from argo cd) and ping (internal).
 func (a *Agent) processIncomingRedisRequest(ev *event.Event) error {
@@ -100,12 +119,12 @@ func (a *Agent) processIncomingRedisRequest(ev *event.Event) error {
 
 	} else {
 		logCtx.Error("Unrecognized redis request body")
-		return nil
+		return fmt.Errorf("unrecognized redis request body")
 	}
 
 	if responseBody == nil {
-		logCtx.Error("Response command not defined")
-		return nil
+		logCtx.Error("unexpected lack of response to redis request: response command not defined")
+		return fmt.Errorf("unexpected lack of response to redis request")
 	}
 
 	q := a.queues.SendQ(defaultQueueName)
@@ -129,7 +148,6 @@ func (a *Agent) handleRedisSubscribeMessage(logCtx *logrus.Entry, rreq *event.Re
 		var err error
 		channelName, err = stripNamespaceFromKeyForAutonomousAgentForRedis(channelName, logCtx)
 		if err != nil {
-			logCtx.WithError(err).Errorf("unable to transform SUBSCRIBE key for autonomous agent '%s'", channelName)
 			return nil, fmt.Errorf("unable to transform SUBSCRIBE key for autonomous agent: %v", err)
 		}
 	}
@@ -192,7 +210,7 @@ func (a *Agent) forwardRedisSubscribeNotificationsToPrincipal(pubsub *redis.PubS
 			}
 
 			// If the connection has not been active for X minutes, close the connection
-			if time.Since(*lastPing) >= time.Minute*10 {
+			if time.Since(*lastPing) >= principalRedisConnectionTimeout {
 				pubsub.Close()
 				logCtx.WithField("lastPing", lastPing).Trace("closing redis connection due to inactivity")
 				return
@@ -236,7 +254,6 @@ func (a *Agent) handleRedisGetMessage(logCtx *logrus.Entry, rreq *event.RedisReq
 		var err error
 		key, err = stripNamespaceFromKeyForAutonomousAgentForRedis(key, logCtx)
 		if err != nil {
-			logCtx.WithError(err).Errorf("unable to transform GET key for autonomous agent '%s'", key)
 			return nil, fmt.Errorf("unable to transform GET key for autonomous agent: %v", err)
 		}
 	}
--- a/agent/inbound_redis.go
+++ b/agent/inbound_redis.go
@@ -1,0 +1,341 @@
+package agent
+
+import (
+	"context"
+	"crypto/tls"
+	"errors"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/argoproj-labs/argocd-agent/internal/event"
+	"github.com/argoproj-labs/argocd-agent/pkg/types"
+	rediscache "github.com/go-redis/cache/v9"
+	"github.com/redis/go-redis/v9"
+	"github.com/sirupsen/logrus"
+)
+
+// redisProxyMsgHandler manages redis connection state for agent
+type redisProxyMsgHandler struct {
+	redisAddress  string
+	redisUsername string
+	redisPassword string
+
+	// argoCDRedisClient is the client API to agent redis
+	argoCDRedisClient *redis.Client
+
+	// argoCDRedisCache is the cache API to agent redis
+	argoCDRedisCache *rediscache.Cache
+
+	// connections maintains statistics about redis connections from principal
+	connections *connectionEntries
+}
+
+// connectionEntries maintains statistics about redis connections from principal
+type connectionEntries struct {
+	// lock should be acquired before reading/writing to connMap
+	lock sync.RWMutex
+
+	// connMap maintains a list of when a ping was last seen from each connection (uuid). This allows us to close stale connections.
+	// - key is connection uuid
+	connMap map[string]connectionEntry // acquire lock before accessing
+}
+
+// connectionEntry describes statistics about a specific connection (uuid)
+type connectionEntry struct {
+	// lastPing the time at which the last ping message was received from principal
+	lastPing time.Time
+}
+
+// processIncomingRedisRequest handles incoming redis-specific events from principal, including get/subscribe (from argo cd) and ping (internal).
+func (a *Agent) processIncomingRedisRequest(ev *event.Event) error {
+
+	rreq, err := ev.RedisRequest()
+	if err != nil {
+		return err
+	}
+	logCtx := log().WithFields(logrus.Fields{
+		"method":         "processIncomingRedisRequest",
+		"uuid":           rreq.UUID,
+		"connectionUUID": rreq.ConnectionUUID,
+	})
+
+	logCtx.Tracef("Start processing incoming redis request %v", rreq)
+
+	var responseBody *event.RedisResponseBody
+
+	if rreq.Body.Get != nil {
+		var err error
+
+		responseBody, err = a.handleRedisGetMessage(logCtx, rreq)
+		if err != nil {
+			return err
+		}
+
+	} else if rreq.Body.Subscribe != nil {
+
+		var err error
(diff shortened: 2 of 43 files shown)


hack/dev-env/start-e2e2.sh
```diff
@@ -0,0 +1,53 @@
...
+# getExternalLoadBalancerIP will set EXTERNAL_IP with the load balancer hostname from the specified Service
+getExternalLoadBalancerIP() {
+  SERVICE_NAME=$1
+
+  MAX_ATTEMPTS=120
+
+  for ((i=1; i<=MAX_ATTEMPTS; i++)); do
+    
+    echo ""
+    EXTERNAL_IP=$(kubectl get svc $SERVICE_NAME $K8S_CONTEXT $K8S_NAMESPACE -o jsonpath='{.status.loadBalancer.ingress[0].hostname}')
```
jannfis: On my local system, the hostname field is not set for the ingress. It looks like the following:

```
kubectl --context vcluster-control-plane get svc -n argocd argocd-redis -o jsonpath='{.status.loadBalancer}' | jq
{
  "ingress": [
    {
      "ip": "192.168.57.221",
      "ipMode": "VIP"
    }
  ]
}
```
We probably should accommodate both, ip and hostname cases.

jannfis:
@jgwest: I made a couple of smaller changes to manifests and scripts to make it work in my local environment. 

I could either push these directly to your PR's branch, submit a PR against the branch in your fork or send you a diff.

Which of those do you prefer? :)

jgwest:
Switching to draft for E2E test debugging.

hack/test.sh
```diff
@@ -16,4 +16,4 @@
 # Wrapper script to run unit tests for argocd-agent
 set -eo pipefail
 UNIT_LIST=$(go list ./... | grep -E -v '(mock|mocks|e2e|fake|pkg/api/grpc)')
-go test -race -timeout 10m -coverprofile test/out/coverage.out $UNIT_LIST
+go test -race -p 1 -timeout 10m -coverprofile test/out/coverage.out $UNIT_LIST
```
jgwest, line 19: See PR description for why `-p 1`

agent/inbound.go
```diff
@@ -60,8 +60,16 @@ func (a *Agent) processIncomingEvent(ev *event.Event) error {
...
 	case event.TargetResourceResync:
 		err = a.processIncomingResourceResyncEvent(ev)
+	case event.TargetRedis:
+		go func() {
+			// Process request in a separate go routine, to avoid blocking the event thread on redis I/O
+			err := a.processIncomingRedisRequest(ev)
+			if err != nil {
+				log().WithError(err).Errorf("Unable to process incoming redis event")
+			}
+		}()
```
jannfis, line 70: I'm wondering whether this is overkill, or if the whole `processIncomingEvent` function should be executed in its own go routine :thinking:

agent/inbound_redis.go
```diff
@@ -0,0 +1,341 @@
+package agent
```
jannfis, line 15: Please add license header

agent/inbound_redis.go
```diff
@@ -0,0 +1,341 @@
...
+
+		// Send pong back to principal
+		responseBody = &event.RedisResponseBody{
+			Pong: &event.RedisResponseBodyPong{},
+		}
+		logCtx.Tracef("Queuing Redis pong response")
+
+	} else {
+		logCtx.Error("Unrecognized redis request body")
+		return nil
```
jannfis: Why is this condition (and similar ones below) logged as an error, but not handled as one (i.e. nil returned instead of error)?

agent/inbound_redis.go
```diff
@@ -0,0 +1,341 @@
...
+
+	logCtx.Tracef("Start processing redis SUBSCRIBE request with key '%s'", channelName)
+
+	// The autonomous Redis does not use namespace in key, so we need to strip it. See strip function for details.
+	if a.mode == types.AgentModeAutonomous {
+		var err error
+		channelName, err = stripNamespaceFromKeyForAutonomousAgentForRedis(channelName, logCtx)
+		if err != nil {
+			logCtx.WithError(err).Errorf("unable to transform SUBSCRIBE key for autonomous agent '%s'", channelName)
+			return nil, fmt.Errorf("unable to transform SUBSCRIBE key for autonomous agent: %v", err)
```
jannfis: The log statement here is redundant IMO, and the caller of this function should log the error instead.

agent/inbound_redis.go
```diff
@@ -0,0 +1,341 @@
...
+			}
+			connLifecycle.lock.RUnlock()
+
+			if lastPing == nil {
+				logCtx.Error("lastPing for a connection should never be nil")
+				continue
+			}
+
+			// If the connection has not been active for X minutes, close the connection
+			if time.Since(*lastPing) >= time.Minute*10 {
```
jannfis: This would make a great constant, and potentially a future configurable :)

agent/inbound_redis.go
```diff
@@ -0,0 +1,341 @@
...
+	key := rreq.Body.Get.Key
+
+	logCtx.Tracef("Start processing redis GET request with key '%s'", key)
+
+	if a.mode == types.AgentModeAutonomous {
+		var err error
+		key, err = stripNamespaceFromKeyForAutonomousAgentForRedis(key, logCtx)
+		if err != nil {
+			logCtx.WithError(err).Errorf("unable to transform GET key for autonomous agent '%s'", key)
+			return nil, fmt.Errorf("unable to transform GET key for autonomous agent: %v", err)
```
jannfis: The log statement here is redundant IMO, and the caller of this function should log the error instead.

agent/inbound_redis_test.go
```diff
@@ -0,0 +1,283 @@
+package agent
```
jannfis, line 15: Please add a license header

cmd/argocd-agent/agent.go
```diff
@@ -196,6 +211,27 @@ func NewAgentRunCommand() *cobra.Command {
...
+		env.StringWithDefault("REDIS_PASSWORD", nil, ""),
+		"The password to connect to redis with")
+
+	command.Flags().StringVar(&autonomousAgentRedisHost, "autonomous-agent-redis-host",
+		env.StringWithDefault("AUTONOMOUS_AGENT_REDIS_ADDR", nil, ""),
+		"The redis address to use when running in autonomous mode")
+
+	command.Flags().StringVar(&managedAgentRedisHost, "managed-agent-redis-host",
+		env.StringWithDefault("MANAGED_AGENT_REDIS_ADDR", nil, ""),
+		"The redis address to use when running in managed mode")
```
jannfis: I don't understand why they need to be different. Shouldn't they use the local argocd-redis service regardless of the agent mode? And what is the difference to the `redis-addr` parameter? Maybe I have a misunderstanding here.

hack/dev-env/start-e2e2.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
```
jannfis, line 1: Please add license header

principal/event.go
```diff
@@ -281,6 +297,91 @@ func (s *Server) processAppProjectEvent(ctx context.Context, agentName string, e
...
+		logCtx = logCtx.WithField("channelName", resReq.Body.PushFromSubscribe.ChannelName)
+
+		logCtx.Tracef("handling PushFromSubscribe, retrieving tracked channel")
+
+		// PushFromSubscribe is asynchronous, so we can't use event id tracker to find the corresponding channel.
+		// - We instead use connection id, which maps directly to the connection from Argo CD
+		trAgent, evChan := s.redisProxy.ConnectionIDTracker.Tracked(resReq.ConnectionUUID)
+		if trAgent != agentName {
+			err := fmt.Errorf("agent mismap between event and tracking for PushFromSubscribe: '%s' '%s' '%s'", trAgent, agentName, resReq.ConnectionUUID)
+			logCtx.WithError(err).Error("agent mismap between event and tracking")
```
jannfis: The log statement here is redundant IMO, and the caller of this function should log the error instead.

principal/event.go
```diff
@@ -281,6 +297,91 @@ func (s *Server) processAppProjectEvent(ctx context.Context, agentName string, e
...
+	logCtx = logCtx.WithField("eventID", resReq.UUID)
+
+	// For all other responses, we send the event on the eventIDTracker channel
+
+	// We need to make sure that a) the event is tracked at all, and b) the
+	// event is for the currently processed agent.
+	trAgent, evChan := s.redisProxy.EventIDTracker.Tracked(resReq.UUID)
+	if evChan == nil {
+		logCtx.Errorf("redis response not tracked for eventID")
+		return fmt.Errorf("redis response not tracked")
```
jannfis, line 362: The log statement here is redundant IMO, and the caller of this function should log the error instead.

principal/event.go
```diff
@@ -281,6 +297,91 @@ func (s *Server) processAppProjectEvent(ctx context.Context, agentName string, e
...
+	}
+	if trAgent != agentName {
+		logCtx.Errorf("agent mismap between event and tracking: '%s' '%s'", trAgent, agentName)
+		return fmt.Errorf("agent mismap between event and tracking: '%s' '%s'", trAgent, agentName)
```
jannfis, line 366: The log statement here is redundant IMO, and the caller of this function should log the error instead.

principal/redisproxy/redisproxy.go
```diff
@@ -0,0 +1,806 @@
+package redisproxy
```
jannfis, line 15: Please add license header

principal/redisproxy/redisproxy.go
```diff
@@ -0,0 +1,806 @@
...
+		if redisKey == "new-revoked-token" {
+			// 'new-revoked-token' is forwarded to principal
+			return "", nil
+		}
+
+		if strings.HasPrefix(redisKey, "mfst|") {
+			// mfst| is forwarded to principal, but I think this may not be a permanent behaviour.
+			// Example key:
+			// - mfst|annotation:app.kubernetes.io/instance|agent-managed_my-app|(revision id)|guestbook|3093507789|1.8.3.gz
+			logCtx.Info("redirecting mfst| redis key to principal redis")
```
jannfis: This message should be debug level imo

principal/redisproxy/redisproxy.go
```diff
@@ -0,0 +1,806 @@
...
+			return "", nil
+		}
+
+		logCtx.Errorf("Unexpected redis key seen: '%s'. Redirecting to principal by default.", redisKey)
```
jannfis: This should be a warning level rather than an error, IMO.

principal/redisproxy/redisproxy.go
```diff
@@ -0,0 +1,806 @@
...
+			return "", nil
+		}
+
+		if strings.HasPrefix(redisKey, "mfst|") {
+			// mfst| is forwarded to principal, but I think this may not be a permanent behaviour.
+			// Example key:
+			// - mfst|annotation:app.kubernetes.io/instance|agent-managed_my-app|(revision id)|guestbook|3093507789|1.8.3.gz
+			logCtx.Info("redirecting mfst| redis key to principal redis")
+			return "", nil
+		}
```
jannfis, line 788: There is at least one more expected key name that should be redirected to the principal's Redis:

* `cluster|` (all related to cluster configuration)

principal/redisproxy/util.go
```diff
@@ -0,0 +1,132 @@
+package redisproxy
```
jannfis, line 15: Please add licence header

jannfis (commented):
Thanks! This is awesome, and I've tested it thoroughly on my local system.

I have mostly nits and some cosmetics.

agent/inbound_redis.go
jgwest: Good catch,  no good reason, I'll add an error return.

agent/inbound_redis.go
jgwest: Done, added `principalRedisConnectionTimeout` const

principal/event.go
jgwest: For  other cases on this PR you have flagged I agree, but for this one I log here because `logCtx` contains additional structured logging fields (`agent-name`, `connectionUUID`, etc) that are not included in the error return value (or the parent `logCtx`)

principal/event.go
jgwest: Same reasoning as above.

cmd/argocd-agent/agent.go
jgwest: I took a look and have found a better way to handle this from Procfile without using these params. They've been removed.

agent/inbound.go
jgwest: IMHO the latter: for maximum throughput/reliability, you generally don't want to block an event loop thread on waiting on external I/O (connecting to redis, connecting to K8s API server, etc). 

BUT, before we switch to processing any OTHER request type on goroutines, we should carefully consider the potential that it would cause messages to be processed out of order.

Switching to processing on go routines can affect the order that messages are processed (due to scheduling), and generally for our needs we want messages to be processed in chronological order.

For redis agent messages, at least, I have verified it is safe to schedule on a go routine (without affecting order) since the protocol is fully synchronous. (This may ALSO be true for the other message types we process here, but we would first need to check)

jannfis: Yeah, we can re-evaluate this later. Thanks for your thoughts on this!

jannfis (approved):
Awesome! Thanks for this incredible PR, @jgwest.

LGTM.

PR #438, opened 2025-07-02
fix: Make volume mounts optional for agent and principal deployments

jannfis:

The agent and the principal both can read credentials from a file. For the pods, we mount these files from a secret.

However, this has changed from being mandatory to optional, since both can read directly from Kubernetes secrets now. For this reason, the volume mounts will now be optional, too.

Fixes #?

**Checklist**


chetan-rns (approved):
LGTM!

PR #440, opened 2025-07-03
fix(cli): Allow DNS names for resource proxy host in agent config

jannfis:

The `agent create` command of the `argocd-agentctl` CLI takes a `--resource-proxy-server` parameter to specify the location of the resource server. The value given is validated, and needed to be a valid IP address and port combination. However, in the cluster, the hostname is most likely to be specified as DNS name, i.e. the name of the service of the resource proxy's endpoint.

Fixes #?

Create an agent configuration that points the RP to a DNS name, e.g.

```
argocd-agentctl agent create --resource-proxy-server resource-proxy:9090 agent-test
```

**Checklist**

### DIFF
--- a/cmd/ctl/agent.go
+++ b/cmd/ctl/agent.go
@@ -443,7 +443,7 @@ func serverURL(address, agentName string) (string, error) {
 		}
 		addr := tok[0]
 		port := tok[1]
-		if len(validation.NameIsDNSLabel(addr, false)) > 0 {
+		if len(validation.NameIsDNSSubdomain(addr, false)) > 0 {
 			return "", fmt.Errorf("invalid address: %s", address)
 		}
 		if _, err := strconv.ParseUint(port, 10, 16); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/ctl/agent.go
```diff
@@ -434,7 +436,19 @@ func clusterSecretName(agentName string) string {
...
 	if err != nil {
-		return "", err
+		// We now have a host:port address, so we need to validate the host
+		tok := strings.SplitN(address, ":", 2)
+		if len(tok) != 2 {
+			return "", fmt.Errorf("invalid address: %s", address)
+		}
+		addr := tok[0]
+		port := tok[1]
+		if len(validation.NameIsDNSLabel(addr, false)) > 0 {
```
chetan-rns: This returns an error for hostnames with a `.` dot. e.g. `example.com`. Should we use `validation.IsDNS1123Subdomain` instead?

jannfis: Good catch! Yes, validating against the subdomain expression is appropriate.

Let me write some unit tests too for this function.

PR #449, opened 2025-07-10
fix: Correctly set destination cluster for autonomous apps

### LINKED ISSUE #448: Live resource view doesn't work for apps on autonomous agents
**Describe the bug**

When an application is created on an autonomous agent, it will be synced back to the principal. A few properties of the Application's spec will be transformed. For example, the namespace on the agent system and the principal differ.

One thing that should be transformed but is currently not is the destination cluster. On the agent system, it is always `in-cluster`, but on the control plane it should be transformed to point to the resource proxy's pseudo-cluster secret to enable resource proxy for autonomous agents.

1. Create Application on autonomous agent system
2. Observe Application being synced to the principal
3. Inspect Application on the principal to see destination cluster is still `in-cluster`.

**Expected behavior**

The destination cluster of the Application on the principal system should match the cluster URL and name that has been created for the appropriate autonomous agent.

**Additional context**

jannfis:

When an autonomous agent sends a create or update event for an Application, the destination field on the principal must be correctly set to the agent or the live resource view (and other features) wouldn't work.

Fixes #448 

1. Create Application on the autonomous agent system
2. Observe creation of the Application on the control plane
3. Open the Application's resource tree in the Argo CD UI
4. Click on one of the resources to retrieve the live resource from the agent's system

**Checklist**


jgwest (approved):
LGTM!

Issue #422, opened 2025-06-24
Flaky e2e test: TestCacheTestSuite/Test_RevertManagedClusterChanges


jannfis:
**Describe the bug**

Flaky end-to-end test: TestCacheTestSuite/Test_RevertManagedClusterChanges

**Steps to reproduce the behaviour**

Passed on PR, but failed once merged to main.

**Expected behavior**

Test passing in main branch.

**Additional context**

```
=== RUN   TestCacheTestSuite/Test_RevertManagedClusterChanges
    fixture.go:68: Test begun at: 2025-06-24 12:20:48.608874283 +0000 UTC m=+41.538069701
    cache_test.go:194: 
        	Error Trace:	/home/runner/work/argocd-agent/argocd-agent/test/e2e2/cache_test.go:194
        	            				/home/runner/work/argocd-agent/argocd-agent/test/e2e2/cache_test.go:53
        	Error:      	Condition never satisfied
        	Test:       	TestCacheTestSuite/Test_RevertManagedClusterChanges
    fixture.go:72: Test ended at: 2025-06-24 12:22:18.622310882 +0000 UTC m=+131.551506300
=== RUN   TestCacheTestSuite/Test_RevertManagedClusterOfflineChanges
    fixture.go:68: Test begun at: 2025-06-24 12:22:18.704906263 +0000 UTC m=+131.634101681
    fixture.go:72: Test ended at: 2025-06-24 12:22:34.78787182 +0000 UTC m=+147.7170[67](https://github.com/argoproj-labs/argocd-agent/actions/runs/15850024435/job/44680962291#step:14:68)228
--- FAIL: TestCacheTestSuite (141.13s)
    --- PASS: TestCacheTestSuite/Test_CacheRecreatedOnRestart (17.36s)
    --- PASS: TestCacheTestSuite/Test_RevertDisconnectedManagedClusterChanges (16.43s)
    --- FAIL: TestCacheTestSuite/Test_RevertManagedClusterChanges (90.09s)
    --- PASS: TestCacheTestSuite/Test_RevertManagedClusterOfflineChanges (16.14s)
```

jparsai:
I will wait for few PRs before closing this issue to ensure E2E tests are not having this issue anymore.

jparsai:
Closing this issue, since error is not seen anymore in PRs, feel free to reopen if it reappears.

PR #456, opened 2025-07-14
feat: add health check endpoint for principal

jparsai:
This PR is to add health check endpoint to principal, which can be used to verify server readiness. This would fix the issue https://github.com/argoproj-labs/argocd-agent/issues/422, where principal is not fully started but tests still executed which causes test failure.
This PR also exposes health check via a service for both Principal and Agent.

### DIFF
--- a/test/e2e2/cache_test.go
+++ b/test/e2e2/cache_test.go
@@ -28,17 +28,43 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+const (
+	PrincipalName    = "principal"
+	AgentManagedName = "agent-managed"
+)
+
 type CacheTestSuite struct {
 	fixture.BaseSuite
 }
 
 func TestCacheTestSuite(t *testing.T) {
-	fixture.WaitForAgent(t, "agent-managed")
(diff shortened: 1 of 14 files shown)


test/e2e2/cache_test.go
```diff
@@ -28,19 +28,45 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+const (
+	PrincipalName     = "principal"
+	AgentName_Managed = "agent-managed"
```
jannfis: Please do not use underscores in symbol naming (https://go.dev/doc/effective_go#mixed-caps)

jannfis (commented):
Thanks, I only have a minor nit otherwise LGTM.

We can also expose this as a service, so that the cluster can make use of the health check too.

jparsai:
> Thanks, I only have a minor nit otherwise LGTM.
> 
> We can also expose this as a service, so that the cluster can make use of the health check too.

Thanks Jann, I added service for it. Since health check is also available in agent, I added service for agent as well. PTAL.

jannfis (approved):
LGTM, thanks!

PR #458, opened 2025-07-17
fix: Applications on managed agent can live in argocd namespace

jannfis:

A managed agent's applications should not be confined to a namespace that matches the agent's name. The applications should really live in the same namespace as the Argo CD instance, just like they do for  autonomous agents.

This also removes the requirement for the Argo CD on a managed agent to be configured with apps-in-any-namespace (even if "any" in this case would just mean  "one other" aka "the agent's namespace").

It will simplify and streamline both, the install/setup of the agent and future code contributions because the difference between autonomous and managed agents will be smaller.

Fixes #?

**Checklist**

### DIFF
--- a/agent/inbound_redis.go
+++ b/agent/inbound_redis.go
@@ -144,7 +144,7 @@ func (a *Agent) handleRedisSubscribeMessage(logCtx *logrus.Entry, rreq *event.Re
 
 	channelName, err := stripNamespaceFromRedisKey(channelName, logCtx)
 	if err != nil {
-		return nil, fmt.Errorf("unable to transform SUBSCRIBE key for autonomous agent: %v", err)
+		return nil, fmt.Errorf("unable to transform SUBSCRIBE key for agent: %v", err)
 	}
 
 	// Subscribe to the argo cd redis channel via redis client
@@ -247,7 +247,7 @@ func (a *Agent) handleRedisGetMessage(logCtx *logrus.Entry, rreq *event.RedisReq
 
 	key, err := stripNamespaceFromRedisKey(key, logCtx)
 	if err != nil {
-		return nil, fmt.Errorf("unable to transform GET key for autonomous agent: %v", err)
+		return nil, fmt.Errorf("unable to transform GET key for agent: %v", err)
 	}
 
 	// Connect to local redis to GET key/value, and store response


mikeshng:
Jann, my managed agent has always be running in the `argocd` ns on the spoke side. The differences I can think of are 1) no resource proxy yet 2) no redis proxy 3) rbac differences(?).

Do you mean on the hub side, the app CR no longer needs to be in the ns that is the same as the queue name?

jannfis:
I saw the description and PR title was a little confusing, sorry about that. I have edited both to better reflect the true intentions of this fix.

Thanks for noticing @mikeshng !

agent/inbound_redis.go
```diff
@@ -143,13 +142,9 @@ func (a *Agent) handleRedisSubscribeMessage(logCtx *logrus.Entry, rreq *event.Re
...
-	// The autonomous Redis does not use namespace in key, so we need to strip it. See strip function for details.
-	if a.mode == types.AgentModeAutonomous {
-		var err error
-		channelName, err = stripNamespaceFromKeyForAutonomousAgentForRedis(channelName, logCtx)
-		if err != nil {
-			return nil, fmt.Errorf("unable to transform SUBSCRIBE key for autonomous agent: %v", err)
-		}
+	channelName, err := stripNamespaceFromRedisKey(channelName, logCtx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to transform SUBSCRIBE key for autonomous agent: %v", err)
```
jgwest: `for autonomous agent` -> `for agent`

agent/inbound_redis.go
```diff
@@ -250,19 +245,16 @@ func (a *Agent) handleRedisGetMessage(logCtx *logrus.Entry, rreq *event.RedisReq
...
 
-	if a.mode == types.AgentModeAutonomous {
-		var err error
-		key, err = stripNamespaceFromKeyForAutonomousAgentForRedis(key, logCtx)
-		if err != nil {
-			return nil, fmt.Errorf("unable to transform GET key for autonomous agent: %v", err)
-		}
+	key, err := stripNamespaceFromRedisKey(key, logCtx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to transform GET key for autonomous agent: %v", err)
```
jgwest: `for autonomous agent` -> `for agent`

test/e2e2/cache_test.go
```diff
@@ -206,33 +211,39 @@ func createApp(ctx context.Context, client fixture.KubeClient, requires *require
...
 
 func updateAppInfo(ctx context.Context, client fixture.KubeClient, key types.NamespacedName, info []string, requires *require.Assertions) {
-	requires.NoError(client.EnsureApplicationUpdate(ctx, key, func(app *argoapp.Application) error {
-		app.Spec.Info = []argoapp.Info{
-			{
-				Name:  info[0],
-				Value: info[1],
-			},
+	requires.Eventually(func() bool {
+		err := client.EnsureApplicationUpdate(ctx, key, func(app *argoapp.Application) error {
```
jgwest, line 233: It seems like the `requires.Eventually()`-style logic is already in `EnsureApplicationUpdate`? Unless there is some other case this is handling.

jannfis: I had some runs where the `Get` within `EnsureApplicationUpdate` returned a 404 occasionally, thus the function failed (it will only recover when the update fails with a conflict).

agent/inbound_redis.go
jannfis: Thanks! Changed it.

agent/inbound_redis.go
jannfis: Thanks! Changed it.

PR #462, opened 2025-07-18
fix: cascaded delete on Application should clean up all managed-agent app resources

### LINKED ISSUE #443: Cascading delete does not work (anymore)
**Describe the bug**

Deleting an Application on the principal should propagate any potential deletion strategy to the agent. This used to work in the past. Right now, the deletion strategy - especially the resources finalizer for cascading delete - does not seem to work. The Application on both, the agent and principal, is deleted, but its resources are left behind on the agent system.

1. Create the guestbook Application on the principal, targeting a managed agent (unsure/untested if the same is true for autonomous agents)
2. Sync the guestbook Application and wait for it to be synced on the agent system
3. Delete the Application on the principal system, using cascading delete
4. Observe the Application on both principal and agent are being deleted
5. Observe the resources of the guestbook Application still exist on the agent's system

**Expected behavior**
Application resources on the agent system being deleted as well.

**Additional context**

jgwest:
- At present, if you perform a cascade delete on an Argo CD Application from principal, the child resources of that Application will not be deleted on managed-agent.
- This is due to the following, which have been rectified: 
    - The Argo CD Application deletion `.metadata.finalizer` was being removed by principal
    - Deletion events for managed-agent were not being sent by agent
    - Deletion events for managed-agent were not being processed by principal
- Plus updated unit/E2E tests

Fixes #443

### DIFF
--- a/agent/outbound_test.go
+++ b/agent/outbound_test.go
@@ -146,6 +146,10 @@ func Test_addAppDeletionToQueue(t *testing.T) {
 		require.False(t, a.appManager.IsManaged("agent/guestbook"))
 	})
 	t.Run("Deletion event for unmanaged application", func(t *testing.T) {
+		a := newAgent(t)
+		a.remote.SetClientID("agent")
+		a.mode = types.AgentModeAutonomous
+
 		app := &v1alpha1.Application{ObjectMeta: v1.ObjectMeta{Name: "guestbook", Namespace: "agent"}}
 		// App must be already managed for event to be generated
 		a.addAppDeletionToQueue(app)
--- a/agent/outbound_test.go
+++ b/agent/outbound_test.go
@@ -138,10 +138,11 @@ func Test_addAppDeletionToQueue(t *testing.T) {
 		defer func() {
 			a.mode = types.AgentModeAutonomous
 		}()
-		// App must be already managed for event to be generated
+		// Delete events will be sent for managed applications on managed agent, whether or not the application is managed.
 		_ = a.appManager.Manage("agent/guestbook")
-		a.addAppDeletionToQueue(app)
 		require.Equal(t, 0, a.queues.SendQ(defaultQueueName).Len())
+		a.addAppDeletionToQueue(app)
+		require.Equal(t, 1, a.queues.SendQ(defaultQueueName).Len())
 		require.False(t, a.appManager.IsManaged("agent/guestbook"))
 	})
 	t.Run("Deletion event for unmanaged application", func(t *testing.T) {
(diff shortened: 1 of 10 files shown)


test/e2e2/fixture/fixture.go
```diff
@@ -85,8 +85,31 @@ func EnsureDeletion(ctx context.Context, kclient KubeClient, obj KubeObject) err
...
+		err := kclient.Get(ctx, key, obj, metav1.GetOptions{})
+		if errors.IsNotFound(err) {
+			return nil
+		} else if err == nil {
+			time.Sleep(1 * time.Second)
+		} else {
+			return err
+		}
+	}
+
```
jgwest, line 102: Not strictly related to the PR, but this makes the deletion function slightly more robust by removing finalizers if we're not able to delete within 60 seconds.

agent/outbound_test.go
```diff
@@ -138,10 +138,11 @@ func Test_addAppDeletionToQueue(t *testing.T) {
 		defer func() {
 			a.mode = types.AgentModeAutonomous
 		}()
-		// App must be already managed for event to be generated
+		// Delete events will be sent for managed applications on managed agent, whether or not the application is managed.
```
jgwest, line 141: The previous comment on this test was incorrect: 
- The current product code allow apps events to be generated for non-managed app (see the function under test)
- The test claimed to prove that unmanaged applications did NOT send events, but actually it was  passing for an unrelated reason.

internal/manager/application/application.go
```diff
@@ -247,6 +259,14 @@ func (m *ApplicationManager) UpdateManagedApp(ctx context.Context, incoming *v1a
 			logCtx.Warnf("Couldn't unignore change %s for app %s: %v", updated.ResourceVersion, updated.QualifiedName(), err)
 		}
 	}
+
+	if deletionTimestampChanged {
+		logCtx.Infof("deletionTimestamp of managed agent changed from nil to non-nil, so deleting Application")
+		if err := m.applicationBackend.Delete(ctx, incoming.Name, incoming.Namespace, ptr.To(backend.DeletePropagationForeground)); err != nil {
+			return nil, err
```
chetan-rns, line 266: Any reason why we should delete the resource again if the incoming resource already has a deletion timestamp set? If we remove the finalizer, it should be deleted automatically.

jgwest: We need to put the managed-agent `Application` into deletion state, so that Argo CD/K8s starts the process of (cascade) deleting the child resources of the Application.

Once the child resources are deleted, the finalizer is removed.

The `deletionTimestamp` field is immutable, so in order to add it to a resource you need to call the K8s Delete API

PR #467, opened 2025-07-23
chore(dev-env): Tweak example/test apps to use resource proxy as destination

jannfis:

This PR modifies the example apps to a) use resource proxy as destination in case of the managed agent app, and b) both Applications to use names instead of URLs.

Fixes #?

**Checklist**


chetan-rns (approved):
LGTM!

PR #478, opened 2025-07-29
fix: Fix owner reference resolution for live resource view

### LINKED ISSUE #468: Some resources get permission denied in live resource view
**Describe the bug**
The resource proxy lets you retrieve and display resources that belong to a given managed resource. It will decide whether a resource is managed by either looking at the resource to see if it has an Argo CD resource tracking label or annotation (e.g. `app.kubernetes.io/managed-by`) or if there is an `ownerReference` that (recursively) resolves to a managed resource (e.g. if you have a `Deployment` in Git that will spawn a `ReplicaSet`, and the `ReplicaSet` will spawn one or more `Pods`, both `ReplicaSet` and any `Pods` should be considered managed).

However, it seems there is a bug in the recursion algorithm. When you deploy the example guestbook application, the `ReplicaSet` is retrievable just fine but trying to retrieve the `Pod` ends up in a 403, because the agent will think the resource is not managed.

1. Create sample guestbook application on control plane, targeting a managed agent system
2. Sync the application and wait for all resources to be synced
3. In the UI, click on the Deployment resource - you should get the live resource from the target system
4. In the UI, click on the ReplicaSet spawned by the Deployment - you should get the live resource from the target system
5. In the UI, click on the Pod spawned by the ReplicaSet - this results in error, and no live resource will be retrieved

**Expected behavior**
`Pod` live resource to be shown in the UI, because its `ownerReference` should be resolved to the `Deployment` resource through the `ReplicaSet` resource.

**Additional context**

jannfis:

This PR fixes some logic issues when resolving owner references in live resource view to determine whether a resource is managed by Argo CD or not.

Fixes #468

**Checklist**

### DIFF
--- a/agent/resource.go
+++ b/agent/resource.go
@@ -325,7 +325,7 @@ func isResourceManaged(kube *kube.KubernetesClient, res *unstructured.Unstructur
 		var gvr schema.GroupVersionResource
 		for _, apiResource := range rif.APIResources {
 			if ref.Kind == apiResource.Kind {
-				gvr = schema.GroupVersionResource{Group: apiResource.Group, Version: apiResource.Version, Resource: apiResource.Name}
+				gvr = schema.GroupVersionResource{Group: gv.Group, Version: gv.Version, Resource: apiResource.Name}
 				break
 			}
 		}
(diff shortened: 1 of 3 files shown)


chetan-rns (approved):
LGTM!

PR #483, opened 2025-07-30
fix(cli): Respect principal namespace flag when issuing certificates

jannfis:

The `pki issue` command of `argocd-agentctl` would not respect the `--principal-namespace` flag when creating the secrets holding the certificates, and instead use `argocd` as the hardcoded namespace.

Fixes #?

**Checklist**


chetan-rns (approved):
LGTM!

PR #489, opened 2025-08-01
feat: Ability to toggle resource proxy per agent

### LINKED ISSUE #480: Live resource view functionality should be toggleable

Agents do provide a feature called live resource view, which enables the principal to send an event to an agent requesting a specific resource from the cluster. Even though there are some safeguards in place limiting which resources can be requested/transmitted, some people might have policies in place which would disallow a feature like this, or want to have it disabled for any other reason.

A new command line switch should be introduced to enable or disable the live resource view feature on the agent side. When the principal transmits a resource request event to an agent which has the feature disabled, the agent should return an error immediately.

**Additional context**

jannfis:

This PR introduces a flag `--enable-resource-proxy` to the agent component. The flag defaults to `true`. If set to `false`, the agent will not provide any local resources through the resource proxy, effectively disabling the resource proxy functionality.

Fixes #480

**Checklist**

### DIFF
--- a/docs/configuration/agent/configuration.md
+++ b/docs/configuration/agent/configuration.md
@@ -241,6 +241,21 @@ The recommended approach for production deployments is to use ConfigMap entries 
 - **Default**: `""` (empty)
 - **Example**: `"redis-password"`
 
+### Resource Proxy Configuration
+
+#### Enable Resource Proxy
+- **Command Line Flag**: `--enable-resource-proxy`
+- **Environment Variable**: `ARGOCD_AGENT_ENABLE_RESOURCE_PROXY`
+- **ConfigMap Entry**: *(Not available in ConfigMap)*
+- **Description**: Enable the resource proxy to allow access to live resources on this agent cluster from the principal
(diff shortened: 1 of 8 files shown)


docs/configuration/agent/configuration.md
```diff
@@ -241,6 +241,21 @@ The recommended approach for production deployments is to use ConfigMap entries
 - **Default**: `""` (empty)
 - **Example**: `"redis-password"`
 
+### Resource Proxy Configuration
+
+#### Enable Resource Proxy
+- **Command Line Flag**: `--enable-resource-proxy`
+- **Environment Variable**: `ARGOCD_AGENT_ENABLE_RESOURCE_PROXY`
+- **ConfigMap Entry**: *(Not available in ConfigMap)*
```
chetan-rns: Nit: Should the configmap entry be `agent.resource-proxy.enable`?

jannfis: Good catch, fixed.

PR #500, opened 2025-08-07
fix: Better transformation of AppProjects from autonomous agents

### LINKED ISSUE #496: AppProjects synced from autonomous agent are missing transformations
**Describe the bug**

When an AppProject is created on a cluster managed by an autonomous agent, it will be synced to the principal. The name of the AppProject will be transformed to be unique before being created on the control plane cluster.

However, due to the nature the Argo CD API uses the AppProject to decide allowed actions for an Application, the transformations are not sufficient. 

Consider the following Application and AppProject to be created on the managed cluster:

**Application:**
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: guestbook
  namespace: argocd
spec:
  project: test-project
  source:
    repoURL: https://github.com/argoproj/argocd-example-apps
    targetRevision: HEAD
    path: kustomize-guestbook
  destination:
    name: in-cluster
    namespace: guestbook
  syncPolicy:
    syncOptions:
    - CreateNamespace=true
```

**AppProject:**
```yaml
apiVersion: argoproj.io/v1alpha1
kind: AppProject
metadata:
  name: test-project
  namespace: argocd
spec:
  destinations:
  - namespace: 'guestbook'
    name: 'in-cluster'
    server: 'https://kubernetes.default.svc'
  sourceRepos:
  - '*'
```

This would allow the guestbook application to be synced into the namespace `guestbook` on the managed cluster. Now, when both the Application and AppProject are synced to the principal and created on the control plane, the name of the AppProject would be correctly transformed to `<agent-name>-test-project` and the `.spec.project` reference in the Application would be transformed to reflect that, BUT:

* The Application will be created in the agent's namespace (`<agent-name>`)
* The `.spec.destination` of the Application will be transformed to point to the agent, i.e. 
   ```
   destination:
      name: <agent-name>
      namespace: guestbook
   ```

Thus, the Argo CD API server will reject any operation on the Application, because:

* `.spec.sourceNamespaces` in the AppProject do not reflect the reality on the control plane (should include `<agent-name>`
* `.spec.destinations` in the AppProject do not reflect the reality on the control plane (should include `<agent-name>` as destination)

1. Create Application and AppProject as shown above on a managed cluster running an agent in autonomous mode
2. See both, AppProject and Application being synced to the control plane
3. Try viewing the Application through the UI and notice an error from the Argo CD API that the project does not allow the action that was tried

**Expected behavior**

The AppProject should be properly transformed when being synced to the control plane:

* `.spec.sourceNamespaces` should be set to exactly the `<agent-name>`. Any existing sourceNamespaces should be deleted from the final resource.
* `.spec.destinations` should be transformed such that all entries should have their destinations set to `<agent-name>` and keep the namespace fields as-is

**Additional context**

### LINKED ISSUE #497: Autonomous agents are able to overwrite the default project on the principal
**Describe the bug**

If a user updates the AppProject` default` on a managed cluster running an autonomous agent, the changes are propagated to the AppProject `default` on the control plane cluster.

This is due to the condition in the code [here](https://github.com/argoproj-labs/argocd-agent/blob/14a6d0125b940ee42dc510784e5be3c98318a1e4/principal/event.go#L296).

1. Modify the AppProject `default` on a workload cluster running an autonomous agent, e.g. add an entry to `.spec.sourceNamespaces`
2. See that this change is propagated to the AppProject `default` on the control plane

**Expected behavior**
I'd expect the change not being propagated to the AppProject `default` on the control plane cluster. I'd probably expect an AppProject `<agent-name>-default` to exist for each agent on the control plane.

**Additional context**

jannfis:

This PR correctly applies transformations to `.spec.sourceNamespaces` and `.spec.destinations` when an AppProject is synced from an autonomous agent to the principal.

Also, it will not treat the `default` AppProject synced from an autonomous agent in a special way anymore.

Fixes #496
Fixes #497

**Checklist**

### DIFF
--- a/principal/event.go
+++ b/principal/event.go
@@ -24,7 +24,6 @@ import (
 	"github.com/argoproj-labs/argocd-agent/internal/event"
 	"github.com/argoproj-labs/argocd-agent/internal/kube"
 	"github.com/argoproj-labs/argocd-agent/internal/manager"
-	"github.com/argoproj-labs/argocd-agent/internal/manager/appproject"
 	"github.com/argoproj-labs/argocd-agent/internal/metrics"
 	"github.com/argoproj-labs/argocd-agent/internal/namedlock"
 	"github.com/argoproj-labs/argocd-agent/internal/resync"
@@ -150,11 +149,9 @@ func (s *Server) processApplicationEvent(ctx context.Context, agentName string, 
 		}
 
 		// AppProjects from the autonomous agents are prefixed with the agent name
-		if incoming.Spec.Project != appproject.DefaultAppProjectName {
-			incoming.Spec.Project, err = agentPrefixedProjectName(incoming.Spec.Project, agentName)
-			if err != nil {
-				return fmt.Errorf("could not prefix project name: %w", err)
-			}
+		incoming.Spec.Project, err = agentPrefixedProjectName(incoming.Spec.Project, agentName)
+		if err != nil {
+			return fmt.Errorf("could not prefix project name: %w", err)
 		}
 
 		// Set the destination name to the cluster mapping for the agent
--- a/principal/event.go
+++ b/principal/event.go
@@ -293,10 +293,17 @@ func (s *Server) processAppProjectEvent(ctx context.Context, agentName string, e
 
 	// AppProjects coming from different autonomous agents could have the same name,
 	// so we prefix the project name with the agent name
-	if agentMode.IsAutonomous() && incoming.Name != appproject.DefaultAppProjectName {
+	if agentMode.IsAutonomous() {
 		incoming.Name, err = agentPrefixedProjectName(incoming.Name, agentName)
 		if err != nil {
 			return fmt.Errorf("could not prefix project name: %w", err)
+		}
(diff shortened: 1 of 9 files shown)


principal/event.go
```diff
@@ -293,11 +293,18 @@ func (s *Server) processAppProjectEvent(ctx context.Context, agentName string, e
 
 	// AppProjects coming from different autonomous agents could have the same name,
 	// so we prefix the project name with the agent name
-	if agentMode.IsAutonomous() && incoming.Name != appproject.DefaultAppProjectName {
+	if agentMode.IsAutonomous() {
```
chetan-rns, line 293: I think we should also remove the default project distinction while handling the app. The autonomous app on the control plane should use the prefixed default project `<agent-name>-default`

https://github.com/argoproj-labs/argocd-agent/blob/14a6d0125b940ee42dc510784e5be3c98318a1e4/principal/event.go#L153

jannfis: Good catch! Fixed.

PR #504, opened 2025-08-11
chore(tests): Remove old e2e tests and rename e2e2 to e2e

### LINKED ISSUE #224: Merge E2E argocd-agent 'test/e2e' and 'tests/e2e2' folder into a single test/e2e folder
At present, we have both:
-  'test/e2e' containing a small number of mock tests that do not require a cluster
- 'test/e2e2' containing a larger number of tests which DO require a cluster.

As previously discussed, we feel these can be merged together into a 'test/e2e', since they are similar enough to be worth combining.

### Criteria:	
- From our testing, it appears these tests may not currently be working: 'test/e2e/e2e_test.go'. We should get them working again, first.
- Next, move this test into 'test/e2e2', and make sure it continues to work (as well as the existing e2e2 tests)
- Rename 'tests/e2e2' to 'test/e2e', and update Makefiles and shell scripts as needed
- Ensure that everything now works when run via PR checks.

jannfis:

When the new e2e framework was introduced, there existed a couple of archaic e2e tests in `test/e2e`. Thus, the new framework was named `e2e2` and placed in `test/e2e2`. Since the old tests were never really used, it was a bit confusing to have to run the `e2e2` test suite.

This PR removes the old, unused tests and renames the `e2e2` test suite and all references to be `e2e`.

Fixes #224

**Checklist**


test/e2e/e2e_test.go
```diff
@@ -1,474 +0,0 @@
...
-	_, err = fakeAppcServer.ApplicationsClientset.ArgoprojV1alpha1().Applications("client").Update(actx, app, v1.UpdateOptions{})
-	require.NoError(t, err)
-
-	<-sctx.Done()
-	<-actx.Done()
-	err = s.Shutdown()
-	require.NoError(t, err)
-}
-
-func Test_WithHTTP1WebSocket(t *testing.T) {
```
chetan-rns, line 305: Should we keep this test since we support HTTP 1 downgrade? I can port it to the new e2e framework in a different PR.

chetan-rns (approved):
LGTM. Thanks!

test/e2e/e2e_test.go
jannfis: Yep, please port it to the new framework. Good catch!

Issue #418, opened 2025-06-17
Populate cluster cache info


jannfis:
**Is your feature request related to a problem? Please describe.**

The cluster secrets and information used by the agent are missing some information that is displayed by the UI, namely, number of APIs, number of resources, and number of Applications.

![Image](https://github.com/user-attachments/assets/ac31c4c7-54c4-465e-8565-1bbb7267d97d)

**Describe the solution you'd like**

Principal should fetch this information periodically from the agent (number of APIs, number of resources) as well as the local cluster (number of Applications) and populate the cluster cache accordingly.

**Describe alternatives you've considered**

**Additional context**

jannfis:
A good implementation would probably to have a go routine in the agent that regularly fetches data from the local cluster cache (Redis), which is populated by the local application controller. The process would encapsulate this as an event to be sent to the principal. The principal will then populate its local cluster cache with this information, which can then be read by Argo CD.

PR #513, opened 2025-08-14
feat: Populate cluster cache info in principal

jparsai:


agent/agent.go
```diff
@@ -353,6 +353,17 @@ func (a *Agent) Start(ctx context.Context) error {
 		go http.ListenAndServe(healthzAddr, nil)
 	}
 
+	// Start the background process of periodic sync of cluster cache info.
+	// This will send periodic updates of Application, Resource and API counts to principal.
+	if a.mode == types.AgentModeManaged {
+		go func() {
+			for {
+				go a.addClusterCacheInfoUpdateToQueue("periodic_sync")
+				time.Sleep(3 * time.Minute)
```
jgwest: Recommend using a ticker rather than a `time.Sleep`

agent/outbound.go
```diff
@@ -261,3 +273,50 @@ func (a *Agent) addAppProjectDeletionToQueue(appProject *v1alpha1.AppProject) {
...
+	clusterServer := "https://kubernetes.default.svc"
+	var clusterInfo *v1alpha1.ClusterInfo
+	var err error
+
+	// If the reason is application_synced, application_created or application_deleted,
+	// we wait for 20 seconds to ensure that updated info is sent to principal.
+	// It is because the cluster cache is not updated immediately by Argo CD.
+	// Instead it is updated in next clusterInfoUpdater cycle.
+	if reason != "periodic_sync" {
+		time.Sleep(20 * time.Second)
```
jgwest: Sleep statements in event code is not a good idea unless there really really really is no other alternative (and I am skeptical that is the case here). 

If it's a problem of waiting for dependent event, then wait for it with a condition/callback. If this is a problem of unit tests taking too long, then we likely need to use (more) mocking. If it's a problem of E2E tests taking too long, that's probably fine (they are usually expected to necessarily be slow).

agent/outbound.go
```diff
@@ -261,3 +273,50 @@ func (a *Agent) addAppProjectDeletionToQueue(appProject *v1alpha1.AppProject) {
...
+
+	clusterServer := "https://kubernetes.default.svc"
+	var clusterInfo *v1alpha1.ClusterInfo
+	var err error
+
+	// If the reason is application_synced, application_created or application_deleted,
+	// we wait for 20 seconds to ensure that updated info is sent to principal.
+	// It is because the cluster cache is not updated immediately by Argo CD.
+	// Instead it is updated in next clusterInfoUpdater cycle.
+	if reason != "periodic_sync" {
```
jgwest: I recommend converting these strings into string constants in all places where they are used

internal/event/event.go
```diff
@@ -176,6 +178,19 @@ func (evs EventSource) AppProjectEvent(evType EventType, appProject *v1alpha1.Ap
...
+func (evs EventSource) ClusterCacheInfoUpdateEvent(evType EventType, clusterInfo *v1alpha1.ClusterInfo) *cloudevents.Event {
+	reqUUID := uuid.NewString()
+	cev := cloudevents.NewEvent()
+	cev.SetSource(evs.source)
+	cev.SetSpecVersion(cloudEventSpecVersion)
+	cev.SetType(evType.String())
+	cev.SetExtension(eventID, reqUUID)
+	cev.SetExtension(resourceID, reqUUID)
+	cev.SetDataSchema(TargetClusterCacheInfoUpdate.String())
+	_ = cev.SetData(cloudevents.ApplicationJSON, clusterInfo)
```
jgwest, line 196: I recommend converting the object from Argo CD `v1alpha1.ClusterInfo` into our own struct, rather than relying on theirs. If we rely on theirs, unexpected changes in their code will modify our protocol fields.

agent/outbound_test.go
```diff
@@ -371,3 +375,51 @@ func Test_addAppProjectDeletionToQueue(t *testing.T) {
...
+	require.NoError(t, err)
+	require.NotNil(t, miniRedis)
+	defer miniRedis.Close()
+	err = WithRedisHost(miniRedis.Addr())(a)
+	require.NoError(t, err)
+
+	// Create the required Redis secret
+	secret := &corev1.Secret{
+		ObjectMeta: v1.ObjectMeta{Name: common.RedisInitialCredentials},
+		Data:       map[string][]byte{common.RedisInitialCredentialsKey: []byte("password123")},
```
jgwest: You may want to use a randomly generated uuid, rather than `password123`. This is a test, so it's unlikely to be a security issue, but it's very easy to just use a random value.

internal/argocd/cluster/cluster.go
```diff
@@ -39,67 +44,154 @@ func (m *Manager) UpdateClusterConnectionInfo(agentName, status string, t time.T
...
+	// Fetch the cluster info from cache.
+	if err := clusterCache.GetClusterInfo(clusterServer, clusterInfo); err != nil {
+		return nil, err
+	}
+
+	return clusterInfo, nil
+}
+
+// getCacheInstance creates a new cluster cache instance from redis.
+func getCacheInstance(ctx context.Context, kubeclient kubernetes.Interface,
```
jgwest: This function appears to create a new cache instance every time the function is called (which seems like it would be often?). Would it make sense to create a single, shared instance?

internal/argocd/cluster/cluster.go
```diff
@@ -39,67 +44,154 @@ func (m *Manager) UpdateClusterConnectionInfo(agentName, status string, t time.T
...
+		"applicationsCount": clusterInfo.ApplicationsCount,
+		"apisCount":         clusterInfo.CacheInfo.APIsCount,
+		"resourcesCount":    clusterInfo.CacheInfo.ResourcesCount,
+		"cluster":           cluster.Name,
+		"agent":             agentName,
+	}).Infof("Updated cluster cache stats in cluster.")
+}
+
+// setClusterInfo saves the given ClusterInfo in the cache.
+func setClusterInfo(ctx context.Context, kubeclient kubernetes.Interface, namespace, clusterServer, redisAddress, agentName, clusterName string,
```
jgwest: Nit: Many of the fields that are being passed are coming from manager. Maybe we should just add this function to manager, e.g. `func (m *Manager) setClusterInfo(...)`. This would get rid of many of the parameters.

internal/argocd/cluster/cluster.go
```diff
@@ -15,20 +15,25 @@
...
 	appstatecache "github.com/argoproj/argo-cd/v3/util/cache/appstate"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
-// UpdateClusterConnectionInfo updates cluster info with connection state and time in mapped cluster
-func (m *Manager) UpdateClusterConnectionInfo(agentName, status string, t time.Time) {
+// SetAgentConnectionStatus updates cluster info with connection state and time in mapped cluster at principal.
+// This is called when the agent is connected or disconnected with the principal.
+func (m *Manager) SetAgentConnectionStatus(agentName, status string, t time.Time) {
```
jgwest: `status string` -> `status appv1.ConnectionStatus`?

`t time.Time` -> `modifiedAt time.Time`?

internal/argocd/cluster/cluster.go
```diff
@@ -39,67 +44,154 @@ func (m *Manager) UpdateClusterConnectionInfo(agentName, status string, t time.T
...
+			APIsCount:      clusterInfo.CacheInfo.APIsCount,
+			ResourcesCount: clusterInfo.CacheInfo.ResourcesCount,
+		},
+	}
+
+	// Preserve existing cluster connection status
+	if existingClusterInfo != nil {
+		newClusterInfo.ConnectionState = existingClusterInfo.ConnectionState
+		if existingClusterInfo.CacheInfo.LastCacheSyncTime != nil {
+			newClusterInfo.CacheInfo.LastCacheSyncTime = existingClusterInfo.CacheInfo.LastCacheSyncTime
```
jgwest: Can you explain the logic of these lines? From my quick read, it seems like the newer (less stale?) value would be overwritten by the older value (in all cases).

agent/inbound.go
```diff
@@ -374,6 +374,12 @@ func (a *Agent) createApplication(incoming *v1alpha1.Application) (*v1alpha1.App
 		return created, nil
 	}
 
+	if a.mode == types.AgentModeManaged {
+		// When sync status changed to Synced, trigger a cluster cache info update event
+		// so that the principal can be informed of the cache update.
+		go a.addClusterCacheInfoUpdateToQueue("application_created")
```
jgwest: We're probably fine to remove `go` and run it on the same thread (presuming we remove the sleep from `addClusterCacheInfoUpdateToQueue`)

If we DONT remove `go`, then we likely need some other mechanism (lock) to ensure that we don't have multiple go-routine all updates the cluster info at the same time,

agent/inbound.go
```diff
@@ -448,8 +454,11 @@ func (a *Agent) deleteApplication(app *v1alpha1.Application) error {
 		return err
 	}
 
-	if a.mode == types.AgentModeManaged && err == nil {
+	if a.mode == types.AgentModeManaged {
 		appCache.DeleteApplicationSpec(app.UID, logCtx)
+		// When sync status changed to Synced, trigger a cluster cache info update event
+		// so that the principal can be informed of the cache update.
+		go a.addClusterCacheInfoUpdateToQueue("application_deleted")
```
jgwest: Same comment re: `go` removal, above

agent/outbound.go
```diff
@@ -105,6 +110,13 @@ func (a *Agent) addAppUpdateToQueue(old *v1alpha1.Application, new *v1alpha1.App
 		WithField(logfields.SendQueueLen, q.Len()).
 		WithField(logfields.SendQueueName, defaultQueueName).
 		Debugf("Added event of type %s to send queue", eventType)
+
+	// When sync status changed to Synced, trigger a cluster cache info update event
+	// so that the principal can be informed of the cache update.
+	if a.mode == types.AgentModeManaged && old.Status.Sync.Status != new.Status.Sync.Status &&
+		new.Status.Sync.Status == v1alpha1.SyncStatusCodeSynced {
+		go a.addClusterCacheInfoUpdateToQueue("application_synced")
```
jgwest: Same comment re: `go` removal, above

internal/argocd/cluster/manager.go
```diff
@@ -133,7 +133,7 @@ func (m *Manager) Start() error {
 	// To avoid this we need to re-save same info before cache is expired.
 	go func() {
 		for {
-			go m.refreshClusterConnectionInfo()
+			go m.refreshClusterInfo()
```
jgwest: Nit, and optional, up to you: Can likely remove `go` here, since it's already running in a `go` func. 

(Also optional: we could also convert the time.Sleep to a ticker, but since this change was already in the code, it's optional for this PR)

principal/apis/eventstream/eventstream.go
```diff
@@ -132,7 +132,7 @@ func (s *Server) newClientConnection(ctx context.Context, timeout time.Duration)
 	}
 	c.start = time.Now()
 
-	s.clusterMgr.UpdateClusterConnectionInfo(c.agentName, v1alpha1.ConnectionStatusSuccessful, c.start)
+	s.clusterMgr.SetAgentConnectionStatus(c.agentName, v1alpha1.ConnectionStatusSuccessful, c.start)
```
jgwest: Optional, just a thought: Maybe `Subscribe`  (which calls this function) would be a better place to put the call to `SetAgentConnectionStatus`? But I don't have a strong opinion.

principal/event.go
```diff
@@ -348,6 +350,29 @@ func (s *Server) processAppProjectEvent(ctx context.Context, agentName string, e
 	return nil
 }
 
+// processClusterCacheInfoUpdateEvent processes the clusterCacheInfoUpdate event received from the agent.
+// It updates the cluster cache stats in the principal's cluster cache.
+func (s *Server) processClusterCacheInfoUpdateEvent(agentName string, ev *cloudevents.Event) error {
+	clusterInfo := &v1alpha1.ClusterInfo{}
```
jgwest: As above, IMHO our protocol should not hardcode embedded Argo CD objects. We should use our own objects (which copy the fields of Argo CD's)

internal/argocd/cluster/cluster.go
```diff
@@ -39,67 +44,154 @@ func (m *Manager) UpdateClusterConnectionInfo(agentName, status string, t time.T
...
 		}
-		redisClient := redis.NewClient(redisOptions)
+	}
+}
 
-		cache := appstatecache.NewCache(cacheutil.NewCache(
-			cacheutil.NewRedisCache(redisClient, time.Minute, m.redisCompressionType)), time.Minute)
+// SetClusterCacheStats updates cluster cache info with Application, Resource and API counts in principal.
+// This is called when principal receives clusterCacheInfoUpdate event from agent.
+func (m *Manager) SetClusterCacheStats(clusterInfo appv1.ClusterInfo, agentName string) {
```
jgwest: Nit: we should return errors here, rather than just logging them. (If calling functions don't want to return them, that's up to them).

This will also allow the unit tests that call this function to verify no errors occurred.

agent/outbound.go
jparsai: Yes @jgwest I didn't want to use sleep, but issue is that [ClusterInfo](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L2279C1-L2290C2) does not have any "Last updated time" field. [ClusterCacheInfo](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L2301C1-L2308C2) has a `LastCacheSyncTime` field but what I found that it is not updated when `ApplicationCount` , `ResourcesCount`, `APIsCount` are updated instead it is updated once in 12 Hours. This is a comment I found.
 
`"LastCacheSyncTime will continue to be updated every 12 hours (or whatever the configured interval is) as part of Argo CD's regular cluster monitoring process."  
`

I also tried by storing counts before creating/deleting an application and compare it values after, and send an event when value has changed (i.e. old + 1 or old - 1). But it might now work properly when multiple apps are being created and deleted at the same time.
So the simplest option I though is a sleep. As it is a background process and delay of few seconds would not affect user experience much.  Also Argo CD updates cache in regular interval, instead of immediately after create/delete/sync operation.

internal/argocd/cluster/cluster.go
jparsai: In principal cache we are updating  `ClusterInfo.ConnectionState` through `SetAgentConnectionStatus` function whenever Agent is connected or disconnected  and `ClusterInfo.ApplicationsCount` & `ClusterInfo.CacheInfo` fields are updated by `SetClusterCacheStats` function. In this block I am trying to avoid overwriting connection state value which is updated by `SetAgentConnectionStatus` function.

internal/argocd/cluster/cluster.go
jparsai: As discussed over call, here 2 instances are created where calling function is passing Redis Address as argument. To change it, I will have to modify logic and fetch the address of Redis for both. Also there is one more issue I need to address that cache expires after some time so I will have to add logic to refresh cache instance. I am working on it, but you can continue with review of other parts of the code if you wish.

agent/inbound.go
jparsai: `go` is added because there is a sleep for few seconds and we should not hold the execution of calling function for it. even sleep is removed there has to be some kind of retry mechanism which will still take some time to complete and this function will be blocked. So I think it is better to execute that logic in separate  go-routine. If you feel it is still not needed I will change it.

About locks, at Agent side it is only reading from Agent's cache (which is updated by Argo CD) and sending that information to Principal. At principal side I have lock in [SetClusterCacheStats](https://github.com/jparsai/argocd-agent/blob/cluster-cache-info/internal/argocd/cluster/cluster.go#L96) function where lock are used based on clusters. Should I still have a lock at agent side as well?

agent/outbound.go
jgwest: Thanks for the detailed explanation of the problem area! Having looked at Argo CD code, I'd still like to get rid of the sleep statement, and here's my attempt at a solution. As always, feel free to poke holes in my logic as you are better versed in the code your PR is touching.
		
So my thinking is: If Argo CD is updating cluster info every 10 seconds (with a ticker), why can't we do the same thing? 

That is, we change the ticker length in our code to 10 seconds, here:

```go
	// Start the background process of periodic sync of cluster cache info.
	// This will send periodic updates of Application, Resource and API counts to principal.
	if a.mode == types.AgentModeManaged {
		go func() {
			ticker := time.NewTicker(10 * time.Seconds)  // <================================ CHANGED HERE 
			defer ticker.Stop()
			// Send initial cache stats to principal
			a.addClusterCacheInfoUpdateToQueue(clusterCacheUpdateReasonPeriodic)

			for {
				select {
				case <-ticker.C:
					a.addClusterCacheInfoUpdateToQueue(clusterCacheUpdateReasonPeriodic)
				case <-a.context.Done():
					return
				}
			}
		}()
	}
```

And we remove the sleep statement.

And at that point, since it's only 10 seconds between ticks, we probably don't need the other update logic. 
For example, this can be removed:
```
if a.mode == types.AgentModeManaged {
		// When sync status changed to Synced, trigger a cluster cache info update event
		// so that the principal can be informed of the cache update.
		go a.addClusterCacheInfoUpdateToQueue(clusterCacheUpdateReasonAppCreated)
	}
```
(and likewise with removing the update event)

This means that, in the worst case scenario, principal will only be about 10 seconds out of date from agent Argo CD, and 20 seconds out of date from the actual resource. (With the median time being 5 seconds and 10 seconds, respectively).

Worst case scenario is:
- @ 0 seconds> agent cache refreshes
- @ 1 seconds> argo cd cache refreshes
- @ 2 seconds> A change is made to an Argo CD Application
- @ 10 seconds> agent cache refreshes (missing change made @ 2 seconds)
- @ 11 seconds> argo cd refreshes, which picks up the change from @ 2 seconds.
- @ 20 seconds> agent refreshes, which picks up change from @ 2 seconds.

After all, if Argo CD itself doesn't even bother to update these statistics after each application update/change, why should we bother to be EVEN MORE granular with the statistics?

For the E2E tests, we just need to use an `Eventually` statement that waits the max possible time (20 seconds).

WDYT?

jgwest: > "LastCacheSyncTime will continue to be updated every 12 hours (or whatever the configured interval is) as part of Argo CD's regular cluster monitoring process."   

Re: 12 hour timer - from Jann's original request, it seems like he is not especially interested in the contents of this field? https://github.com/argoproj-labs/argocd-agent/issues/418. He only mentioned API count, resources count, and applications count. So we likely don't need to worry about the data from that field being out of date (if this was a concern you had)

internal/argocd/cluster/cluster.go
jgwest: Thanks Jayendra, I definitely think it's worth doing.

> As discussed over call, here 2 instances are created where calling function is passing Redis Address as argument. To change it, I will have to modify logic and fetch the address of Redis for both. 

Makes sense, hopefully this isn't too much work.

> Also there is one more issue I need to address that cache expires after some time so I will have to add logic to refresh cache instance. I am working on it, but you can continue with review of other parts of the code if you wish.

My guess is that the other change I suggested above likely fixes the expiration issue: if we are refreshing (writing to) the cluster cache state every 10 seconds, we likely don't need to worry about the value expiring after a minute or 10 minutes.

agent/inbound.go
jgwest: I think my other suggestion above re: only periodically syncing every 10 seconds (and not worrying about application_created/updated events) resolves this issue as well.

agent/outbound.go
jparsai: Yes, reducing ticket to 10 sec will fix all issues, but now we will be sending very frequent events to principal. Is it fine?

jgwest: It would be once every 10 seconds per agent, which seems perfectly reasonable to me.

jparsai: Updated it, now managed agent will send events to principal in every 10 sec.

internal/argocd/cluster/cluster.go
jparsai: Now a shared cache instance is created for both agent and principal. The instance will expire after 10 minutes and on next function call it will be recreated and it will replace old instance.

We still need to have an expiration time. Yes, Redis cache will be updated in every 10 seconds now, but this is connection object to the Redis cache. It is recommended to refresh it frequently so it will up to date. Max expiration time is 24 hours, but it is suggested to keep up to few minutes.

agent/outbound.go
jannfis: Not something for this PR, but we can make this configurable in the future and always adapt the default if we think it is too frequent.

agent/outbound.go
```diff
@@ -261,3 +268,66 @@ func (a *Agent) addAppProjectDeletionToQueue(appProject *v1alpha1.AppProject) {
...
+	// If one event has been sent after application count changed to 0, then do nothing.
+	if eventSentAfterZeroApps {
+		return
+	}
+
+	// If the cluster has no applications, we set the eventSentAfterZeroApps flag to true.
+	// and send one event to principal to update the cluster cache info.
+	if clusterInfo.ApplicationsCount == 0 {
+		eventSentAfterZeroApps = true
+	}
```
jgwest: Is the only goal of this code block to prevent sending a message when application count is 0? If so, I think it should be removed as IMHO this not a case worth optimizing for (messages are cheap, and it is extremely rare for app count to be 0).

internal/cache/sharedcache.go
```diff
@@ -0,0 +1,145 @@
+// Copyright 2025 The argocd-agent Authors
+//
```
jgwest: As a general rule for me, if you're adding a mutable `var` at package scope, it's going to make me go :thinking: when I see it. `var sharedCache = &SharedCache{}` here (and the corresponding logic) is an example of that. AFAICT there are no other mutable package-level fields in the codebase.

Rather than creating a singleton, why not just maintain a reference to cache in...

Agent:
`type Agent struct {` in `agent/agent.go`
- See `redisProxyMsgHandler *redisProxyMsgHandler` for an example of redis cache object I use in agent case


Principal:
`type Server struct {` in `server/server.go` (where we store most other principal state)
or perhaps
`type Manager struct {` in `internal/argocd/cluster/manager.go`

internal/argocd/cluster/cluster.go
jgwest: Correct me if I'm wrong, but the expiration time is ONLY used to set TTL of keys in the redis cache. It has nothing to do with the connection lifecycle. From my working with go-redis cache API, there is no need to consistently refresh the connection.

For example:
```go
	clusterCache := appstatecache.NewCache(cacheutil.NewCache(
		cacheutil.NewRedisCache(redisClient, cacheInstanceExpirationDuration, redisCompressionType),
	), cacheInstanceExpirationDuration)
```

Is 2 calls with an expiration parameter specified:
- cacheutil.NewRedisCache(redisClient, cacheInstanceExpirationDuration, redisCompressionType)
    - expiration value is used for `Set`s on Redis, which [controls TTL of key](https://redis.io/docs/latest/commands/ttl/).
- appstatecache.NewCache(, cacheInstanceExpirationDuration)
    - expiration value set here is likely not used by our code: it is only used to control TTL on `Set` calls to redis for managed resource / app resources tree, which our agent code doesn't touch.

jgwest (changes requested):

internal/argocd/cluster/cluster.go
jparsai: I guess, I misunderstood the go-redis implementation, now I changed it as you suggested and not refreshing the cache instance. PTAL.

internal/cache/sharedcache.go
```diff
@@ -0,0 +1,111 @@
...
+const (
+	SourceAgent     = "agent"
+	SourcePrincipal = "principal"
+)
+
+type SharedCache struct {
+	mu                     sync.RWMutex
+	agentCacheInstance     *appstatecache.Cache
+	principalCacheInstance *appstatecache.Cache
+}
```
jgwest: The contents of this cache struct/file is IMHO over-engineered: AFAICT the code between the agent and principal case is exactly the same, so there is no point in having separate agent/principal functions here.

Likewise the
```go
const (
	SourceAgent     = "agent"
	SourcePrincipal = "principal"
)
```
values are only used to call the separate agent/principal functions, and thus are not being used.


So why not just modify SharedCache struct to look like this:
```go
type SharedCache struct {
	mu sync.RWMutex // This may not even be needed, if cacheInstance is always set by `NewCache`
	cacheInstance          *appstatecache.Cache
}
```

And remove the `GetCacheInstance` function entirely. 

You can replace it with something like
```go

// This is pseudocode, not real code

func NewCache(ctx context.Context, kubeclient kubernetes.Interface,
	namespace, redisAddress string, redisCompressionType cacheutil.RedisCompressionType) (*SharedCache) {

			cache, err := sc.createCacheInstance(ctx, kubeclient, namespace, redisAddress, redisCompressionType)
	if err != nil {
		return nil, err
	}

	return &SharedCache{
		cacheInstance: cache
	}
	
}
```
to initialize the cache.

And a simple get function to retrieve the appstatecache.Cache for it, 
```go
// getAgentCacheInstance returns the agent cache instance
func (sc *SharedCache) GetCache() (*appstatecache.Cache, error) {

	sc.mu.Lock()
	defer sc.mu.Unlock()

	return sc.cacheInstance
}
```

You probably don't even need the mutex.

internal/cache/sharedcache.go
```diff
@@ -0,0 +1,111 @@
...
+
+// getAgentCacheInstance returns the agent cache instance
+func (sc *SharedCache) getAgentCacheInstance(ctx context.Context, kubeclient kubernetes.Interface,
```
jgwest: Thinking out loud, do we even need this `SharedCache` file/package at all? Why not just put `cacheInstance *appstatecache.Cache` in all the places where we are currently adding a reference to SharedCache?

e.g. remove SharedCache from `type Agent struct {` and in `type Manager struct {`, and just add appstatecache.Cache

jgwest (changes requested):

jgwest (changes requested):
LGTM, thanks @jparsai!

jgwest (approved):

PR #515, opened 2025-08-14
test: wait for resource action to be available

chetan-rns:

* The e2e test should wait for the resource action to be available. This could prevent a race condition where the test executes the resource action before Argo CD discovers the action.

* Switch to resource action API version V2

Fixes #?

**Checklist**


chetan-rns:
This PR should fix the failing resource action e2e test. Switching to `RunResourceActionV2` fixed the issue. I think it was a breaking change. Although the previous version was [deprecated](https://github.com/argoproj/argo-cd/blob/d737f8fe43adfa005f94a2ad2478d1d29f60676c/server/application/application.go#L2495).

jannfis (approved):
LGTM, thanks!

PR #524, opened 2025-08-19
fix: Drop owner references from autonomous agent events

### LINKED ISSUE #523: Applications with ownerReferences might get deleted when synced from an autonomous agent
**Describe the bug**

As we found out in this Slack thread, if ApplicationSet generates Applications on a workload cluster managed by an autonomous agent, they will be successfully synced to the principal but are getting deleted a couple of moments later without any obvious cause.

It turns out that GKE (and potentially other Kubernetes engines) delete resources that have `ownerReferences` pointing to an owner that does not exist. In this case, Applications generated by ApplicationSet on the managed cluster will carry `ownerReferences` to the ApplicationSet, but the ApplicationSet itself does not exist on the principal's system. Thus, these Applications get garbage collected on the control plane cluster.

Probably, any ownerReferences should be stripped from the Applications when they are created or updated on the control plane cluster.

1. (...)
2. (...)
3. (...)

**Expected behavior**

Application not getting deleted on the control plane cluster.

**Additional context**

ncr38:
Drops Owner reference from Application and App project metadata while sending update/create events. 

GKE generic-garbage-collector was deleting applications created on control plane from autonomous agent events if Appset was missing.

Audit log from GKE console : 

> Kubernetes Apiservice Requests delete test-autonomous-agent:test-app system:serviceaccount:kube-system:generic-garbage-collector {@type: type.googleapis.com/google.cloud.audit.AuditLog, authenticationInfo: {…}, authorizationInfo: […], methodName: io.argoproj.v1alpha1.applications.delete, requestMetadata: {…}, resourceName: argoproj.io/v1alpha1/namespaces/test-autonomous-agent/applications/test-app , serviceName: k8s.io

Fixes #523 

**Checklist**

### DIFF
--- a/agent/inbound_test.go
+++ b/agent/inbound_test.go
@@ -586,7 +586,7 @@ func Test_UpdateApplication(t *testing.T) {
 		napp, err := a.updateApplication(app)
 		require.NoError(t, err)
 		require.NotNil(t, napp)
-		require.Empty(t, app.OwnerReferences, "OwnerReferences should not be applied on managed app")
+		require.Empty(t, napp.OwnerReferences, "OwnerReferences should not be applied on managed app")
 	})
 
 	t.Run("Update application using update in managed mode", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


jannfis (approved):
LGTM! Thank you @ncr38

PR #525, opened 2025-08-20
feat: use informer cache for Get method in Kubernetes backend

### LINKED ISSUE #251: Kubernetes backend should use informer cache for retrieving resources

Right now, the [Kubernetes backend](https://github.com/argoproj-labs/argocd-agent/tree/main/internal/backend/kubernetes) for Applications and AppProjects talk directly to the Kubernetes API server when retrieving (e.g. list or get) resources. However, since we use shared informers for both type of resources, we should make use of its caches instead. This will reduce the number of requests being made to the Kubernetes API significantly.

The `Get` and `List` methods of the Kubernetes backends for `Application` and `AppProject` type resources should use the informer's cache instead of reaching out to the API.

juanxiu:
This PR enhances the KubernetesBackend by modifying the Get method to use the informer cache instead of querying the Kubernetes API server directly when retrieving Application resources. It leverages the generic Informer's Lister() method to efficiently access cached resources, thereby reducing load on the API server and improving performance. Corresponding unit tests have been added to verify correct informer cache usage. This change creates a foundation for more efficient resource management via the informer caching mechanism.

Fixes #251 

- The Get() method now primarily uses the informer cache by default.
- Differentiation between update-related Get() and regular Get() calls is done via a ForUpdateContextKey context flag.
- Added informer synchronization checks and ensured nil safety.

**Checklist**


internal/backend/kubernetes/application/kubernetes.go
```diff
@@ -82,7 +89,15 @@ func (be *KubernetesBackend) Create(ctx context.Context, app *v1alpha1.Applicati
...
-	return be.appClient.ArgoprojV1alpha1().Applications(namespace).Get(ctx, name, v1.GetOptions{})
+	obj, err := be.appLister.ByNamespace(namespace).Get(name)
+	if err != nil {
+		return nil, err
+	}
+	app, ok := obj.(*v1alpha1.Application)
+	if !ok {
+		return nil, fmt.Errorf("object is not an Application: %T", obj)
+	}
+	return app, nil
```
jannfis: Items returned from the cache need to be treated read-only. I believe that's why unrelated unit tests are failing and panicing.

There are two options imho:

1. We clearly document that objects returned by this function are to be treated read-only, because they are retrieved directly from the cache and caller needs to make a copy if they want to modify it, or
2. We return a copy of the object retrieved from the cache, to take the burden from the caller.

The first option puts more responsibility to the caller, but is resource efficient. The second option would ensure that the caller can treat the objects lightly, but for the cost of increased memory consumption.

I have not yet made up my mind with regards to which solution I'd prefer. Which one do you think makes more sense?

jannfis (commented):
Thanks @juanxiu for this PR.

I have a comment requiring some more discussion, PTAL.

internal/backend/kubernetes/application/kubernetes.go
juanxiu: In option 1, when objects are retrieved through multiple informers, there is a risk that callers may forget to use DeepCopy(). Such oversights can make it difficult to trace and resolve bugs. On the other hand, callers can use resources efficiently and have the flexibility to decide when to create copies.

Option 2 always returns a copy of the object from the function, so callers cannot control when the copy is made. However, callers can still customize behavior by registering event handlers with the informer. Returning a copy from the function enforces consistent object usage and prevents inconsistent handling of copying across different callers.

Personally, I chose option 2 because I believe minimizing the potential for bugs and ensuring consistent and stable behavior throughout the codebase is important.

For these reasons, I have modified the code to `return app.DeepCopy(), nil.`
Additionally, I plan to update the List methods in the Kubernetes backend for Application and appproject type resources to use the informer cache as well. I would appreciate it if we could merge this after those changes are completed.

jannfis: It seems there are other problems with this approach, at least regarding the tests. Some of them are still failing.

juanxiu: There is no issue with loading the informer cache itself. However, in the current unit tests, a different problem arises. To load the cache, informer.Run must be executed, which requires calling the Start method of either the manager or the server. Until now, the code has been written without assuming that Start would be called in the test code. As a result, timing issues related to goroutine execution occur during test runs. How can we resolve this situation?

principal/event_test.go
```diff
@@ -128,6 +128,7 @@ func Test_CreateEvents(t *testing.T) {
 		wq.On("Get").Return(&ev, false)
 		wq.On("Done", &ev)
 		s, err := NewServer(context.Background(), fac, "argocd", WithGeneratedTokenSigningKey(), WithAutoNamespaceCreate(true, "", nil))
+		s.Start(context.Background(), make(chan error))
```
juanxiu, line 131: During testing in this way, a Start call is required to load the informer.

juanxiu: @jannfis Can you take a look at this issue and let me know if you have any ideas to resolve it?

juanxiu:
Hi, @jannfis ! Kindly requesting your review.

The Get() method has been improved by applying the informer cache for better performance. Regular Get() calls now use the informer cache, while Get() calls within the update retry loop are conditionally routed to directly query the API server via a context flag. All tests have passed successfully.

jannfis:
Thank you @juanxiu ! During review, I noticed that this PR actually contains three features instead of one: Ability to disable Redis proxy, ability to specify informer sync timeout, and the use of the informer cache.

It would be great if you could untangle these, and submit as separate PRs. That way, we can keep better track of things and independently revert changes, should it be necessary. Thank you.

juanxiu:
Hi, @jannfis 
Thank you for the review. I would like to explain the relationship between the three features.
The core feature is the addition of the informer cache. The goal was to improve performance by reading from the local cache instead of the API server in KubernetesBackend.Get().
The issues arose during testing of this feature:

1. Redis proxy disable option:
To initialize Principal, NewServer() must be called, which by default starts the Redis proxy. Since there was no Redis server in the test environment, the principal itself could not be created. (All existing tests in principal/server_test.go also use WithRedisProxyDisabled())

2. Informer sync timeout setting:
The informer cache must be fully synced (HasSynced()) to be used. If tests start the server and immediately validate, the cache is not synced yet, causing test instability. So, a timeout setting was added to allow adjusting this duration in tests for stability.

Ultimately, the informer cache is the main feature, and the other two are necessary to properly test it.

If separated, there would be a situation where intermediate states without test coverage or unstable tests would be merged.
Is there a better approach?

Thanks

internal/backend/kubernetes/application/kubernetes.go
```diff
@@ -82,6 +89,23 @@ func (be *KubernetesBackend) Create(ctx context.Context, app *v1alpha1.Applicati
 }
 
 func (be *KubernetesBackend) Get(ctx context.Context, name string, namespace string) (*v1alpha1.Application, error) {
+	forUpdate, _ := ctx.Value(backend.ForUpdateContextKey).(bool)
+
+	if !forUpdate && be.appLister != nil && be.appInformer != nil && be.appInformer.HasSynced() {
```
jannfis, line 94: Can you please explain this construct?

internal/informer/informer.go
```diff
@@ -109,6 +113,12 @@ func NewInformer[T runtime.Object](ctx context.Context, opts ...InformerOption[T
 	i := &Informer[T]{}
 	var r T
 	i.resType = reflect.TypeOf(r)
+
+	i.groupResource = schema.GroupResource{
+		Group:    "argoproj.io",
+		Resource: "applications",
+	}
```
jannfis: With this hard-coded, every informer would only ever have a lister for Applications. I think the groupResource must be set from the outside, potentially at initialization, depending on the type concrete type of the informer.

internal/manager/application/application.go
```diff
@@ -555,8 +555,14 @@ func (m *ApplicationManager) Delete(ctx context.Context, namespace string, incom
...
 	var updated *v1alpha1.Application
+
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	ctxForUpdate := context.WithValue(ctx, backend.ForUpdateContextKey, true)
+
 	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		existing, ierr := m.applicationBackend.Get(ctx, incoming.Name, incoming.Namespace)
+		existing, ierr := m.applicationBackend.Get(ctxForUpdate, incoming.Name, incoming.Namespace)
```
juanxiu, line 565: @jannfis

The reason why `forUpdate` is necessary lies here. Kubernetes implements **Optimistic Concurrency Control** through `resourceVersion`. When an Update request is made, the API server compares the `resourceVersion` in the request with the one currently stored, and if they differ, it returns a Conflict.

The issue arises when using the informer cache during updates. Although the informer is synchronized with the API server, there is a slight lag, meaning that the cache may return a stale `resourceVersion`. As a result, the Update request fails, and even when using `retry.RetryOnConflict()` to retry, it keeps reading the same stale `resourceVersion` from the cache, causing repeated failures.

In contrast, when `forUpdate=true`, the object is fetched directly from the API server, guaranteeing the latest `resourceVersion`. This ensures that during retries, the latest version is retrieved, allowing the Update to succeed.

Therefore, we use the informer cache for regular reads to optimize performance, and direct API reads during updates to ensure correctness.

internal/backend/kubernetes/application/kubernetes.go
juanxiu: Could you please take a look at my explanation below?

internal/informer/informer.go
juanxiu: Thanks for the feedback. You’re right — I removed the hard-coded groupResource.
Instead, I added `WithGroupResource[T](group, resource string)` as a new InformerOption, which allows each informer to be configured with the correct groupResource at initialization time based on its concrete type.

PR #526, opened 2025-08-20
feat: sync private repository secrets to managed agents

### LINKED ISSUE #474: Support for private repositories

argocd-agent synchronizes Applications and AppProjects between the principal and agents. Both principal and agents use their own, dedicated repository server to access Git. The user can create repository credentials on the principal (e.g. via the UI), and the principal might actually be able to access that private repository. However, once an Application is synced to an agent, the agent does not have access to the credentials, and that will lead to a sync failure.

The principal should sync repository secrets created on the control plane to managed agents. The initial thought is to only sync project-scoped repository credentials, and sync them to those agents where the AppProject is also synced to.

For autonomous agents, the source of truth is the agent. I cannot see a requirement for those repository secrets to be synced to the control plane, because the control plane (principal) will not need access to those repositories. The desired manifests will be stored in the agent's redis cache, which the principal has access to through the redis-proxy component.

This feature needs a bit of validation and thought, potentially a design document.

**Additional context**

chetan-rns:

- Add informers to both agent and principal to reconcile repository secrets
- Propagate the repository secrets to the correct target agent based on AppProjects.
- Cache the mapping between repos to agents in case the AppProject is deleted.
- Add unit tests and e2e tests

Fixes #474 

1. Create an AppProject on the principal cluster targeting one of the managed agents.
2. Create a repository secret referring to the newly created AppProject
3. Repository must be synced to the managed agent.

**Checklist**


jannfis:
I'm just testing this PR locally, and most of the things work really great.

I noticed one thing:

- I have scoped a repository to a project. The project itself allows the `agent-managed` cluster. 
- The repository gets propagated to the `agent-managed` agent, and is created on the target cluster. 
- I stop the principal temporarily and make a change to the project, removing the `agent-managed` from the list of allowed destinations. 
- I start the principal again. The agent reconnects, but the repository credentials are never removed from the `agent-managed` cluster

jannfis (approved):
LGTM.

Awesome PR, @chetan-rns 🙇

PR #529, opened 2025-08-21
fix: sync default appproject updates to managed agents

### LINKED ISSUE #509: Default AppProject updates are not synced to managed agents
**Describe the bug**
When the default AppProject is updated on the control plane, the changes are not propagated to the default AppProject on the managed agent cluster.

1. Update the default AppProject on the control plane
2. Default AppProject on the managed agent cluster remains unchanged.

**Expected behavior**
The principal should propagate any changes made to the default AppProject to the managed agents. The agent should update the default AppProject locally.

**Additional context**
The default AppProject on the control plane may not have a name field in its destination. But the principal tries to map the AppProject to the agents based on the destination name. This needs further verification.

chetan-rns:
The principal should propagate default AppProject updates to the managed agents. This PR handles several issues:

- The user could have already created the default AppProject before running the agent. The agent should remove the user-created default AppProject and create the default AppProject from the principal. It should add the source UID annotation.
- The default AppProject may only have the destination server field and not the name. The principal should consider both the name and the server to match the agent. Since the principal doesn't know the workload cluster's server, we only check if the server is a wildcard.
- Add support for deny patterns
- Add e2e tests to verify the behavior

Fixes #509 

**Checklist**


principal/callbacks.go
```diff
@@ -303,20 +305,46 @@ func (s *Server) mapAppProjectToAgents(appProject v1alpha1.AppProject) map[strin
...
+	return agents
+}
+
+// doesAgentMatchWithProject checks if the agent name matches the given AppProject.
+// We match the agent to an AppProject if:
+// 1. The destination name is not empty and matches the agent name OR
+// 2. The destination server is a wildcard AND
+// 3. The agent name is not denied by any of the destination names
+// Ref: https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/app_project_types.go#L477
+func doesAgentMatchWithProject(agentName string, appProject v1alpha1.AppProject) bool {
```
chetan-rns: Looking at the [argo-cd code](https://github.com/argoproj/argo-cd/blob/88fccc91c67eb9145691ae93e3b995d949ecd41c/pkg/apis/application/v1alpha1/app_project_types.go#L477), it checks for both the destination name and the server, and either one of them should match. But the principal isn't aware of the workload cluster's server URL, and the cluster secret points to the resource proxy. So, if the user doesn't want to use the name, the only option is to use a wildcard for the server. i.e, we match the agent if the server is * even if the name doesn't match. WDYT?

I guess we could pass the workload cluster URL to the principal in the client context. But so far, the principal doesn't know about the workload cluster, and the agent is the only point of contact.

jannfis: > and either one of them should match

Either one should match or be empty. Setting both fields as destination at the same time is imho not supported by Argo CD, i.e. you need to set either name or server, but not both.

> the cluster secret points to the resource proxy

The cluster secret points to a specific URL including the agent-name in the query param, so a user could match it. However, I think it would be fair to force the user to use the name of the agent instead of the server URL for now. The URLs will all look similar, except for the suffix, anyway.

chetan-rns: Thanks! I've updated the logic to compare the agent name from the server URL query param. The server URL check is required for the default AppProject, which doesn't have the destination name set initially. The function will only compare the server if the name is not provided.

jannfis:
@chetan-rns The merge of https://github.com/argoproj-labs/argocd-agent/pull/526 introduced some conflicts that need to be resolved.

jannfis (approved):
LGTM

PR #530, opened 2025-08-21
fix: Remove delete patch for repo server

ncr38:
Removes delete patch commands for repo-server resources.
We'll need repo-server for argocd UI/CLI commands to work. For example : Sync


chetan-rns (approved):
Thanks! LGTM.

PR #537, opened 2025-08-27
test(e2e): Fix flaky test for ClusterCacheInfo validation

jparsai:
This PR is to modify the validation of ClusterCacheInfo.

The test fails sometimes because of delay in ClusterCacheInfo update by Argo CD. 
Here is the scenario when test fails:

1) At 1st second `ClusterCacheInfo` is refreshed by Argo CD periodic update (every 10 seconds).
2) 5th second, app is created by a test, but cache is yet to be refreshed in next iteration.

3) 10th second, cache is refreshed and now it has app count 1.
4) 15th second, last test completed and app is deleted, but again cache is yet to be refreshed on 20 second.
5) 18th second, `Test_ClusterCacheInfo` started and it fetched current number of apps from agent cache (appCountBefore) which is still set to 1 from last iteration (on 10th second).

6) 20th second, cache refreshed and application count is now set to 0 (based on last test).
7) Now a new application is created by `Test_ClusterCacheInfo` and it is assuming that no of apps in cache would be 1+1=2 (appCountBefore + new app), but actually it is 0+1, hence the test case fails here expecting no of app to be 2.

The easiest fix is to wait for Argo CD to refresh cache before starting the `Test_ClusterCacheInfo` test to allow Argo CD to refresh the cache.

### DIFF
--- a/test/e2e/clusterinfo_test.go
+++ b/test/e2e/clusterinfo_test.go
@@ -42,31 +42,31 @@ func TestClusterTestSuite(t *testing.T) {
 }
 
 func (suite *ClusterInfoTestSuite) SetupTest() {
-	if !fixture.IsProcessRunning(PrincipalName) {
+	if !fixture.IsProcessRunning(fixture.PrincipalName) {
 		// Start the principal if it is not running and wait for it to be ready
-		suite.Require().NoError(fixture.StartProcess(PrincipalName))
-		fixture.CheckReadiness(suite.T(), PrincipalName)
+		suite.Require().NoError(fixture.StartProcess(fixture.PrincipalName))
+		fixture.CheckReadiness(suite.T(), fixture.PrincipalName)
 	} else {
 		// If principal is already running, verify that it is ready
-		fixture.CheckReadiness(suite.T(), PrincipalName)
+		fixture.CheckReadiness(suite.T(), fixture.PrincipalName)
 	}
 
-	if !fixture.IsProcessRunning(AgentManagedName) {
+	if !fixture.IsProcessRunning(fixture.AgentManagedName) {
 		// Start the agent if it is not running and wait for it to be ready
-		suite.Require().NoError(fixture.StartProcess(AgentManagedName))
-		fixture.CheckReadiness(suite.T(), AgentManagedName)
+		suite.Require().NoError(fixture.StartProcess(fixture.AgentManagedName))
+		fixture.CheckReadiness(suite.T(), fixture.AgentManagedName)
 	} else {
 		// If agent is already running, verify that it is ready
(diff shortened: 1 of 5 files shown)


test/e2e/clusterinfo_test.go
```diff
@@ -155,6 +155,13 @@ func (suite *ClusterInfoTestSuite) Test_ClusterInfo_Autonomous() {
 // This test suite validates the cluster cache info reporting for managed agent.
 // It checks that the cluster cache info is correctly propagated from agent to principal.
 func (suite *ClusterInfoTestSuite) Test_ClusterCacheInfo() {
+	// Sleep for few seconds to allow Argo CD to refresh the cache in agent cluster,
+	// because Argo CD doesn't refresh the cache immediately after application is created/deleted.
+	// Instead it is done periodically in background, now even though application is deleted after last test,
+	// but the application count may still have previous value, which will provide wrong value for "appCountBefore"
+	// variable and our validation for application count in this test will fail.
+	time.Sleep(20 * time.Second)
```
jannfis: I don't think that a 20 second sleep is a proper fix for this issue. It prolongs the test's runtime by a fixed 20. If we'd do that with 20 tests, we'd add a fix 400 seconds (that's almost 7 minutes!) to the running time of the end-to-end tests that just do nothing else than sleep.

We should find a proper solution, e.g. figuring out whether the cache has been updated and only proceed then.

jparsai: Yes, it increases the runtime, but the sleep is required only for the tests where cluster info cache is tested. As of now it is tested only in one test `Test_ClusterCacheInfo`. I tried to find  better solution in first PR itself (https://github.com/argoproj-labs/argocd-agent/pull/513#discussion_r2282625908) and discussed with @jgwest for some better approach, but we are limited by Argo CD cluster info updater logic. By default it updates cache in every 10 seconds (which is configurable) and after that only agent gets updated cache and sends it to principal. I am happy to try a better approach but Argo CD cluster info logic is the root cause. 

Here is the Argo CD logic which is responsible for updating cluster info in a cluster https://github.com/argoproj/argo-cd/blob/master/controller/clusterinfoupdater.go#L59

jannfis: > 15th second, last test completed and app is deleted, but again cache is yet to be refreshed on 20 second.

Can't we just reset the cluster cache for all agents and the principal manually when the test concludes?

jparsai: Yes this works, now I am cleaning cache as part of cleanup process. PTAL.

jannfis: Nice! :+1:

PR #541, opened 2025-09-03
test(e2e): e2e tests should be runnable individually

jparsai:
This PR is to update logic for E2E execution. Existing logic needs few env variables to be set which was done via `run-e2e.sh` script but this stops users from running individual tests directly from IDE. Now all configurations needed are set in tests itself.


jannfis (approved):

jannfis:
LGTM, thanks!

Issue #403, opened 2025-06-05
Certificate errors while starting to test principal


deB4SH:
**Describe the bug**
First off: thanks for building something to solve a big scaling issue with many small scale clusters on one big hub instance and I know this is pretty much alpha, but I wanted to test things out.

When installing the current argocd-agent principal from main based on the provided manifests under `install/kubernetes/principal` the principal seems to miss a cluster role to access secrets. While installing everything I followed https://github.com/argoproj-labs/argocd-agent/blob/main/docs/hack/quickstart.md closely. 

```
{"level":"info","msg":"Setting loglevel to info","time":"2025-06-05T11:43:42Z"}
time="2025-06-05T11:43:42Z" level=info msg="Loading gRPC TLS configuration from files cert=/app/config/tls/tls.crt and key=/app/config/tls/tls.key"
[FATAL]: Error reading TLS config for resource proxy: error getting proxy certificate: could not read TLS secret /argocd-agent-resource-proxy-tls: secrets "argocd-agent-resource-proxy-tls" is forbidden: User "system:serviceaccount:argocd:argocd-agent-principal" cannot get resource "secrets" in API group "" at the cluster scope
stream closed EOF for argocd/argocd-agent-principal-846b7bcd4d-hxfnz (argocd-agent-principal)
```

Adding the following extra role resolves the issue and the principal runs into a second issue.

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: argocd-agent-principal-secrets-read
rules:
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - get
  - list
  - watch
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: argocd-agent-principal-secrets-read
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: argocd-agent-principal-secrets-read
subjects:
- kind: ServiceAccount
  name: argocd-agent-principal
  namespace: argocd
```

The rising new issue is the following listing.

```
{"level":"info","msg":"Setting loglevel to info","time":"2025-06-05T11:47:04Z"}
time="2025-06-05T11:47:04Z" level=info msg="Loading gRPC TLS configuration from files cert=/app/config/tls/tls.crt and key=/app/config/tls/tls.key"
[...]

PR #542, opened 2025-09-03
fix: principal errors due to forbidden - in bindings subject updated the principal ServiceAccount namespace to argocd (from default)

jblsk:
**What does this PR do / why we need it**: This PR fixes the CrashLoopBackOff and forbidden errors when argocd-agent-principal is deployed due to ineffective Role and ClusterRole ment to be bound to `argocd-agent-principal` ServiceAccount.

**Which issue(s) this PR fixes**: #403 

Fixes #?

The `argocd-agent-principal` ServiceAccount which is the subject of the ClusterRoleBinding and RoleBinding should be deployed to `argocd` namespace (according to documentation [here](https://argocd-agent.readthedocs.io/latest/getting-started/kubernetes/#31-deploy-principal-component)) and then non-existent in the `default` namespace. This PR fixes the false namespace value in `argocd-agent-principal` ClusterRoleBinding and RoleBinding subject's namespace key

**Checklist**

* [x] No documentation update is required.


jannfis:
I think that one is supposed to run `kustomize edit set namespace argocd` before running `kustomize build`.

If we're considering `argocd` to be the default target namespace for the installation, it might make sense to incorporate the changes from this PR. Maybe a proper fix is to have `namespace: argocd` in the `kustomization.yaml`. But I agree, using the `default` namespace as default doesn't make much sense.

mikeshng:
My vote is option 1 as well.

Setting the namespace to argocd will help us simplify installs. IE. we can remove the first troubleshooting in this doc regarding namespace: https://argocd-agent.readthedocs.io/latest/getting-started/openshift/#troubleshooting

PR #549, opened 2025-09-07
feat: add e2e test for interaction with ArgoCD API

### LINKED ISSUE #413: Add e2e tests for interaction through Argo CD API

Our end-to-end tests create apps to be synced in a declarative way, i.e. create the `Application` resources directly on the clusters. However, there are subtle differences between creating the apps in a declarative way and creating them through Argo CD.

For example, we broke something (https://github.com/argoproj-labs/argocd-agent/issues/412) without the end-to-end test noticing that.

The end-to-end tests should have one or more test cases creating and deleting applications through the Argo CD API. There are already some fixtures in place to interact through the Argo CD API, they might have to be enhanced. We do not need a full sync-to-agent cycle. For now, we should test whether we're able to:

1) Create an Application
2) List and get more details about the Application
3) Delete the Application

juanxiu:
This PR introduces a new end-to-end test case to validate the lifecycle of an Application resource by interacting directly with the Argo CD API.

Our existing end-to-end tests create Application resources declaratively, by applying manifests directly to the cluster. As discovered in issue #412, this approach has a gap: it doesn't fully simulate user actions performed through the Argo CD API. This can lead to subtle bugs in the API interaction path going undetected.

To improve test coverage and prevent similar issues, this change adds a test that specifically covers the API-driven creation, retrieval, and deletion of applications, ensuring our integration with the Argo CD API is robust.

Fixes #413 

This pull request includes the following changes:

New E2E Test for Application Lifecycle ([test/e2e/application_test.go](code-assist-path:/Users/yeonsookim/%EA%B0%9C%EB%B0%9C/argocd-agent/test/e2e/application_test.go))

A new test case, TestApplicationLifecycle_ViaAPI, has been added.
This test performs a full lifecycle of an application via the REST API, including:
Creating a new application using CreateApplication.
Verifying its creation by fetching it with GetApplication and ensuring it appears in the output of ListApplications.
Deleting the application using DeleteApplication.
Confirming successful deletion by attempting to get the application again and asserting that a "not found" error is returned.
Enhanced Argo CD API Client Fixture ([test/e2e/fixture/argoclient.go](code-assist-path:/Users/yeonsookim/%EA%B0%9C%EB%B0%9C/argocd-agent/test/e2e/fixture/argoclient.go))

The existing test fixture for the Argo CD REST client has been enhanced to support the new test case.
The following methods have been added to the ArgoRestClient:
CreateApplication
GetApplication
ListApplications
DeleteApplication
The client now correctly encapsulates the logic for authentication and request signing, making it a reliable tool for our E2E test suite.
This new test provides greater confidence that creating and managing applications via the Argo CD API works as expected in an environment where the agent is active.

**Checklist**


test/e2e/fixture/argoclient.go
```diff
@@ -331,3 +331,123 @@ func GetArgoCDServerEndpoint(k8sClient KubeClient) (string, error) {
...
+
+	return &appList, nil
+}
+
+// DeleteApplication deletes an application by its name
+func (c *ArgoRestClient) DeleteApplication(name string) error {
+	reqURL := c.url()
+	reqURL.Path = fmt.Sprintf("/api/v1/applications/%s", name)
+
+	req, err := http.NewRequest(http.MethodDelete, reqURL.String(), nil)
```
jannfis, line 436: I think you need to set proper content-type for this request

juanxiu: I added code to set the Content-Type header to application/json in a new commit

.github/workflows/ci.yaml
```diff
@@ -183,9 +183,20 @@ jobs:
         run: |
           curl -L -o vcluster "https://github.com/loft-sh/vcluster/releases/latest/download/vcluster-linux-amd64" && sudo install -c -m 0755 vcluster /usr/local/bin && rm -f vcluster
           vcluster --version
+      - name: Wait for MetalLB components to be ready
+        run: |
+          microk8s.kubectl wait --for=condition=ready deployment -l app=metallb -n metallb-system --timeout=120s
+      
+      - name: Install ArgoCD CRDs
+        run: kubectl apply -f https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/crds.yaml
```
jannfis: This is actually part of setting up the test environment later.

juanxiu: Thanks for the feedback! You were right, that part was related to the test environment setup.

Based on your comment, I've reverted the changes to ci.yaml. 

I've modified the test setup logic instead of changing the CI configuration. The test suite now checks for the existence of the required contexts at startup. If a context is not found (as is common in a local environment), the test is now skipped with an informative message instead of failing.

This change has no impact on the CI environment, where the contexts are present and the tests will run as they always have. It simply makes the test suite more robust and convenient for local development.

test/e2e/application_test.go
```diff
@@ -0,0 +1,100 @@
...
+	requires.NoError(err)
+
+	found := false
+	for _, item := range appList.Items {
+		if item.Name == appName {
+			found = true
+			break
+		}
+	}
+	asserts.True(found, "application %s not found in list", appName)
```
Jack-R-lantern, line 85: ~~The message is not correct~~

test/e2e/application_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// 4. Delete Application
+	err = s.argoClient.DeleteApplication(appName)
+	requires.NoError(err)
+
+	// 5. Verify Deletion
+	_, err = s.argoClient.GetApplication(appName)
+	asserts.Error(err)
+	asserts.Contains(err.Error(), "not found")
```
Jack-R-lantern: I think this step is unnecessary.

Jack-R-lantern (changes requested):

juanxiu:
@jannfis The test case failure was caused by a nil issue. Since the test turned out to be unnecessary, I have removed it. Could you please approve the actions?

PR #551, opened 2025-09-09
feat: Selectively prevent resources from being synced

### LINKED ISSUE #490: Configuration aynchronization should be selectable
No

We should be able to select what configuration resources (application, appProject) are being synced between control plane and workload clusters.

The configmap for the agent could have an option similar to the principal.allowed-namespaces were we can write list  of resources to be synced.

**Additional context**
I am very exicted for this project, but it is still at it's early stages and some users might be reluctant to integrate it into their environment. Using the project only for UI observability can encourage teams to start using the project early. I understand the autonomous option is supposed to cover this need, but appProjects and applications are still synced into the control plane, which makes it difficult to use [OCM pull model](https://github.com/open-cluster-management-io/argocd-pull-integration)  along with the agent to later transition to using only the agent.

Looking forward to this project being GA, very cool!

jannfis:

These changes let users put a label on resources (Applications, AppProjects and repository configurations) and the principal and agent will ignore these resources, i.e. not perform any sync actions whatsoever with them.

The label is `argocd-agent.argoproj-labs.io/skip-sync` and it needs to be set to `"true"` to take effect.

This lets you, temporarily or permanently, disable updates on these resources.

Fixes #490

Unit and e2e tests have been generally generated by Cursor.

**Checklist**

### DIFF
--- a/principal/server.go
+++ b/principal/server.go
@@ -32,6 +32,7 @@ import (
 	kubeappproject "github.com/argoproj-labs/argocd-agent/internal/backend/kubernetes/appproject"
 	kubenamespace "github.com/argoproj-labs/argocd-agent/internal/backend/kubernetes/namespace"
 	kuberepository "github.com/argoproj-labs/argocd-agent/internal/backend/kubernetes/repository"
+	"github.com/argoproj-labs/argocd-agent/internal/config"
 	"github.com/argoproj-labs/argocd-agent/internal/event"
 	"github.com/argoproj-labs/argocd-agent/internal/filter"
 	"github.com/argoproj-labs/argocd-agent/internal/informer"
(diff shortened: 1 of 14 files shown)


principal/server.go
```diff
@@ -804,9 +805,17 @@ func (s *Server) loadTLSConfig() (*tls.Config, error) {
...
 func (s *Server) defaultAppFilterChain() *filter.Chain[*v1alpha1.Application] {
 	c := filter.NewFilterChain[*v1alpha1.Application]()
+	// Admit based on namespace of the application
 	c.AppendAdmitFilter(func(res *v1alpha1.Application) bool {
 		return glob.MatchStringInList(s.options.namespaces, res.Namespace, glob.REGEXP)
 	})
+	// Ignore applications that have the skip sync label
+	c.AppendAdmitFilter(func(res *v1alpha1.Application) bool {
+		if v, ok := res.Labels[config.SkipSyncLabel]; ok && v == "true" {
+			return false
```
chetan-rns, line 815: Nit: Do we need this filter? The label selector skips all apps with the skip sync label.

jannfis: You're right, we don't really need it. I added it as a safeguard only.

PR #567, opened 2025-09-16
fix: revert updates to autonomous apps on the control plane

### LINKED ISSUE #465: Autonomous agent applications not recreated after deletion on control plane
**Describe the bug**

When deleting an application on the control plane, where the source of truth is the autonomous agent, the Application will not be recreated on the control plane until the agent reconnects.

This removes visibility on the control plane, and could lead to some confusion and false alerts.

1. Create an Application on an autonomous agent system
2. Observe it being created on the control plane
3. Delete the Application using the UI (or other means) on the control plane
4. Observe the Application gone on the control plane and not being recreated (bad)
5. Observe the Application still existing on the agent (good)
6. Restart autonomous agent
7. Observe Application being recreated/resynced on the control plane

**Expected behavior**
After deletion of an Application on the control plane that is managed by an autonomous agent, it should be recreated right way (in the sense of self-heal) by the principal.

**Additional context**

chetan-rns:
Workload cluster is the source of truth for apps created via an autonomous agent. 

* Any updates to the autonomous apps on the control plane must be reverted immediately.
* If the autonomous app on the control is deleted, it must be recreated immediately.
* Add unit and e2e tests

Fixes #465 

1. Create an autonomous app and let it sync to the control plane
2. Update/Delete the app on the control plane
3. The changes must be reverted immediately.

**Checklist**


internal/manager/application/application.go
```diff
@@ -644,6 +644,39 @@ func (m *ApplicationManager) RevertManagedAppChanges(ctx context.Context, app *v
...
+
+	sourceUID, exists := app.Annotations[manager.SourceUIDAnnotation]
+	if !exists {
+		return false
+	}
+
+	if cachedAppSpec, ok := appCache.GetApplicationSpec(ty.UID(sourceUID), logCtx); ok {
+		logCtx.Debugf("Application: %s is available in agent cache", app.Name)
+
+		if diff := reflect.DeepEqual(cachedAppSpec, app.Spec); !diff {
```
jannfis: Nit: I think the variable name `diff` is a bit misleading, because it is set to true when there is no difference and to false when there is one.

internal/manager/application/application.go
```diff
@@ -644,6 +644,39 @@ func (m *ApplicationManager) RevertManagedAppChanges(ctx context.Context, app *v
...
+			app.Spec = cachedAppSpec
+			logCtx.Infof("Reverting modifications to the application: %s", app.Name)
+			if _, err := m.UpdateAutonomousApp(ctx, app.Namespace, app); err != nil {
+				logCtx.Errorf("Unable to revert modifications done in application: %s. Error: %v", app.Name, err)
+				return false
+			}
+			return true
+		}
+	} else {
+		logCtx.Debugf("Application: %s is not available in agent cache", app.Name)
```
jannfis: Is this kind of an invalid situation? Are there scenarios where this can happen legitimately?

jannfis (commented):
I had a few minor nits, please see below.

Also, I have taken this to a local test, and I found the following situation:

1. Create Application on autonomous agent. It will be synced to the control plane.
2. Sync that application
3. Delete that application in the UI with foreground deletion
4. Observe the "Deleting" operation icon spinning - a resources finalizer has been put on the Application on the control plane.
5. The finalizer will be removed after some time (with the next status update from the agent , I think)

However, I think the finalizer should be removed right away from the Application on the control plane.

Without the finalizer, it works like a charme.

internal/manager/application/application.go
chetan-rns: I've changed the variable name to `isEqual`.

internal/manager/application/application.go
chetan-rns: Yeah. Cache miss shouldn't happen here. I've changed the log level to Error.

chetan-rns:
> However, I think the finalizer should be removed right away from the Application on the control plane.

@jannfis I've removed the finalizer on the control plane, and the foreground/background deletion no longer gets stuck. I've included it in the same PR. Do let me know if I need to raise a separate one.

jannfis (approved):
LGTM, thanks @chetan-rns

PR #569, opened 2025-09-16
feat(logstream): Log streaming for argocd agent

Mangaal:
This PR introduces a unidirectional(agent → principal) log streaming service and wires it into the resource-proxy path so the Principal can serve Kubernetes pod logs to the Argo CD UI. The Agent handles both static logs (follow=false) and live streaming (follow=true) with resume support.

What’s included:

- New LogStreaming service (gRPC) — Agent opens a client-streaming RPC and pushes log chunks keyed by request_uuid; Principal writes directly to the HTTP response stream and returns a final status when the stream ends.
- Principal resource-proxy integration — /…/pods/{name}/log requests are recognised, the HTTP writer is registered, and a log event is enqueued to the Agent.
- Agent log workers — static and live log handlers; time-window flush or 64KiB chunk flush; live streaming has resume (SinceTime) on transient errors.

- Key feature:
- Principal LogStream gRPC server & HTTP bridge.
- Agent log streaming implementation (static + live + resume).
- Principal resource proxy: log-subresource branch & handoff to LogStream.


chetan-rns:
@Mangaal I see this error intermittently on the UI. Works fine after requesting the logs again. I guess we are not handling `EOF` somewhere?
```
Get "https://rathole-container-internal:9090/api/v1/namespaces/
guestbook/pods/kustomize-guestbook-ui-7689b675bc-cbv8h/log?container=guestbook-ui&follow=true&
tailLines=1000&timestamps=true": EOF
```

agent/log.go
```diff
@@ -0,0 +1,586 @@
+package agent
```
chetan-rns, line 15: Some of the files are missing the `Copyright` headers. Please include them.

agent/log.go
```diff
@@ -0,0 +1,586 @@
...
+	eof := false
+	for !eof {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+
+		line, err := br.ReadString('\n')
+		switch err {
```
chetan-rns: Doesn't the argocd server process the logs before streaming to the UI? Just want to make sure the agent does not repeat the same work since it's only supposed to send back the responses from the k8s API. Can we directly send the chunks back and let the argocd server handle the raw logs? Then the agent doesn't have to go through the logs line by line.

principal/resource.go
```diff
@@ -97,24 +100,98 @@ func (s *Server) processResourceRequest(w http.ResponseWriter, r *http.Request,
...
+	} else {
+		logCtx.Infof("Proxying request for resources of type %s in namespace %s", gvr.String(), requestedNamespace)
+	}
+
+	if requestedSubresource == "log" {
+		logCtx.WithFields(logrus.Fields{
+			"namespace": requestedNamespace,
+			"pod":       requestedName,
+			"params":    reqParams,
+			"agent":     agentName,
```
chetan-rns, line 167: Can we add these fields above so that they are also available for the previous log statements?

principal/apis/logstream/logstream.proto
```diff
@@ -0,0 +1,32 @@
...
+package logstreamapi;
+
+option go_package = "github.com/argoproj-labs/argocd-agent/pkg/api/grpc/logstreamapi";
+
+// LogStreamData represents a line (or chunk) of log data sent from the agent
+message LogStreamData {
+  // Unique identifier matching the original request/event UUID
+  string request_uuid = 1;
+  // Log data content. Recommended to be a single line including trailing newline
+  string data = 2;
```
chetan-rns: Should we use bytes as a data type instead of a string? Prevents the overhead of conversion at both the agent and the principal.

agent/log.go
Mangaal: In my first approach, I have tried sending raw log chunks back to the server since the argocd server handles them, but on the ui, they were displayed unprocessed. I will revisit.

Mangaal:
@chetan-rns, Thanks for reviewing my PR. I’ve updated it and addressed your suggestions. Please take a look when you get a chance.

agent/log.go
```diff
@@ -0,0 +1,579 @@
...
+		default:
+		}
+
+		n, err := br.Read(readBuf)
+
+		// First, handle any bytes we got (even if err == io.EOF)
+		if n > 0 {
+			remain := readBuf[:n]
+			for len(remain) > 0 {
+				space := chunkMax - len(buf)
```
chetan-rns: Any specific reason why we append the chunks? `Read()` will fill the buffer (up to chunkMax or less) and return the number of bytes read. Can we just send all the bytes read? I'm thinking if that would simplify the logic.

agent/log.go
```diff
@@ -0,0 +1,579 @@
...
+				RequestUuid: logReq.UUID,
+				Error:       "log stream read failed",
+			})
+			return err
+		}
+
+		// Timer-based flush (fires only after a read iteration)
+		select {
+		case <-ticker.C:
+			if err := flush("timer"); err != nil {
```
chetan-rns: I think `Read()` blocks until the data is available. So we wait after sending all the read data until the next bytes are available. Do we still need a timer-based flush in that case?

agent/log.go
```diff
@@ -0,0 +1,579 @@
...
+						return lastTimestamp, err
+					}
+					space = chunkMax
+				}
+				toCopy := len(remain)
+				if toCopy > space {
+					toCopy = space
+				}
+				buf = append(buf, remain[:toCopy]...)
+				remain = remain[toCopy:]
```
chetan-rns: Same as above. What if we send all the read data directly? For example:

```
for {
    n,err = rc.Read(readBuf)
    
    if n > 0 {
      stream.Send(readBuf[:n])
    }

    // handle EOF
}
```

agent/log.go
```diff
@@ -0,0 +1,579 @@
...
+	b.MaxInterval = 5 * time.Second
+	b.MaxElapsedTime = 30 * time.Second
+	bo := backoff.WithContext(b, ctx)
+
+	for {
+		// Build resume request
+		resumeReq := *logReq
+		if lastTimestamp != nil {
+			t := lastTimestamp.Add(-100 * time.Millisecond)
+			resumeReq.SinceTime = t.Format(time.RFC3339)
```
chetan-rns, line 252: `SinceSeconds` and `SinceTime` are sent from the Argo CD server as log stream options. The agent will use these options while creating the k8s log stream. Why does the agent need to modify it?

agent/log.go
```diff
@@ -0,0 +1,579 @@
+// Copyright 2024 The argocd-agent Authors
```
chetan-rns:
```suggestion
// Copyright 2025 The argocd-agent Authors
```

agent/log.go
Mangaal: This is used for retry logic. Suppose there is an authentication issue/token expiration, or any temporary error for long-lived streaming. In that case, I need to resume from the last timestamp that was sent to the ArgoCD server, and the tiny overlap ensures no log lines are lost.

agent/log.go
Mangaal: I was thinking that if we send for every read, it will lead to more gRPC calls when reads are small, which will result in more CPU usage. I was batching the logs to 64 KiB or flushing every 50 ms to keep the UI moving and consistent.

agent/log.go
Mangaal: SinceSeconds and SinceTime are originally sent by the Argo CD server as part of the log request options.
However, the agent updates SinceTime locally for retry and resume logic, not for the initial request flow.
When the agent is doing long-lived log streaming (with Follow=true), temporary interruptions due to token expiration or re-authentication or transient errors.
To handle this gracefully, the agent tracks the last seen log timestamp and the tiny overlap ensures no log lines are lost. Since we send an early ACK, the event streamer doesn’t retry the request or handle these errors itself, so the agent needs to manage resuming locally.

agent/log.go
Mangaal: For long‑lived live log streaming, we send an early ACK. The event streamer expects a response(including error)/ACK within its timeout. Otherwise, it may resend the request. Since a live stream can exceed that timeout, we ACK immediately and run the stream in the background. If the stream drops for any intermediate error, we resume from the last seen timestamp (minus 100ms).

chetan-rns: I'm okay with keeping the batching logic since it doesn't affect the functionality. It does reduce the number of gRPC messages. Flushing every 50 ms may minimize the effect on latency. My only concern was the additional complexity it adds. gRPC streams are designed for handling continuous streams of data. Serialization/deserialization is also efficient (thanks to protobuf). I guess gRPC may already batch multiple messages internally when sending out HTTP/2 frames. We could benchmark to find the difference. But it is out of scope for this PR.

Mangaal: Thanks for pointing that out! I agree , the batching adds a bit of extra complexity. I’ll revert it to the simpler version for now, and we can revisit  later if we notice any performance issues.

chetan-rns: We should include a comment if we keep the batching logic. I see that it's used for both static and live streaming.

agent/agent.go
```diff
@@ -105,6 +105,8 @@ type Agent struct {
 	cacheRefreshInterval time.Duration
 	clusterCache         *appstatecache.Cache
 
+	inflightMu   sync.Mutex
+	inflightLogs map[string]context.CancelFunc
```
jgwest: Two things:
- Can you add a description of what this field is used for
- AFAICT the value (cancel func) is never called. Is this intentional?

agent/log.go
```diff
@@ -0,0 +1,423 @@
...
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				_ = stream.Send(&logstreamapi.LogStreamData{RequestUuid: logReq.UUID, Eof: true})
+				return lastTimestamp, nil
+			}
+			_ = stream.Send(&logstreamapi.LogStreamData{RequestUuid: logReq.UUID, Error: err.Error()})
+			return lastTimestamp, err
+		}
+	}
+}
```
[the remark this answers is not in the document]
jgwest: This comment (https://github.com/argoproj-labs/argocd-agent/pull/569#discussion_r2510784493)  appears to be valid, and should be addressed.

agent/log.go
```diff
@@ -0,0 +1,423 @@
...
+
+// streamLogs streams logs until the context is done, returning the last seen timestamp.
+// It flushes raw data, using chunk size (64KB) or time-based flushing.
+// Timestamps are extracted from raw lines for retry capability.
+// If an error occurs during send, it attempts to close the stream and propagate
+// the appropriate error back to the caller for retry or termination.
+func (a *Agent) streamLogs(ctx context.Context, stream logstreamapi.LogStreamService_StreamLogsClient, rc io.ReadCloser, logReq *event.ContainerLogRequest, logCtx *logrus.Entry) (*time.Time, error) {
```
jgwest, line 340: `logCtx` is unused

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/alicebob/miniredis/v2 v2.35.0
 	github.com/argoproj/argo-cd/v3 v3.1.9
 	github.com/argoproj/gitops-engine v0.7.1-0.20250905160054-e48120133eec
+	github.com/cenkalti/backoff/v4 v4.3.0
```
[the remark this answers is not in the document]
Mangaal: For this PR, I feel we can stick with v4  because MaxElapsedTime has been removed in v5. We can safely use v4 as it has no known security vulnerabilities.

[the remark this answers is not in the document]
Mangaal: I got this error when trying to use

b.MaxElapsedTime undefined (type *"github.com/cenkalti/backoff/v5".ExponentialBackOff has no field or method MaxElapsedTime)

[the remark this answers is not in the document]
Mangaal: Sticking with v4.3.0 for now

agent/agent.go
Mangaal: I have added this to the block duplicate log request. I have added a comment.

jgwest:
@Mangaal PR LGTM, if you rebase I will merge. I tried rebasing but there were too many conflicts (and I didn't have knowledge on what the changes were). You may want to squash your commit first before rebasing to make the rebase process easier.

PR #571, opened 2025-09-17
fix: Set namespace for AppProjects and repositories

### LINKED ISSUE #570: AppProject namespace not updated by Agent
**Describe the bug**

When the Agent receives the AppProject it doesn't update the namespace to reflect the namespace it is configured to use, as a result if you have the Principal and Agent in different namespaces it fails.

1. Deploy Principal in one namespace (openshift-gitops-agent) on the control plane cluster
2. Deploy the Agent in a different namespace (openshift-gitops) on its cluster
3. Check the Agent logs to see this message.

```
time="2025-09-16T23:29:33Z" level=error msg="Unable to process incoming event" clientAddr="52.118.50.50:443" direction=Recv error="failed to validate source UID of appProject: appprojects.argoproj.io \"remote-cluster\" is forbidden: User \"system:serviceaccount:openshift-gitops:argocd-agent-agent\" cannot get resource \"appprojects\" in API group \"argoproj.io\" in the namespace \"openshift-gitops-agent\"" event_id=remote-cluster_14f172fe-adde-4535-8254-ab8e22171f74_184519 module=StreamEvent resource_id=remote-cluster_14f172fe-adde-4535-8254-ab8e22171f74 type=io.argoproj.argocd-agent.event.spec-update
```

**Expected behavior**
The Agent updates the AppProject namespace and creates it successfully in its own namespace.

**Additional context**
It should be checked if any other resources will have the same issue.

This is needed to support an Agent on the control-plane cluster since by definition in this situation the Agent will live in a different namespace then the Principal.

jannfis:

The AppProject and Repository Credentials management of both agent and principal  do not set the namespace properly. Thus, syncing those resources works only when principal and agent  are running in the same namespace (i.e. "argocd" or similar).

This PR sets the namespace explicitly to the principal's or agent's namespace before trying to create/update/delete said resources.

Fixes #570 

**Checklist**

### DIFF
--- a/agent/inbound.go
+++ b/agent/inbound.go
@@ -105,6 +105,9 @@ func (a *Agent) processIncomingApplication(ev *event.Event) error {
 		return err
 	}
 
+	// Applications must exist in the same namespace as the agent
+	incomingApp.SetNamespace(a.namespace)
+
 	var exists, sourceUIDMatch bool
 
 	if a.mode == types.AgentModeManaged {
@@ -196,6 +199,9 @@ func (a *Agent) processIncomingAppProject(ev *event.Event) error {
 		return err
 	}
 
+	// AppProjects must exist in the same namespace as the agent
+	incomingAppProject.SetNamespace(a.namespace)
+
 	exists, sourceUIDMatch, err := a.projectManager.CompareSourceUID(a.context, incomingAppProject)
 	if err != nil {
 		return fmt.Errorf("failed to validate source UID of appProject: %w", err)
@@ -276,6 +282,9 @@ func (a *Agent) processIncomingRepository(ev *event.Event) error {
 	if err != nil {
 		return err
 	}
+
+	// Repository secrets must exist in the same namespace as the agent
+	incomingRepo.SetNamespace(a.namespace)
 
 	var exists, sourceUIDMatch bool
 
@@ -422,7 +431,9 @@ func (a *Agent) processIncomingResourceResyncEvent(ev *event.Event) error {
 // createApplication creates an Application upon an event in the agent's work
 // queue.
 func (a *Agent) createApplication(incoming *v1alpha1.Application) (*v1alpha1.Application, error) {
+	// Applications must exist in the same namespace as the agent
 	incoming.SetNamespace(a.namespace)
+
 	logCtx := log().WithFields(logrus.Fields{
 		"method": "CreateApplication",
 		"app":    incoming.QualifiedName(),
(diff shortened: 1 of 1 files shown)


agent/inbound.go
chetan-rns, line 1: Thanks! Should we also update the namespace before comparing the source UID? Otherwise it may try to fetch the resource from the agent's namespace.

https://github.com/argoproj-labs/argocd-agent/blob/a2a7a63d8a93a8ccf20c6d34ab1d8621945baf3f/agent/inbound.go#L199

https://github.com/argoproj-labs/argocd-agent/blob/a2a7a63d8a93a8ccf20c6d34ab1d8621945baf3f/agent/inbound.go#L284

chetan-rns (commented):
Thanks! Added a small question. Everything else looks good to me.

agent/inbound.go
gnunn1: I'm testing this PR out and I wonder if I'm hitting this issue, I have the principal in `openshift-gitops-principal` and the agent in `openshift-gitops` on the same cluster. The AppProject doesn't get copied over to the Agents namespace and I see this message in the agent log:

```
time="2025-10-01T22:10:24Z" level=error msg="Unable to process incoming event" clientAddr="172.31.75.84:443" direction=Recv error="failed to validate source UID of appProject: appprojects.argoproj.io \"local-cluster\" is forbidden: User \"system:serviceaccount:openshift-gitops:argocd-agent-agent\" cannot get resource \"appprojects\" in API group \"argoproj.io\" in the namespace \"openshift-gitops-principal\"" event_id=local-cluster_7a8ac8fd-5fe9-4dc6-823d-f37c783b77ea_143200 module=StreamEvent resource_id=local-cluster_7a8ac8fd-5fe9-4dc6-823d-f37c783b77ea type=io.argoproj.argocd-agent.event.spec-update
```

I did build and deployed the agent from the branch with the fix.

chetan-rns: > system:serviceaccount:openshift-gitops:argocd-agent-agent\" cannot get resource \"appprojects\" in API group \"argoproj.io\" in the namespace \"openshift-gitops-principal\

I think it's the same issue. The agent is trying to fetch the AppProject from the principal namespace instead of it's own agent namespace.

jannfis: Thanks. Now thinking about it, I believe the `SetNamespace` call makes most sense early in the `processIncoming...` functions, instead of in the `create`/`delete`/`update` functions.

I'll make the change and test it thoroughly.

jannfis:
@gnunn1 Would you be able to re-test this PR to see if the recent changes fixed the issues you were seeing?

gnunn1:
I can but not until next week if that's OK.

gnunn1:
I tested this on the same cluster with the principal in `argocd` and the agent in `argocd-agent` and it worked fine. However in the logs for the agent I see this warning, not sure if there is a concern there? 

```
time="2025-10-14T22:02:54Z" level=info msg="Creating a new AppProject on behalf of an incoming event" appProject=managed-cluster method=CreateAppProject module=Agent
time="2025-10-14T22:02:54Z" level=warning msg="Could not manage app managed-cluster: resource managed-cluster is already managed" component=AppProjectManager
```

PR #575, opened 2025-09-21
fix: follow what Argo CD does and set the rolebinding namespace to argocd

mikeshng:
Follow what Argo CD does and set the rolebinding namespace to argocd for both principal and agent

See https://github.com/argoproj-labs/argocd-agent/pull/542#issuecomment-3253788558 for more details.

This default ns has been pretty annoying when setting up the argocd-agent env for some time now. Following the setup/installation steps is arguably the hardest part of this project, so we should make it as simple as possible for beginners.

Fixes https://github.com/argoproj-labs/argocd-agent/issues/403

**Checklist**

### DIFF
--- a/install/kubernetes/agent/agent-clusterrolebinding.yaml
+++ b/install/kubernetes/agent/agent-clusterrolebinding.yaml
@@ -13,4 +13,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: argocd-agent-agent
-  namespace: default
+  namespace: argocd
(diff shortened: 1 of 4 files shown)


docs/getting-started/openshift/index.md
```diff
@@ -206,20 +206,7 @@ data:
 #### Troubleshooting
 ___
 
-1. If pod fails to come up with error
```
jannfis: I do not think this should go. This will happen at any time when the namespace in the (cluster-)rolebinding does not match the installation namespace.

mikeshng: You are right. Reverted.

jannfis (approved):
LGTM

PR #576, opened 2025-09-22
docs: fix a few issues with the k8s getting started guide

mikeshng:
docs: fix a few issues with the k8s getting started guide: https://argocd-agent.readthedocs.io/latest/getting-started/kubernetes/

- Add the missing propagate CA step
- Add the missing propagate AppProject step
- Fix the principal app example by populating the destination server with the correct url
- Add another troubleshooting item regarding missing server.secretkey 

- When I get to the verify agent secrets step from the guide, I got
```
$ kubectl get secret argocd-agent-ca -n argocd --context kind-cluster1
Error from server (NotFound): secrets "argocd-agent-ca" not found
```
That's because the guide doesn't have the CA propagation step.

- If I don't propagate (or manually create) the AppProject resource on workload cluster then I get the following in Application:
```
status:
  conditions:
  - lastTransitionTime: "2025-09-22T17:04:09Z"
    message: Application referencing project default which does not exist
    type: InvalidSpecError
```

- If I don't have the proper destination.server url then the Application doesn't propagated to workload cluster.

- After the application is propagated to the workload cluster, there is an error in the status
```
status:
  conditions:
  - lastTransitionTime: "2025-09-22T16:20:56Z"
    message: 'error getting cluster by name "in-cluster": server.secretkey is missing'
    type: InvalidSpecError
```

**Checklist**

### DIFF
--- a/docs/getting-started/kubernetes/index.md
+++ b/docs/getting-started/kubernetes/index.md
@@ -273,15 +273,17 @@ argocd-agentctl pki issue agent my-first-agent \
   --upsert
 ```
 
-### 5.3 Create Agent Namespace on Principal
-
-For managed agents, create a namespace on the principal where the agent's Applications will be created and managed:
+### 5.3 Propagate Certificate Authority to Agent
 
 ```bash
-kubectl create namespace my-first-agent --context <control-plane-context>
+argocd-agentctl pki propagate \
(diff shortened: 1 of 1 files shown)


jannfis (approved):
Still LGTM, thanks a lot!

PR #580, opened 2025-09-24
feat: reject agents that use same name as ArgoCD namespace

### LINKED ISSUE #510: Principal should reject agents named like the argo-cd installation namespace

If an agent is named like (and mapped to) the Argo CD installation namespace on the control plane cluster, all kind of unexpected errors occur because the principal expects the application's namespace to be different for many of the functionality, such as the redis proxy, to work correctly.

The principal should reject any connection from an agent that matches the Argo CD installation namespace on the control plane cluster with a clear message that this agent name is not allowed.

**Additional context**

Rizwana777:
Rejects agents hat use same name as ArgoCD installation namespace

Fixes #510 
JIRA - https://issues.redhat.com/browse/GITOPS-7692

### DIFF
--- a/principal/auth.go
+++ b/principal/auth.go
@@ -106,6 +106,12 @@ func (s *Server) authenticate(ctx context.Context) (context.Context, error) {
 		return unauthenticated()
 	}
 
+	// Reject agents that use the same name as the Argo CD installation namespace
+	if agentInfo.ClientID == s.namespace {
+		logCtx.Warnf("Agent name '%s' is not allowed as it matches the Argo CD installation namespace. Please use a different agent name.", agentInfo.ClientID)
+		return unauthenticated()
+	}
+
 	// If we require client certificates, we enforce any potential rules for
 	// the certificate here, instead of at time the connection is made.
 	if s.options.requireClientCerts {
--- a/principal/auth_test.go
+++ b/principal/auth_test.go
@@ -1,0 +1,735 @@
+// Copyright 2025 The argocd-agent Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


principal/auth.go
```diff
@@ -106,6 +106,12 @@ func (s *Server) authenticate(ctx context.Context) (context.Context, error) {
 		return unauthenticated()
 	}
 
+	// Reject agents that use the same name as the Argo CD installation namespace
+	if agentInfo.ClientID == s.namespace {
+		logCtx.Errorf("Agent name '%s' is not allowed as it matches the Argo CD installation namespace. Please use a different agent name.", agentInfo.ClientID)
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Agent name '%s' is not allowed as it matches the Argo CD installation namespace. Please use a different agent name.", agentInfo.ClientID))
```
jannfis: Please just `return unauthenticated()` here. We do not want to propagate too much information to the client here.

It is ok to log the full error and cause, but the client should not be informed about it.

principal/auth.go
```diff
@@ -106,6 +106,12 @@ func (s *Server) authenticate(ctx context.Context) (context.Context, error) {
 		return unauthenticated()
 	}
 
+	// Reject agents that use the same name as the Argo CD installation namespace
+	if agentInfo.ClientID == s.namespace {
+		logCtx.Errorf("Agent name '%s' is not allowed as it matches the Argo CD installation namespace. Please use a different agent name.", agentInfo.ClientID)
```
jannfis: Please use warning instead of error level here, since it's technically not an error on the principal side.

Rizwana777:
> Thanks for this PR, @Rizwana777 !
> 
> I have two comments, PTAL.

I have addressed the comments, PTAL, thank you

Rizwana777:
> Thanks, @Rizwana777. Now we do need some unit tests for this, too.

I have added unit tests for auth.go file as there was no existing tests available for that file and covered existing scenario from this PR as well and  I have taken help of AI to write unit tests, I have  made few changes accordingly, please let me know if there are any mistakes, I will fix it, thank you

jannfis (approved):
LGTM, great job.

Thanks for disclosing use of AI, I appreciate that!

PR #582, opened 2025-09-25
docs: fix principal namespaces config in getting started guide

llavaud:

Update Getting Started Guide to update principal configuration, we need to add the agent namespace in the `principal.allowed-namespaces` parameter.


jannfis (approved):
LGTM, thanks @llavaud

PR #587, opened 2025-09-29
fix: Remove dependency on argocd-redis secret

jparsai:
This PR is to remove dependency of `argocd-redis` secret in `NewClusterCacheInstance` function, because of this users of GitOps operator are forced to create `argocd-redis` secret manually as operator uses `argocd-redis-initial-password` instead. Now we have a new env variable in principal to provide redis password and principal would use this to get redis password instead of looking for `argocd-redis` secret

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -304,8 +304,7 @@ func NewAgent(ctx context.Context, client *kube.KubernetesClient, namespace stri
 		connMap: map[string]connectionEntry{},
 	}
 
-	clusterCache, err := cluster.NewClusterCacheInstance(ctx, client.Clientset,
-		a.namespace, a.redisProxyMsgHandler.redisAddress, cacheutil.RedisCompressionGZip)
+	clusterCache, err := cluster.NewClusterCacheInstance(a.redisProxyMsgHandler.redisAddress, a.redisProxyMsgHandler.redisPassword, cacheutil.RedisCompressionGZip)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create cluster cache instance: %v", err)
 	}
--- a/agent/outbound_test.go
+++ b/agent/outbound_test.go
@@ -396,7 +396,7 @@ func Test_addClusterCacheInfoUpdateToQueue(t *testing.T) {
 	a.emitter = event.NewEventSource("principal")
 
 	// First populate the cache with dummy data
-	clusterMgr, err := cluster.NewManager(a.context, a.namespace, miniRedis.Addr(), cacheutil.RedisCompressionGZip, a.kubeClient.Clientset)
+	clusterMgr, err := cluster.NewManager(a.context, a.namespace, miniRedis.Addr(), "", cacheutil.RedisCompressionGZip, a.kubeClient.Clientset)
 	require.NoError(t, err)
 	err = clusterMgr.MapCluster("test-agent", &v1alpha1.Cluster{
 		Name:   "test-cluster",
--- a/cmd/argocd-agent/principal.go
+++ b/cmd/argocd-agent/principal.go
@@ -83,6 +83,7 @@ func NewPrincipalRunCommand() *cobra.Command {
 		keepAliveMinimumInterval time.Duration
 
 		redisAddress         string
+		redisPassword        string
 		redisCompressionType string
 		healthzPort          int
 	)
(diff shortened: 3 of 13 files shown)


jgwest (changes requested):
> Since we will be deploying principal via ArgoCD resource in GitOps operator and it does not create argocd-redis secret and makes use of argocd-redis-initial-password and because of this users are forced to create argocd-redis secret and copy password from argocd-redis-initial-password. To avoid this inconvenience we could use argocd-redis-initial-password secret itself.

Not all users of Argo CD agent are installing principal via GitOps operator. GitOps operator is a vendor-specific (Red Hat) operator. Many users would be installing Argo CD via [helm](https://github.com/argoproj/argo-helm) or [install.yaml manifests](https://argo-cd.readthedocs.io/en/stable/getting_started/).

The `(name)-redis-initial-password` Secret format (with key `admin.password`) is an argocd-operator invention which is not followed by other install methods. 

For example, upstream argocd install.yaml uses an [initContainer on the redis container to create a Secret named `argocd-redis` with key `auth`.](https://github.com/argoproj/argo-cd/blob/1988c704d54ee819fe54e3f05c6090e3e5cb3372/manifests/install.yaml#L25602). I presume Helm is similar.


To me, in this repo PR we should:
- Principal should have a new `REDIS_PASSWORD` env/param (✅ already included in this PR)
- Principal should read that value in `cmd/argocd-agent/principal.go` and pass that password throughout the codebase wherever it is needed (✅ aincluded in this PR)
- In the Principal `Deployment`, we should ensure that `REDIS_PASSWORD` is mapped to the correct Secret.
- We should not change the Secret name/field from default `argocd-redis` to `(name)-redis-initial-password`, nor change the field name from `auth` to `admin.password`


Then, in argocd-operator repo: when we create the argocd-agent principal `Deployment`, we ensure that we map the REDIS-PASSWORD to the argocd-operator specific redis Secret format. This would be an example Deployment YAML created by argocd-operator:
```yaml
kind: Deployment
#(...)
metadata:
  name: (principal Argo CD agent name)
spec:
  template:
    spec:
      containers:
      - args:
        - # (...)
        env:
        - name: REDIS_PASSWORD
          valueFrom:
            secretKeyRef:
              key: admin.password
              name: (name of CR)-initial-password  # (replace this with whatever the redis Secret name will be)
          # (non-optional env)
      # (...)
```

WDYT? Is there a scenario this doesn't cover?

In both cases, Argo CD agent will fail to start until the redis Secret is created, but that's fine because Kubernetes will keep trying until it exists.

jparsai:
> > Since we will be deploying principal via ArgoCD resource in GitOps operator and it does not create argocd-redis secret and makes use of argocd-redis-initial-password and because of this users are forced to create argocd-redis secret and copy password from argocd-redis-initial-password. To avoid this inconvenience we could use argocd-redis-initial-password secret itself.
> 
> Not all users of Argo CD agent are installing principal via GitOps operator. GitOps operator is a vendor-specific (Red Hat) operator. Many users would be installing Argo CD via [helm](https://github.com/argoproj/argo-helm) or [install.yaml manifests](https://argo-cd.readthedocs.io/en/stable/getting_started/).
> 
> The `(name)-redis-initial-password` Secret format (with key `admin.password`) is an argocd-operator invention which is not followed by other install methods.
> 
> For example, upstream argocd install.yaml uses an [initContainer on the redis container to create a Secret named `argocd-redis` with key `auth`.](https://github.com/argoproj/argo-cd/blob/1988c704d54ee819fe54e3f05c6090e3e5cb3372/manifests/install.yaml#L25602). I presume Helm is similar.
> 
> To me, in this repo PR we should:
> 
> * Principal should have a new `REDIS_PASSWORD` env/param (✅ already included in this PR)
> * Principal should read that value in `cmd/argocd-agent/principal.go` and pass that password throughout the codebase wherever it is needed (✅ included in this PR)
> * In the Principal `Deployment`, we should ensure that `REDIS_PASSWORD` is mapped to the correct Secret name/field.
>   
>   * In this repo, that would be `argocd-redis` with field `auth`
> * We should not change the Secret name/field from default `argocd-redis` to `(name)-redis-initial-password`, nor change the field name from `auth` to `admin.password`
> 
> Then, in argocd-operator repo: when we create the argocd-agent principal `Deployment`, we ensure that we map the REDIS-PASSWORD to the argocd-operator specific redis Secret format. This would be an example Deployment YAML created by argocd-operator:
> 
> ```yaml
> kind: Deployment
> #(...)
> metadata:
>   name: (principal Argo CD agent name)
> spec:
>   template:
>     spec:
>       containers:
>       - args:
>         - # (...)
>         env:
>         - name: REDIS_PASSWORD
>           valueFrom:
>             secretKeyRef:
>               key: admin.password
>               name: (name of CR)-initial-password  # (replace this with whatever the redis Secret name will be)
>           # (non-optional env)
>       # (...)
> ```
> 
> WDYT? Is there a scenario this doesn't cover?
> 
> In both cases, Argo CD agent will fail to start until the redis Secret is created, but that's fine because Kubernetes will keep trying until it exists.

Yes I missed that use case, my approach was to ensure that there is no other function in agent or upstream function that it calls has hard dependency on `argocd-redis` secret just like  `SetOptionalRedisPasswordFromKubeConfig` function that I fixed here. So I removed `argocd-redis` secret to check code still works without it. I reverted my changes PTAL now.

jgwest (approved):
LGTM, thanks @jparsai!

PR #596, opened 2025-10-07
fix: revert non-source updates to AppProjects and Repositories 

### LINKED ISSUE #546: Direct updates on AppProject and Repository must be reverted by the agent
**Describe the bug**

The control plane is the source of truth in managed mode. So, any direct changes made by the user on the workload cluster must be reverted by the agent. We have already added this functionality for [apps](https://github.com/argoproj-labs/argocd-agent/pull/400). We need to follow a similar logic for AppProjects and Repositories. 

1. Create an AppProject/Repository on the control plane
2. The resource will get synced to the workload cluster
3. Make changes to the resource on the workload cluster
4. The changes are not reverted.

**Expected behavior**
The agent must revert any direct changes made to the resources on the workload cluster.

**Additional context**

chetan-rns:

This PR extends https://github.com/argoproj-labs/argocd-agent/pull/400

All updates to the AppProject and the Repository must come from the source, i.e, the principal in managed mode and the agent in autonomous mode. Any changes made by the user on the non-source cluster must be reverted to keep the resources in sync with the source cluster. In this PR:

* Use a generic cache to store spec to avoid code repetition.
* Revert the invalid updates and deletions in the respective callbacks
* We also use the same cache to differentiate between user-initiated deletion and the valid deletions coming from the source. 
* Include unit tests and e2e tests

Fixes #546 

1. Update/Delete the AppProject on a non-source cluster.
2. It must be reverted to be in sync with the source
3. Update/Delete the Repository on the agent and observe the agent reverting the changes.

**Checklist**


internal/manager/application/application.go
```diff
@@ -258,6 +267,13 @@ func (m *ApplicationManager) UpdateManagedApp(ctx context.Context, incoming *v1a
 
 	if deletionTimestampChanged {
 		logCtx.Infof("deletionTimestamp of managed agent changed from nil to non-nil, so deleting Application")
+		// Mark this as a valid deletion so the callback does not treat it as a user-initiated deletion.
+		if m.deletions != nil {
```
jgwest, line 271: Do we need this nil check? It seems like (at present) this value can never be nil (it is always set to non nil in `NewAgent`.

Is there a scenario where we would not need deletion tracking (and thus the object would be nil for those scenarios?)

internal/manager/manager.go
```diff
@@ -187,3 +193,88 @@ func (o *ObservedResources) Len() int {
...
+	}
+}
+
+func (d *DeletionTracker) MarkExpected(uid types.UID) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	d.expected[uid] = true
+}
+
+func (d *DeletionTracker) IsExpected(uid types.UID) bool {
```
jgwest: `IsExpected` also removes the UID from the map (in addition to returning the value).  IMHO it would be beneficial to rename the function to reflect that behaviour, e.g. something like `RemoveExpected`?

agent/agent.go
```diff
@@ -482,3 +481,46 @@ func (a *Agent) healthzHandler(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusServiceUnavailable)
 	}
 }
+
+func (a *Agent) populateSourceCache(ctx context.Context) {
```
jgwest: Should we return the errors that are generated by this function, rather than logging and burying them?

Just by looking at the nature of the errors, they seem fairly fatal.

internal/manager/application/application.go
```diff
@@ -616,7 +632,7 @@ func (m *ApplicationManager) List(ctx context.Context, selector backend.Applicat
 
 // RevertManagedAppChanges compares the actual spec with expected spec stored in cache,
 // if actual spec doesn't match with cache, then it is reverted to be in sync with cache, which is same as principal.
-func (m *ApplicationManager) RevertManagedAppChanges(ctx context.Context, app *v1alpha1.Application) bool {
+func (m *ApplicationManager) RevertManagedAppChanges(ctx context.Context, app *v1alpha1.Application, appCache *cache.ResourceCache[v1alpha1.ApplicationSpec]) bool {
```
jgwest: For the name of this function, perhaps a name like `RevertManagedAppChangesIfNeeded`. 

(The function had this old name before your PR, but it can't hurt to tweak it now, while we're working on this part of the code!)

The `IfNeeded` suffix helps reflect that the function is called in many cases where no revert is needed. The name is a bit longer, but it makes the logic of calling functions a bit clearer IMHO. WDYT?

If you agree, we could also add `IfNeeded` to other `Revert*` functions in this PR, like `RevertRepositoryChanges`.

test/e2e/fixture/fixture.go
```diff
@@ -167,6 +170,33 @@ func WaitForDeletion(ctx context.Context, kclient KubeClient, obj KubeObject) er
 	return fmt.Errorf("WaitForDeletion: timeout waiting for deletion of %s/%s", key.Namespace, key.Name)
 }
 
+// EnsureUpdate will ensure that the object is updated by retrying if there is a conflict.
+func EnsureUpdate(ctx context.Context, kclient KubeClient, obj KubeObject, updateFn func(obj KubeObject)) error {
```
jgwest, line 174: :+1:

principal/server.go
```diff
@@ -909,3 +881,46 @@ func (s *Server) healthzHandler(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "text/plain")
 	w.WriteHeader(http.StatusOK)
 }
+
+func (s *Server) populateSourceCache(ctx context.Context) {
```
jgwest: Likewise here, the errors generated by this function seem fairly fatal, so I presume it couldn't hurt to return them, rather than just logging.

internal/manager/manager.go
```diff
@@ -187,3 +193,88 @@ func (o *ObservedResources) Len() int {
...
+	metav1.Object
+}
+
+type resourceManager[R kubeResource] interface {
+	Create(ctx context.Context, obj R) (R, error)
+}
+
+// RevertUserInitiatedDeletion detects if a resource deletion was unauthorized and recreates the resource.
+// Returns true if the resource was recreated, false otherwise.
+func RevertUserInitiatedDeletion[R kubeResource](ctx context.Context,
```
jgwest, line 208: We can return errors from this function, rather than just logging them. 
- It's true that some (or all) of the calling functions might themselves just be logging them, but it at least opens up the possibility for handling the error by callers.

WDYT?

internal/manager/appproject/appproject.go
```diff
@@ -363,6 +366,41 @@ func (m *AppProjectManager) CompareSourceUID(ctx context.Context, incoming *v1al
 	return true, string(incoming.UID) == sourceUID, nil
 }
 
+// RevertAppProjectChanges compares the actual spec with expected spec stored in cache,
+// if actual spec doesn't match with cache, then it is reverted to be in sync with cache, which is same as the source cluster.
+func (m *AppProjectManager) RevertAppProjectChanges(ctx context.Context, project *v1alpha1.AppProject, projectCache *cache.ResourceCache[v1alpha1.AppProjectSpec]) bool {
```
jgwest: We likewise may want to return errors here, rather than just logging them.

jgwest (changes requested):
PR looks great! Thanks for combining the logic to avoid duplication where appropriate.

While reviewing, the only items that jumped out at me were a few nits; feel free to disagree if they don't make sense.

internal/manager/application/application.go
chetan-rns: We don't initialize the deletion tracking map for the principal. But we use the `ApplicationManager` struct in both the agent and the principal. Added the nil check as a safety measure. Although this particular method is not used in the principal.

PR #605, opened 2025-10-09
fix: recreate application in managed mode upon deletion from spoke

Rizwana777:
Application was not recreating when deleted from managed mode even though Principal is not in deletion state, In this fix application is recreating immediately once it is deleted from managed mode

Fixes #?
https://issues.redhat.com/browse/GITOPS-7836

**Checklist**

## Release Notes

* **New Features**
  * Applications from the principal cluster now automatically recreate if deleted from the managed cluster.

* **Tests**
  * Added test cases validating app recreation scenarios and handling expected deletions.


principal/event.go
```diff
@@ -257,6 +257,21 @@ func (s *Server) processApplicationEvent(ctx context.Context, agentName string,
 			// When we receive an 'application deleted' event from a managed agent, it should only cause the principal application to be deleted IF the principal Application is ALSO in the deletion state (deletionTimestamp is non-nil)
 			if app.DeletionTimestamp == nil {
 				logCtx.Warn("application was detected as deleted from managed agent, even though principal application is not in deletion state")
+				// In managed mode, if the application was deleted from the agent but the principal
+				// application is not in deletion state, we should recreate it by sending the
+				// application spec back to the managed agent
+				logCtx.Info("Recreating application in managed agent to maintain desired state")
```
chetan-rns: We maintain a cache of the desired state app spec on the agent. Can we use that to [revert the deletion](https://github.com/argoproj-labs/argocd-agent/blob/main/principal/callbacks.go#L130)? 

Ref: https://github.com/argoproj-labs/argocd-agent/pull/567

Also, some of these methods are being refactored to avoid duplication here https://github.com/argoproj-labs/argocd-agent/pull/596

Rizwana777: I will update this logic

chetan-rns: @Rizwana777 https://github.com/argoproj-labs/argocd-agent/pull/596 has been merged

Rizwana777: Thanks @chetan-rns  I will modify my PR accordingly

test/e2e/resync_test.go
```diff
@@ -1018,6 +1018,60 @@ func sampleRepository() *corev1.Secret {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      app.Name,
+			Namespace: "argocd",
+		},
+	}, metav1.DeleteOptions{})
+	requires.NoError(err)
+
+	t.Log("Verify application is deleted from managed agent")
+	requires.Eventually(func() bool {
+		app := argoapp.Application{}
```
chetan-rns: This check may not always pass. What if the app gets recreated before we reach this check?

test/e2e/resync_test.go
```diff
@@ -1018,6 +1018,60 @@ func sampleRepository() *corev1.Secret {
 	}
 }
 
+// Test_ManagedAppRecreationOnDirectDeletion tests the scenario where:
+//  1. A managed application exists on both principal and managed agent
+//  2. The application is directly deleted from the managed agent (while agent is running)
+//  3. The agent should detect this unauthorized deletion and recreate the application automatically
+func (suite *ResyncTestSuite) Test_ManagedAppRecreationOnDirectDeletion() {
```
chetan-rns: Can we move this test to `appcache_test.go`? We have all the revert-based tests in that file. For example, here's a similar test for the autonomous agent https://github.com/argoproj-labs/argocd-agent/blob/main/test/e2e/appcache_test.go#L235

chetan-rns (approved):
Thanks @Rizwana777!

LGTM!

PR #631, opened 2025-10-30
feat(ctl): add a 'check-config' principal/agent commands to argocd-agentctl which will verify the  user's configuration

Rizwana777:
JIRA - https://issues.redhat.com/browse/GITOPS-7958
Commands implemented via cobra

Implemented check-config with principal and agent subcommands

Principal: requires --principal-namespace; --principal-context is available and optional.

Agent: requires all four flags (--agent-context, --agent-namespace, --principal-context, --principal-namespace).

Principal checks implemented: 
Verify CA secret exists and is valid: argocd-agent-ca
Verify principal gRPC TLS secret exists and is valid: argocd-agent-principal-tls
Verify resource proxy TLS secret exists and is valid: argocd-agent-resource-proxy-tls
Verify JWT signing key secret exists and parses: argocd-agent-jwt
Argo CD cluster-scoped, apps-any-namespace, OpenShift Route host matches IPS/DNS.

Agent checks implemented:
Verify agent CA secret with ca.crt exists: argocd-agent-ca
Verify agent client TLS secret exists, not expired, and signed by principal CA: argocd-agent-client-tls
Verify a namespace on principal matches agent cert CN (e.g., CN jonathans-cluster => namespace jonathans-cluster)
Also runs all principal checks (above) using provided principal context/namespace

Steps to verify this -
make setup-e2e
make cli
principal -
argocd-agentctl check-config principal --principal-namespace argocd --principal-context <principal-context>

agent - 
argocd-agentctl check-config agent --agent-context <agent-context> --agent-namespace argocd --principal-context <principal-context> --principal-namespace argocd

## Summary by CodeRabbit

* **New Features**
  * Added a check-config CLI with principal and agent modes to validate cluster and cross-cluster configuration: Argo CD scope, TLS/CA and JWT key validity, certificate signing and expiration, route host/SAN alignment, and namespace presence; prints per-check results and exits non-zero on failures.

* **Tests**
  * Added integration-style tests covering positive and negative TLS/CA, JWT, certificate signing/expiration, route/SAN, and namespace scenarios.

### DIFF
--- a/cmd/ctl/check_config.go
+++ b/cmd/ctl/check_config.go
@@ -1,0 +1,621 @@
+// Copyright 2025 The argocd-agent Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"crypto/rsa"
+	"crypto/x509"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/argoproj-labs/argocd-agent/cmd/cmdutil"
+	"github.com/argoproj-labs/argocd-agent/internal/config"
+	"github.com/argoproj-labs/argocd-agent/internal/kube"
+	"github.com/argoproj-labs/argocd-agent/internal/tlsutil"
+	"github.com/spf13/cobra"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes"
+)
+
+type checkResult struct {
+	name string
+	err  error
+}
+
+func (r checkResult) String() string {
+	if r.err == nil {
+		return fmt.Sprintf("* %s: ✅", r.name)
+	}
+	return fmt.Sprintf("* %s: ❌\nERROR: %v", r.name, r.err)
+}
+
+func NewCheckConfigCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "check-config",
+		Short: "Validate principal and agent configuration",
+		Run: func(cmd *cobra.Command, args []string) {
+			_ = cmd.Help()
+		},
+		GroupID: "config",
+	}
+	cmd.AddCommand(NewCheckConfigPrincipalCommand())
+	cmd.AddCommand(NewCheckConfigAgentCommand())
+	return cmd
+}
+
+func NewCheckConfigPrincipalCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:   "principal",
+		Short: "Validate principal configuration",
+		Run: func(cmd *cobra.Command, args []string) {
+			if strings.TrimSpace(globalOpts.principalNamespace) == "" {
+				cmdutil.Fatal("--principal-namespace is required")
+			}
+			ctx := context.TODO()
+			clt, err := kube.NewKubernetesClientFromConfig(ctx, globalOpts.principalNamespace, "", globalOpts.principalContext)
+			if err != nil {
+				cmdutil.Fatal("Could not create Kubernetes client: %v", err)
+			}
+			results := RunPrincipalChecks(ctx, clt, globalOpts.principalNamespace)
+			printResultsAndExit(results)
+		},
+	}
+	return command
+}
+
+func NewCheckConfigAgentCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:   "agent",
+		Short: "Validate agent configuration (and principal cross-checks)",
+		Run: func(cmd *cobra.Command, args []string) {
+			if strings.TrimSpace(globalOpts.agentContext) == "" ||
+				strings.TrimSpace(globalOpts.agentNamespace) == "" ||
+				strings.TrimSpace(globalOpts.principalContext) == "" ||
+				strings.TrimSpace(globalOpts.principalNamespace) == "" {
+				cmdutil.Fatal("--agent-context, --agent-namespace, --principal-context, --principal-namespace are all required")
+			}
+			ctx := context.TODO()
+			agentClt, err := kube.NewKubernetesClientFromConfig(ctx, globalOpts.agentNamespace, "", globalOpts.agentContext)
+			if err != nil {
+				cmdutil.Fatal("Could not create agent Kubernetes client: %v", err)
+			}
+			principalClt, err := kube.NewKubernetesClientFromConfig(ctx, globalOpts.principalNamespace, "", globalOpts.principalContext)
+			if err != nil {
+				cmdutil.Fatal("Could not create principal Kubernetes client: %v", err)
+			}
+			// Run principal checks as part of agent checks
+			results := []checkResult{}
+			results = append(results, RunPrincipalChecks(ctx, principalClt, globalOpts.principalNamespace)...)
+			results = append(results, RunAgentChecks(ctx, agentClt, globalOpts.agentNamespace, principalClt, globalOpts.principalNamespace)...)
+			printResultsAndExit(results)
+		},
+	}
+	return command
+}
+
+func printResultsAndExit(results []checkResult) {
+	hasErr := false
+	fmt.Println("Configuration validation results:")
+	for _, r := range results {
+		fmt.Println(r.String())
+		if r.err != nil {
+			hasErr = true
+		}
+	}
+	if hasErr {
+		cmdutil.Fatal("one or more checks failed")
+	}
+}
+
+// getArgoCDCR retrieves the ArgoCD CR from the specified namespace, if it exists.
+func getArgoCDCR(ctx context.Context, kc *kube.KubernetesClient, ns string) (*unstructured.Unstructured, bool, error) {
+	if kc.DynamicClient == nil {
+		return nil, false, fmt.Errorf("dynamic client is not available")
+	}
+	gvr := schema.GroupVersionResource{Group: "argoproj.io", Version: "v1beta1", Resource: "argocds"}
+	list, err := kc.DynamicClient.Resource(gvr).Namespace(ns).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		// If CR doesn't exist, return found=false with no error
+		if k8serrors.IsNotFound(err) {
+			return nil, false, nil
+		}
+		// For other errors, return the error
+		return nil, false, fmt.Errorf("failed to list ArgoCD CRs: %w", err)
+	}
+	if list == nil || len(list.Items) == 0 {
+		return nil, false, nil // CR not found, use defaults
+	}
+	// Use the ArgoCD CR found in the namespace
+	return &list.Items[0], true, nil
+}
+
+// getPrincipalSecretNames reads secret names from ArgoCD CR if available, otherwise returns defaults.
+func getPrincipalSecretNames(ctx context.Context, kc *kube.KubernetesClient, ns string) (string, string, string, string, error) {
+	caSecretName := config.SecretNamePrincipalCA
+	tlsSecretName := config.SecretNamePrincipalTLS
+	proxyTLSSecretName := config.SecretNameProxyTLS
+	jwtSecretName := config.SecretNameJWT
+
+	// Try to read from ArgoCD CR
+	cr, found, err := getArgoCDCR(ctx, kc, ns)
+	if err != nil {
+		// Return error for real failures (permission denied, network issues, etc.)
+		return "", "", "", "", fmt.Errorf("failed to access ArgoCD CR in namespace %s: %w", ns, err)
+	}
+	if found {
+		// Read principal.tls.rootCASecretName
+		if val, found, err := unstructured.NestedString(cr.Object, "spec", "argoCDAgent", "principal", "tls", "rootCASecretName"); err != nil {
+			return "", "", "", "", fmt.Errorf("failed to read principal.tls.rootCASecretName from ArgoCD CR: %w", err)
+		} else if found && val != "" {
+			caSecretName = val
+		}
+		// Read principal.tls.secretName
+		if val, found, err := unstructured.NestedString(cr.Object, "spec", "argoCDAgent", "principal", "tls", "secretName"); err != nil {
+			return "", "", "", "", fmt.Errorf("failed to read principal.tls.secretName from ArgoCD CR: %w", err)
+		} else if found && val != "" {
+			tlsSecretName = val
+		}
+		// Read principal.resourceProxy.tls.secretName
+		if val, found, err := unstructured.NestedString(cr.Object, "spec", "argoCDAgent", "principal", "resourceProxy", "tls", "secretName"); err != nil {
+			return "", "", "", "", fmt.Errorf("failed to read principal.resourceProxy.tls.secretName from ArgoCD CR: %w", err)
+		} else if found && val != "" {
+			proxyTLSSecretName = val
+		}
+		// Read principal.jwt.secretName
+		if val, found, err := unstructured.NestedString(cr.Object, "spec", "argoCDAgent", "principal", "jwt", "secretName"); err != nil {
+			return "", "", "", "", fmt.Errorf("failed to read principal.jwt.secretName from ArgoCD CR: %w", err)
+		} else if found && val != "" {
+			jwtSecretName = val
+		}
+	}
+	return caSecretName, tlsSecretName, proxyTLSSecretName, jwtSecretName, nil
+}
+
+// getAgentSecretNames reads secret names from ArgoCD CR if available, otherwise returns defaults.
+func getAgentSecretNames(ctx context.Context, kc *kube.KubernetesClient, ns string) (string, string, error) {
+	caSecretName := config.SecretNameAgentCA
+	clientCertSecretName := config.SecretNameAgentClientCert
+
+	// Try to read from ArgoCD CR
+	cr, found, err := getArgoCDCR(ctx, kc, ns)
+	if err != nil {
+		// Return error for real failures (permission denied, network issues, etc.)
+		return "", "", fmt.Errorf("failed to access ArgoCD CR in namespace %s: %w", ns, err)
+	}
+	if found {
+		// Read agent.tls.rootCASecretName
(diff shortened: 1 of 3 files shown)


cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+
+	// Ensure Argo CD is cluster-scoped
+	out = append(out, checkResult{
+		name: "Verifying Argo CD is cluster-scoped",
+		err:  verifyArgoCDClusterScoped(ctx, kubeClient, principalNS),
+	})
+
+	// CA secret exists in principal namespace and is a valid TLS secret
+	out = append(out, checkResult{
+		name: fmt.Sprintf("Verifying principal public CA certificate exists and is valid (%s/%s)", principalNS, config.SecretNamePrincipalCA),
```
jgwest: JGW-TODO: What if name is different, for example `cr.Spec.ArgoCDAgent.Principal.TLS.RootCASecretName` points to a different value?

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+	}
+	return nil
+}
+
+// verifyArgoCDClusterScoped checks operator CR or argocd-cm to ensure no namespaced mode is configured.
+func verifyArgoCDClusterScoped(ctx context.Context, kc *kube.KubernetesClient, ns string) error {
+	// Try operator CR first: group argoproj.io, version v1alpha1, resource argocds
+	gvr := schema.GroupVersionResource{Group: "argoproj.io", Version: "v1alpha1", Resource: "argocds"}
+	// If CRD exists and an ArgoCD instance exists in ns, assert spec.applicationNamespaces is empty
+	if kc.DynamicClient != nil {
```
jgwest: Return an error if dynamic client is non-nil, it should always be defined

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+		return nil
+	}
+	// If neither CR nor ConfigMap found, do not fail the check
+	return nil
+}
+
+// verifyRouteHostMatchesCert checks OpenShift Route host in ns is present in IPS?DNS of the given TLS secret.
+func verifyRouteHostMatchesCert(ctx context.Context, kc *kube.KubernetesClient, ns string, tlsSecretName string) error {
+	// Discover route API by attempting a list via dynamic client
+	if kc.DynamicClient == nil {
```
jgwest, line 472: Return an error if dynamic client is nil, this should never happen

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+	if err != nil {
+		return err
+	}
+	if err := agentX509.CheckSignatureFrom(caX509); err != nil {
+		return fmt.Errorf("agent certificate not signed by principal CA: %w", err)
+	}
+	return nil
+}
+
+func namespaceMatchesAgentSubject(ctx context.Context, agentKube kubernetes.Interface, agentNS string, principalKube kubernetes.Interface, principalNS string) error {
```
jgwest: VS Code is showing a warning for this function:
`unused parameter: principalNS`

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+	parsed, err := x509FromTLSSecret(ctx, kc.Clientset, ns, tlsSecretName)
+	if err != nil {
+		return err
+	}
+
+	// Try routes.route.openshift.io
+	gvr := schema.GroupVersionResource{Group: "route.openshift.io", Version: "v1", Resource: "routes"}
+	routes, err := kc.DynamicClient.Resource(gvr).Namespace(ns).List(ctx, metav1.ListOptions{})
+	if err != nil || routes == nil {
+		return nil // skip if Route API not present or inaccessible
```
jgwest: If the Route API doesn't exist, we can return nil, but for any other error we should return the error rather than ignoring it.

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+}
+
+// verifyArgoCDClusterScoped checks operator CR or argocd-cm to ensure no namespaced mode is configured.
+func verifyArgoCDClusterScoped(ctx context.Context, kc *kube.KubernetesClient, ns string) error {
+	// Try operator CR first: group argoproj.io, version v1alpha1, resource argocds
+	gvr := schema.GroupVersionResource{Group: "argoproj.io", Version: "v1alpha1", Resource: "argocds"}
+	// If CRD exists and an ArgoCD instance exists in ns, assert spec.applicationNamespaces is empty
+	if kc.DynamicClient != nil {
+		list, err := kc.DynamicClient.Resource(gvr).Namespace(ns).List(ctx, metav1.ListOptions{})
+		if err == nil && list != nil && len(list.Items) > 0 {
```
jgwest: If an error occurs, we should return it, rather than ignoring it

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+			}
+			return nil
+		}
+	}
+	// Fallback when Operator CR is absent: check the core Argo CD ConfigMap (argocd-cm)
+	// and fail if application.namespaces is set (namespaced mode). If the ConfigMap
+	// does not exist or the key is empty, assume cluster-scoped and pass.
+	cm, err := kc.Clientset.CoreV1().ConfigMaps(ns).Get(ctx, "argocd-cm", metav1.GetOptions{})
+	if err == nil {
+		if v := strings.TrimSpace(cm.Data["application.namespaces"]); v != "" {
```
jgwest, line 428: If an error occurs, we should return it, rather than ignoring it

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+		return nil // skip if Route API not present or inaccessible
+	}
+
+	if len(routes.Items) == 0 {
+		return nil // skip if no routes exist
+	}
+
+	// Pass if any route host matches SANs
+	for _, r := range routes.Items {
+		host, _, _ := unstructured.NestedString(r.Object, "spec", "host")
```
jgwest: Should return error here if it occurs

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+// verifyArgoCDClusterScoped checks operator CR or argocd-cm to ensure no namespaced mode is configured.
+func verifyArgoCDClusterScoped(ctx context.Context, kc *kube.KubernetesClient, ns string) error {
+	// Try operator CR first: group argoproj.io, version v1alpha1, resource argocds
+	gvr := schema.GroupVersionResource{Group: "argoproj.io", Version: "v1alpha1", Resource: "argocds"}
+	// If CRD exists and an ArgoCD instance exists in ns, assert spec.applicationNamespaces is empty
+	if kc.DynamicClient != nil {
+		list, err := kc.DynamicClient.Resource(gvr).Namespace(ns).List(ctx, metav1.ListOptions{})
+		if err == nil && list != nil && len(list.Items) > 0 {
+			for _, it := range list.Items {
+				if arr, found, _ := unstructured.NestedSlice(it.Object, "spec", "applicationNamespaces"); found && len(arr) > 0 {
```
jgwest: Should return error here if it occurs

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+		return nil // skip if no routes exist
+	}
+
+	// Pass if any route host matches SANs
+	for _, r := range routes.Items {
+		host, _, _ := unstructured.NestedString(r.Object, "spec", "host")
+		if host == "" {
+			continue
+		}
+		if err := parsed.VerifyHostname(host); err == nil {
```
jgwest: Return error rather than ignoring it

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+	if _, ok := key.(*rsa.PrivateKey); !ok {
+		return fmt.Errorf("JWT signing key is not an RSA private key")
+	}
+	return nil
+}
+
+// verifyArgoCDClusterScoped checks operator CR or argocd-cm to ensure no namespaced mode is configured.
+func verifyArgoCDClusterScoped(ctx context.Context, kc *kube.KubernetesClient, ns string) error {
+	// Try operator CR first: group argoproj.io, version v1alpha1, resource argocds
+	gvr := schema.GroupVersionResource{Group: "argoproj.io", Version: "v1alpha1", Resource: "argocds"}
```
jgwest: Use `v1beta1` as `v1alpha1` is deprecated

cmd/ctl/check_config_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+func TestCheckConfigPrincipal_RequiredSecretsAndConfig_Valid(t *testing.T) {
+	principalNS := "argocd"
+	cl := fake.NewSimpleClientset()
+
+	// Create a scheme and register ArgoCD CRD
+	scheme := runtime.NewScheme()
+	// Create dynamic client with the registered types
+	dynCl := dynamicfake.NewSimpleDynamicClientWithCustomListKinds(scheme, map[schema.GroupVersionResource]string{
+		{Group: "argoproj.io", Version: "v1alpha1", Resource: "argocds"}: "ArgoCDList",
```
jgwest: Use `v1beta1` as `v1alpha1` is deprecated

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+	// does not exist or the key is empty, assume cluster-scoped and pass.
+	cm, err := kc.Clientset.CoreV1().ConfigMaps(ns).Get(ctx, "argocd-cm", metav1.GetOptions{})
+	if err == nil {
+		if v := strings.TrimSpace(cm.Data["application.namespaces"]); v != "" {
+			return fmt.Errorf("argocd-cm sets application.namespaces (%s)", v)
+		}
+		return nil
+	}
+	// If neither CR nor ConfigMap found, do not fail the check
+	return nil
```
jgwest, line 452: If neither could be found, we should return an error.

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+		return err
+	}
+	// Require RSA for now
+	if _, ok := key.(*rsa.PrivateKey); !ok {
+		return fmt.Errorf("JWT signing key is not an RSA private key")
+	}
+	return nil
+}
+
+// verifyArgoCDClusterScoped checks operator CR or argocd-cm to ensure no namespaced mode is configured.
```
jgwest: It appears that this logic is not checking for cluster scoped, but is instead checking if source namespaces field is non-empty. Which is a fine thing to check, but is different from whether the cluster is cluster-scoped.

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+			hasErr = true
+		}
+	}
+	if hasErr {
+		cmdutil.Fatal("one or more checks failed")
+	}
+}
+
+// RunPrincipalChecks validates the principal installation and related security assets.
+func RunPrincipalChecks(ctx context.Context, kubeClient *kube.KubernetesClient, principalNS string) []checkResult {
```
jgwest, line 221: Many of the checks in this function assume that the default Secret name is being used.

For example, 
```go
	out = append(out, checkResult{
		name: fmt.Sprintf("Verifying principal public CA certificate exists and is valid (%s/%s)", principalNS, config.SecretNamePrincipalCA),
		err:  principalCheckCA(ctx, kubeClient.Clientset, principalNS),
	})
```

Here we assume that the principal's public CA certificate will always be a Secret with the name defined in `config.SecretNamePrincipalCA` (`"argocd-agent-ca"`).

While it's fine to fall back to the default Secret name, we should first check that:
- If there exists an `ArgoCD` CR...
    - then, check the `ArgoCD` CR's `.spec.argoCDAgent.principal.tls.rootCASecretName` field, and if it is non-empty, use that value as the secret name

Likewise, for all the other checks in this function, for example, `config.SecretNamePrincipalTLS` can be replaced by the value in `.spec.argoCDAgent.principal.tls.secretName`

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+
+	// JWT signing key exists
+	out = append(out, checkResult{
+		name: fmt.Sprintf("Verifying JWT signing key exists and is parseable (%s/%s)", principalNS, config.SecretNameJWT),
+		err:  jwtKeyValid(ctx, kubeClient.Clientset, principalNS, config.SecretNameJWT),
+	})
+
+	// Route host matches TLS SANs (OpenShift-only)
+	out = append(out, checkResult{
+		name: "Verifying principal TLS secret ips/dns match Route host (OpenShift)",
```
jgwest: Let's skip this check if we detect that Route CR does not exist on the principal cluster. (It will be weird to users to see a check for OpenShift Route if there K8s cluster has nothing to do with OpenShift)

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,376 @@
...
+
+	return out
+}
+
+// RunAgentChecks validates agent-side security assets and cross-validates them
+// against the principal cluster.
+func RunAgentChecks(ctx context.Context, agentKube kubernetes.Interface, agentNS string, principalKube kubernetes.Interface, principalNS string) []checkResult {
+	out := []checkResult{}
+
+	// Agent CA secret exists and has CA data (opaque secret with ca.crt)
```
jgwest, line 320: Similar to my comment in `RunPrincipalChecks`, users can customize the Secret via `.spec.argoCDAgent.agent.(...)` fields, so we should use those values instead, if they are defined.

cmd/ctl/check_config_test.go
```diff
@@ -0,0 +1,152 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	dynamicfake "k8s.io/client-go/dynamic/fake"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
```
jgwest, line 47: Correct me if I'm wrong, but I think these unit tests are only check the 'no error' case. They should also check the error case: e.g. if something is missing, an error should be printed.

Rizwana777: I have not covered all the error cases before, I have added error tests now, PTAL and let me know if I need to add more

cmd/ctl/check_config.go
Rizwana777: Now I have added kubectl get applications --all-namespaces logic to check whether it is cluster scoped, please let me know if this is correct or is there any other way we can check this, thank you

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,580 @@
...
+	}
+	// If list succeeds, we can confirm cluster-scoped mode is working
+	return nil
+}
+
+// routeAPIExists checks if the OpenShift Route API exists on the cluster.
+func routeAPIExists(kc *kube.KubernetesClient) bool {
+	if kc.Clientset == nil {
+		return false
+	}
```
jgwest, line 459: Should return an error if `kc.Clientset` is nil

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,580 @@
...
+	_, err = principalKube.CoreV1().Namespaces().Get(ctx, subj, metav1.GetOptions{})
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return fmt.Errorf("namespace '%s' not found on principal cluster", subj)
+		}
+		return err
+	}
+	// optional: ensure agent's own namespace exists
+	_, _ = agentKube.CoreV1().Namespaces().Get(ctx, agentNS, metav1.GetOptions{})
+	return nil
```
jgwest: What is the purpose of these two final lines?

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,580 @@
...
+// x509FromTLSSecret retrieves a Kubernetes TLS secret and parses the first certificate
+// in the chain into an *x509.Certificate.
+func x509FromTLSSecret(ctx context.Context, kubeClient kubernetes.Interface, ns, name string) (*x509.Certificate, error) {
+	cert, err := tlsutil.TLSCertFromSecret(ctx, kubeClient, ns, name)
+	if err != nil {
+		return nil, err
+	}
+	if len(cert.Certificate) == 0 || cert.Certificate[0] == nil {
+		return nil, fmt.Errorf("%s/%s: secret does not contain certificate data", ns, name)
+	}
```
jgwest, line 612: This code assumes there is only 1 certificate. We should return an error if there is more than one (OR we should parse them all)

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,580 @@
...
+// getPrincipalSecretNames reads secret names from ArgoCD CR if available, otherwise returns defaults.
+func getPrincipalSecretNames(ctx context.Context, kc *kube.KubernetesClient, ns string) (string, string, string, string) {
+	caSecretName := config.SecretNamePrincipalCA
+	tlsSecretName := config.SecretNamePrincipalTLS
+	proxyTLSSecretName := config.SecretNameProxyTLS
+	jwtSecretName := config.SecretNameJWT
+
+	// Try to read from ArgoCD CR
+	cr, err := getArgoCDCR(ctx, kc, ns)
+	if err != nil {
```
jgwest, line 159: We should return an error here if err is non-nil.

Or, said another way:

We shouldn't assume that an error means the ArgoCD CR isn't available. It may just be that another error occurred.

If you want to know whether ArgoCD CRs could not be found, for example, you can instead return a boolean from `getArgoCDCR`

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,580 @@
...
+		// Log warning but continue with defaults
+		// This allows the check to proceed even if CR access fails
+		fmt.Printf("WARNING: Failed to access ArgoCD CR in namespace %s: %v. Using default secret names.\n", ns, err)
+		return caSecretName, tlsSecretName, proxyTLSSecretName, jwtSecretName
+	}
+	if cr != nil {
+		// Read principal.tls.rootCASecretName
+		if val, found, err := unstructured.NestedString(cr.Object, "spec", "argoCDAgent", "principal", "tls", "rootCASecretName"); err != nil {
+			fmt.Printf("WARNING: Failed to read principal.tls.rootCASecretName from ArgoCD CR: %v. Using default.\n", err)
```
jgwest: Please return errors for this and the other cases in this function.

cmd/ctl/check_config.go
```diff
@@ -0,0 +1,580 @@
...
+	return caSecretName, tlsSecretName, proxyTLSSecretName, jwtSecretName
+}
+
+// getAgentSecretNames reads secret names from ArgoCD CR if available, otherwise returns defaults.
+func getAgentSecretNames(ctx context.Context, kc *kube.KubernetesClient, ns string) (string, string) {
+	caSecretName := config.SecretNameAgentCA
+	clientCertSecretName := config.SecretNameAgentClientCert
+
+	// Try to read from ArgoCD CR
+	cr, err := getArgoCDCR(ctx, kc, ns)
```
jgwest: As above, this function should return errors when they occur.

cmd/ctl/check_config.go
Rizwana777: I left this line in by mistake, it’s not required. I will remove it.

cmd/ctl/check_config.go
Rizwana777: Updated code to return an error if there is more than one certificate , enforcing exactly one certificate per secret, Please let me know if we need to parse all the certificates

PR #664, opened 2025-11-27
feat: redis TLS encryption enabled by default for all connections

### LINKED ISSUE #454: Redis proxy should support TLS (inbound and outbound)

The Redis proxy introduced in #436 uses plain text connections only, for both incoming connections from the Argo CD components (server and repo-server) as well as when proxying connections to the local argocd-redis for certain keys. The Redis cache may contain sensitive data, and trusting the cluster's network isn't a good idea.

1. The Redis proxy component should be configurable to accept only TLS connection on its listening endpoint. The configuration would likely require the endpoint's TLS certificate and private key to be specified by the user, or the name of a Kubernetes TLS secret holding the appropriate certificate/key combination.
2. The Redis proxy component should be configurable to use TLS when connecting to its upstream argocd-redis on the control plane. The configuration would likely require the CA certificate used to sign the upstream's TLS certificate to be specified by the user, or the name of a Kubernetes secret holding the CA certificate.

**Completion Criteria**:
- Add cmd line parameters to argocd-agent agent/principal for redis tls config
	- Command line parameter to tell agent/principal that it should require tls for redis
	- Command line paramter which specifies the TLS certificate that agent/principal should use to validate the redis TLS connection
- Change the default install manifests/scripts for argocd-agent, so that they now use TLS for redis
	- Adding redis tls config to the `hack/dev-env` scripts
		- This includes enabling redis on upstream argo cd which is installed to control-plane, autonomous-agent, and managed-agent vclusters for dev and E2E tests
	- Adding redis tls config to `install/kubernetes` (enabled by default)
	- Adding redis tls config to `install/helm-repo` (enabled by default)
- From now on, the argocd-agent E2E tests would ONLY run on 'redis w/ TLS' configuration (that is, it is the 'happy path')
- Update install documentation to ensure redis tls is enabled, and generating redis certificate (for k8s case only, openshift will have a separate configuration mechanism)

Rizwana777:
fixes #454 

JIRA - https://issues.redhat.com/browse/GITOPS-8091

Using argocd-agentctl for generating CA certificates required for tls enablement

 E2E Tests
- All E2E tests run with Redis TLS enabled (the "happy path")
- Tests pass on both CI (Linux with MetalLB) and local development
- Reverse tunnel tested for remote vcluster scenarios

Unit Tests
- All unit tests updated to provide required `cacheRefreshInterval`

## Release Notes

* **New Features**
  * Added Redis TLS encryption support for secure connections with configurable CA certificates
  * Support for multiple TLS configuration modes: insecure (testing only), CA file path, or Kubernetes secret-based CA
  * New CLI flags and environment variables for Redis TLS configuration

* **Documentation**
  * Updated deployment guides with Redis TLS setup instructions
  * Added reverse tunnel setup requirements to testing documentation


hack/dev-env/Procfile.e2e
```diff
@@ -1,3 +1,7 @@
-principal: hack/dev-env/start-principal.sh
-agent-managed: sleep 5s && REDIS_ADDR=$MANAGED_AGENT_REDIS_ADDR  hack/dev-env/start-agent-managed.sh
-agent-autonomous: sleep 5s && REDIS_ADDR=$AUTONOMOUS_AGENT_REDIS_ADDR  hack/dev-env/start-agent-autonomous.sh
+pf-control-plane: kubectl port-forward svc/argocd-redis -n argocd 6380:6379 --context vcluster-control-plane
+pf-managed: kubectl port-forward svc/argocd-redis -n argocd 6381:6379 --context vcluster-agent-managed
+pf-autonomous: kubectl port-forward svc/argocd-redis -n argocd 6382:6379 --context vcluster-agent-autonomous
+pf-argocd-server: kubectl port-forward svc/argocd-server -n argocd 8444:443 --context vcluster-control-plane
```
jannfis: Why are these required?

docs/getting-started/kubernetes/index.md
```diff
@@ -156,6 +156,82 @@ argocd-agentctl jwt create-key \
...
+
+!!! warning "Redis TLS is Required"
+    Redis TLS is **enabled by default** in argocd-agent. All Redis connections must use TLS.
+
+#### Generate Certificates and Create Secret
+
+```bash
+# Generate CA certificate
+openssl genrsa -out redis-ca.key 4096
+openssl req -new -x509 -days 3650 -key redis-ca.key -out redis-ca.crt -subj "/CN=Redis CA"
```
jannfis: Is there a reason why we need a dedicated CA for Redis? Why don't we reuse the agent CA that is already existing?

hack/dev-env/configure-redis-tls.sh
```diff
@@ -0,0 +1,264 @@
+#!/bin/bash
```
jannfis: I feel that this should be integrated into argocd-agentctl instead of being a shell script, to be honest.

It does have all the necessary stubs (multiple contexts, PKI for development, etc). With https://github.com/argoproj-labs/argocd-agent/pull/676 merged, it will be able to use external certificates (i.e. generated by cert manager) to configure TLS for the agents, so could apply to Redis TLS as well.

hack/dev-env/start-agent-autonomous.sh
```diff
@@ -34,14 +34,58 @@ if [ -f "$E2E_ENV_FILE" ]; then
     source "$E2E_ENV_FILE"
 fi
 
+# Check if Redis TLS certificates exist
```
jannfis: This seems to be adding quite some unnecessary complexity. Neither agent nor principal need to load TLS certificates from files, they load from Kubernetes secrets. So there is no need to extract the certs from the secrets, and configure principal/agent components to use the certs from the extracted data.

hack/dev-env/start-agent-autonomous.sh
```diff
@@ -34,14 +34,58 @@ if [ -f "$E2E_ENV_FILE" ]; then
...
+else
+    echo "Redis TLS certificates not found, running without TLS"
+    echo "Run './hack/dev-env/gen-redis-tls-certs.sh' to generate certificates"
+fi
+
+# Set Redis address for local development
+# Agents connect to their vcluster Redis via localhost port-forward
+# (in-cluster DNS is not accessible from host machine)
+if [ -z "${ARGOCD_AGENT_REDIS_ADDRESS}" ]; then
+    # Default to localhost:6382 for local E2E testing (requires port-forward)
```
jannfis: Redis runs in the cluster, so `localhost` is a bit off?

principal/tracker/tracking.go
```diff
@@ -72,7 +72,10 @@ func (p *Tracker) Track(eventID string, agentName string) (<-chan *cloudevents.E
 	if ok {
 		return nil, fmt.Errorf("resource with ID %s already tracked", eventID)
 	}
-	ch := make(chan *cloudevents.Event)
+	// Use a buffered channel to prevent blocking sends.
+	// The sender (processRedisEventResponse) and receiver (sendSynchronousRedisMessageToAgent)
+	// both run in goroutines, and an unbuffered channel can deadlock if timing is off.
+	ch := make(chan *cloudevents.Event, 1)
```
jannfis: How is this change related?

principal/resource.go
```diff
@@ -39,7 +39,7 @@ const resourceRequestRegexp = `^/(?:api|apis|(?:api|apis/(?P<group>[^\/]+))/(?P<
 // resource.
 //
 // TODO(jannfis): Make the timeout configurable
-const requestTimeout = 10 * time.Second
+const requestTimeout = 30 * time.Second
```
jannfis: How is this change related?

Mangaal (commented):
I feel the issue related to e2e failure for the log was due to the log fetch window. It was resolved after you increased the window size. Could you please apply the same adjustment for the autonomous mode as well?

Makefile
```diff
@@ -56,6 +56,46 @@ endif
 ifneq (${ARGOCD_AGENT_IN_CLUSTER},)
 	./hack/dev-env/restart-all.sh
 endif
+	@echo ""
```
jgwest: Can we move this code into its own shell script, for example `./hack/dev-env/setup-e2e-enable-redis-tls.sh`  (to make up a random name)

hack/dev-env/configure-redis-tls.sh
```diff
@@ -0,0 +1,264 @@
...
+# Copyright 2024 The argocd-agent Authors
```
jgwest: JGW: Is there a less complex way to enable TLS on redis than this?

agent/agent.go
```diff
@@ -139,10 +142,11 @@ type AgentOption func(*Agent) error
...
 	a := &Agent{
-		version:      version.New("argocd-agent"),
-		deletions:    manager.NewDeletionTracker(),
-		sourceCache:  cache.NewSourceCache(),
-		inflightLogs: make(map[string]struct{}),
+		version:              version.New("argocd-agent"),
+		deletions:            manager.NewDeletionTracker(),
+		sourceCache:          cache.NewSourceCache(),
+		inflightLogs:         make(map[string]struct{}),
+		cacheRefreshInterval: 30 * time.Second, // Default interval, can be overridden via WithCacheRefreshInterval
```
jgwest: Rather than setting a default, let's just return an error if WithCacheRefreshInterval didn't set the vaue.

agent/agent.go
```diff
@@ -421,20 +447,22 @@ func (a *Agent) Start(ctx context.Context) error {
...
-				select {
-				case <-ticker.C:
-					a.addClusterCacheInfoUpdateToQueue()
-				case <-a.context.Done():
-					return
-				}
+	// Both managed and autonomous agents need to send cluster cache info updates
+	go func() {
+		// Send initial update immediately on startup (don't wait for first ticker)
+		a.addClusterCacheInfoUpdateToQueue()
```
jgwest: Is this change required for/related to this PR?

agent/agent.go
```diff
@@ -321,7 +325,29 @@ func NewAgent(ctx context.Context, client *kube.KubernetesClient, namespace stri
...
+	// Create TLS config for cluster cache Redis client (same as for Redis proxy)
+	var clusterCacheTLSConfig *tls.Config = nil
+	if a.redisProxyMsgHandler.redisTLSEnabled {
+		clusterCacheTLSConfig = &tls.Config{
+			MinVersion: tls.VersionTLS12,
+		}
+		if a.redisProxyMsgHandler.redisTLSInsecure {
+			log().Warn("INSECURE: Not verifying Redis TLS certificate for cluster cache")
+			clusterCacheTLSConfig.InsecureSkipVerify = true
+		} else if a.redisProxyMsgHandler.redisTLSCAPath != "" {
```
jgwest, line 419: Should we return an error if `a.redisProxyMsgHandler.redisTLSCAPat == ""` or is this still a valid configuration?

cmd/argocd-agent/agent.go
```diff
@@ -176,6 +181,23 @@ func NewAgentRunCommand() *cobra.Command {
...
+					cmdutil.Fatal("Only one Redis TLS mode can be specified: --redis-tls-insecure or --redis-tls-ca-path")
+				}
+
+				if redisTLSInsecure {
+					logrus.Warn("INSECURE: Not verifying Redis TLS certificate")
+					agentOpts = append(agentOpts, agent.WithRedisTLSInsecure(true))
+				} else if redisTLSCAPath != "" {
+					logrus.Infof("Loading Redis CA certificate from file %s", redisTLSCAPath)
+					agentOpts = append(agentOpts, agent.WithRedisTLSCAPath(redisTLSCAPath))
+				}
```
jgwest, line 291: `if redisTLSCAPath == ""` is this still a valid configuration, or should we just return an error?

cmd/argocd-agent/principal.go
```diff
@@ -86,6 +86,16 @@ func NewPrincipalRunCommand() *cobra.Command {
...
 		healthzPort          int
+
+		// Redis TLS configuration
+		redisTLSEnabled              bool
+		redisServerTLSCertPath       string
+		redisServerTLSKeyPath        string
+		redisServerTLSSecretName     string
+		redisUpstreamTLSCAPath       string
+		redisUpstreamTLSCASecretName string
+		redisUpstreamTLSInsecure     bool
```
jgwest: In this PR, I believe you are using the following terminology:
- `redisServer*` - agent redis proxy
- `redisUpstream*` - argo cd redis (e.g the redis Deployment that is normally used directly by Argo CD)

If I have this correct, my vote would instead be to use this naming scheme:
- `redisServer*` -> `redisProxyServer*`
- `redisUpstream*` -> `redis*`

This makes it more obivous that redisServer is referring specifically to the redis proxy functionality, while `redis` is just the normal connection to redis. WDYT?

If you agree, we can make the change throughout the PR.

hack/dev-env/start-agent-autonomous.sh
```diff
@@ -34,14 +34,58 @@ if [ -f "$E2E_ENV_FILE" ]; then
...
+kubectl --context vcluster-agent-autonomous -n argocd get secret argocd-agent-client-tls \
+  -o jsonpath='{.data.tls\.key}' | base64 -d > "${TLS_KEY_PATH}" || { echo "ERROR: Failed to extract TLS key from argocd-agent-client-tls secret"; exit 1; }
+kubectl --context vcluster-agent-autonomous -n argocd get secret argocd-agent-ca \
+  -o jsonpath='{.data.ca\.crt}' | base64 -d > "${ROOT_CA_PATH}" || { echo "ERROR: Failed to extract CA cert from argocd-agent-ca secret"; exit 1; }
+echo " mTLS client certificates and CA extracted"
+
 go run github.com/argoproj-labs/argocd-agent/cmd/argocd-agent agent \
     --agent-mode autonomous \
     --creds mtls:any \
+    --tls-client-cert="${TLS_CERT_PATH}" \
```
jgwest: Can you provide more info on why we are including the `--tls-*` and `--root-ca-path` args here?
- For example, if there are necessary, why did we previously not have them?
- e.g. I'm referring to the non-redis params (I understand why we added the redis items)

hack/dev-env/configure-argocd-redis-tls.sh
```diff
@@ -0,0 +1,405 @@
...
+    fi
+    
+    echo "  argocd-application-controller configured"
+fi
+
+echo ""
+echo "Scaling up Argo CD components with Redis TLS configuration..."
+echo ""
+
+# Read replica counts from ConfigMap (created by configure-redis-tls.sh)
```
jgwest: What's the goal of this replica count check code? Shouldn't the replicas always just be 1?

hack/dev-env/configure-redis-tls.sh
```diff
@@ -0,0 +1,264 @@
...
+if ! kubectl get deployment argocd-redis -n ${NAMESPACE} &>/dev/null; then
+    echo "Error: argocd-redis Deployment not found in namespace ${NAMESPACE}"
+    exit 1
+fi
+
+echo "Found Redis Deployment"
+echo ""
+
+# Scale down ArgoCD components that connect to Redis BEFORE enabling TLS
+# This prevents SSL errors during the transition (old pods trying to connect without TLS)
```
jgwest: `SSL errors` -> `TLS errors`

hack/dev-env/configure-argocd-redis-for-tls.sh
```diff
@@ -0,0 +1,405 @@
+#!/bin/bash
+# Configure Argo CD components to use Redis TLS (for E2E tests and local development)
```
jgwest: I recommend renaming this file from `configure-argocd-redis-tls.sh` to `configure-argocd-redis-for-tls.sh`, IMHO it makes it more obvious what the purpose of the file is. WDYT?

internal/argocd/cluster/cluster.go
```diff
@@ -131,6 +132,13 @@ func (m *Manager) SetClusterCacheStats(clusterInfo *event.ClusterCacheInfo, agen
 		if existingClusterInfo.CacheInfo.LastCacheSyncTime != nil {
 			newClusterInfo.CacheInfo.LastCacheSyncTime = existingClusterInfo.CacheInfo.LastCacheSyncTime
 		}
+	} else {
```
jgwest: Is this change related to this PR?

principal/redisproxy/redisproxy.go
```diff
@@ -740,25 +833,96 @@ func (are *argoCDRedisWriterInternal) writeToArgoCDRedisSocket(logCtx *logrus.En
...
+		return nil, fmt.Errorf("unable to connect to redis '%s': %w", rp.principalRedisAddress, err)
+	}
+
+	// Check if upstream TLS configuration is provided
+	hasUpstreamTLSConfig := rp.upstreamTLSCA != nil || rp.upstreamTLSCAPath != "" || rp.upstreamTLSInsecure
+
+	// Warn if server TLS is enabled but upstream TLS is not configured
+	// This creates a security gap: agent→proxy is encrypted, but proxy→redis is not
+	if rp.tlsEnabled && !hasUpstreamTLSConfig {
+		logCtx.Warn("SECURITY WARNING: Redis proxy server has TLS enabled, but no upstream TLS configuration provided. Connection to principal Redis will be UNENCRYPTED. This exposes data in transit within the cluster.")
```
jgwest: It seems like this should return an error, rather than just logging a WARN?

principal/tracker/tracking.go
```diff
@@ -72,7 +72,10 @@ func (p *Tracker) Track(eventID string, agentName string) (<-chan *cloudevents.E
 	if ok {
 		return nil, fmt.Errorf("resource with ID %s already tracked", eventID)
 	}
-	ch := make(chan *cloudevents.Event)
+	// Use a buffered channel to prevent blocking sends.
```
jgwest: Is this change necessary? There is no inherent reason that a channel with size 0 with deadlock, if the goroutine message passing logic is correct.

If yes, can you provide more details on why this is required?

test/e2e/README.md
```diff
@@ -18,24 +18,96 @@ Both the vcluster and Argo CD installations require that LoadBalancer functional
 
 ## Running the tests
 
-To setup the test environment on the cluster, execute the following command from the repository root:
+### Step 1: Setup the test environment
```
jgwest: JGW: Do I think this is a value add?

test/e2e/redis_proxy_test.go
```diff
@@ -300,6 +328,11 @@ func (suite *RedisProxyTestSuite) Test_RedisProxy_AutonomousAgent_Argo() {
 
 	requires.NotNil(msgChan)
 
+	// Wait for SSE stream to fully establish and Redis SUBSCRIBE to propagate
+	// This prevents a race condition where the pod is deleted before the subscription is active
+	t.Log("Waiting for SSE stream to fully establish...")
+	time.Sleep(5 * time.Second)
```
jgwest: Remove sleep

test/e2e/redis_proxy_test.go
```diff
@@ -539,7 +595,7 @@ func createArgoCDAPIClient(ctx context.Context, argoServerEndpoint string, passw
 // - resource tree events (changes in Application resources) are an example of one type of data that can be received via this API
 func streamFromEventSourceNew(ctx context.Context, eventSourceAPIURL string, sessionToken string, t *testing.T) (chan string, error) {
 
-	msgChan := make(chan string)
+	msgChan := make(chan string, 100) // Buffered channel to prevent message loss
```
jgwest: If the logic is implemented correctly, there should be no need to use a channel of greater size. If you believe a channel of size 0 can is not correct, can you provide additional detail?

test/e2e/redis_proxy_test.go
```diff
@@ -594,17 +650,27 @@ func streamFromEventSourceNew(ctx context.Context, eventSourceAPIURL string, ses
 			req.AddCookie(cookie)
 
 			tr := &http.Transport{
-				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+				TLSClientConfig:       &tls.Config{InsecureSkipVerify: true},
```
jgwest: For E2E tests, I would expect that redis would always have TLS enabled. So is it necessary to enable InsecureSkipVerify, here?

test/e2e/fixture/argoclient.go
```diff
@@ -278,6 +279,8 @@ func (c *ArgoRestClient) GetApplicationLogs(app *v1alpha1.Application, namespace
 	if tailLines > 0 {
 		q.Set("tailLines", fmt.Sprint(tailLines))
 	}
+	// Set sinceSeconds to avoid time parsing errors - fetch logs from last 5 minutes
+	q.Set("sinceSeconds", "300")
```
jgwest: Is this related to this PR?

test/e2e/fixture/cluster.go
```diff
@@ -167,23 +180,142 @@ func getCacheInstance(source string, clusterDetails *ClusterDetails) *appstateca
...
 	redisClient := redis.NewClient(redisOptions)
 	cache := appstatecache.NewCache(cacheutil.NewCache(
 		cacheutil.NewRedisCache(redisClient, 0, cacheutil.RedisCompressionGZip)), 0)
 
 	return cache
 }
 
+// cachedRedisClients stores Redis clients to prevent connection leaks
+var (
+	cachedRedisClients     = make(map[string]*appstatecache.Cache)
```
jgwest: Is it necessary to cache the redis instance? Since this is only from E2E tests, I can't imagine there is much of a performance impact. Can you provide more info on this change?

test/e2e/fixture/cluster.go
```diff
@@ -223,9 +355,29 @@ func getManagedAgentRedisConfig(ctx context.Context, managedAgentClient KubeClie
...
 	if redisAddr == "" {
-		return fmt.Errorf("could not get Redis server address from LoadBalancer ingress")
+		return fmt.Errorf("could not get Redis server address from LoadBalancer ingress, spec, or ClusterIP")
+	}
+
+	// Redis TLS is always enabled for E2E tests
+	clusterDetails.ManagedAgentRedisTLSEnabled = true
+
+	// Set CA certificate path (same as used by agents)
+	clusterDetails.ManagedAgentRedisTLSCAPath = "hack/dev-env/creds/redis-tls/ca.crt"
```
jgwest: Is there a better way of getting this value instead of hardcoding the path? I notice in the codebase we have not previously needed to hardcode the path.

test/e2e/fixture/cluster.go
```diff
@@ -223,9 +355,29 @@ func getManagedAgentRedisConfig(ctx context.Context, managedAgentClient KubeClie
...
+
+	// Allow override via environment variable for local development with port-forward
+	if envAddr := os.Getenv("MANAGED_AGENT_REDIS_ADDR"); envAddr != "" {
```
jgwest: Should this be higher up in the method, e.g. before the `redisAddr == ""` check?

test/e2e/fixture/fixture.go
```diff
@@ -229,14 +229,15 @@ func CleanUp(ctx context.Context, principalClient KubeClient, managedAgentClient
 
 		err = EnsureDeletion(ctx, autonomousAgentClient, &app)
 		if err != nil {
-			return err
+			fmt.Printf("Warning: Failed to delete Application %s from autonomous agent: %v\n", app.Name, err)
```
jgwest: In the changes to this file you have removed a bunch of error checking: converting `return err` to `fmt.Printf`. What is the purpose of this?

test/e2e/fixture/toxyproxy.go
```diff
@@ -116,14 +116,21 @@ func CheckReadiness(t require.TestingT, compName string) {
 		healthzAddr = "http://localhost:8003/healthz"
 	}
 
+	// Use a longer timeout for the principal since it needs to wait for informers to sync
+	// The principal's informer sync timeout is 120s in E2E tests, so we need to wait longer
+	timeout := 120 * time.Second
+	if compName == "principal" {
+		timeout = 180 * time.Second
+	}
+
```
jgwest: We can just change it to 180, we don't need to differentiate  between principal here.

test/e2e/redis_proxy_test.go
```diff
@@ -176,42 +181,65 @@ func (suite *RedisProxyTestSuite) Test_RedisProxy_ManagedAgent_Argo() {
...
 
-	}, time.Second*30, time.Second*5)
+	}, time.Second*60, time.Second*5)
 
 	// Verify the name of the new pod exists in what has been sent from the channel (this will only be true if redis proxy subscription is working)
 	requires.Eventually(func() bool {
-		for {
-			// drain channel looking for name of new pod
+		// Drain available messages in the channel (bounded to prevent hanging on noisy streams)
+		const maxMessagesPerIteration = 200 // Limit processing to prevent infinite loop on continuous SSE streams
```
jgwest: Did this 'infinite loop on continous SSE stream' actually happen?

cmd/argocd-agent/principal.go
```diff
@@ -86,6 +86,16 @@ func NewPrincipalRunCommand() *cobra.Command {
...
+		informerSyncTimeout          time.Duration
```
jgwest: Is this related to this PR? If yes, can you provide more info about this change?

hack/dev-env/configure-redis-tls.sh
```diff
@@ -0,0 +1,264 @@
...
+echo "Scaling down ArgoCD components to prevent SSL errors during transition..."
+
+# Save current replica counts for restoration by configure-argocd-redis-tls.sh
+REPO_SERVER_REPLICAS=$(kubectl get deployment argocd-repo-server -n ${NAMESPACE} -o jsonpath='{.spec.replicas}' 2>/dev/null || echo "0")
+CONTROLLER_REPLICAS=$(kubectl get statefulset argocd-application-controller -n ${NAMESPACE} -o jsonpath='{.spec.replicas}' 2>/dev/null || echo "0")
+SERVER_REPLICAS=$(kubectl get deployment argocd-server -n ${NAMESPACE} -o jsonpath='{.spec.replicas}' 2>/dev/null || echo "0")
+
```
jgwest: As above, not 100% sure what the goal of REPLICA env vars is

hack/dev-env/configure-redis-tls.sh
jgwest: `2024` -> `2025` for any new files added in this PR

hack/dev-env/start-agent-managed.sh
```diff
@@ -34,13 +34,57 @@ if [ -f "$E2E_ENV_FILE" ]; then
...
+kubectl --context vcluster-agent-managed -n argocd get secret argocd-agent-ca \
+  -o jsonpath='{.data.ca\.crt}' | base64 -d > "${ROOT_CA_PATH}" || { echo "ERROR: Failed to extract CA cert from argocd-agent-ca secret"; exit 1; }
+echo " mTLS client certificates and CA extracted"
+
 go run github.com/argoproj-labs/argocd-agent/cmd/argocd-agent agent \
     --agent-mode managed \
     --creds "mtls:any" \
+    --tls-client-cert="${TLS_CERT_PATH}" \
+    --tls-client-key="${TLS_KEY_PATH}" \
+    --root-ca-path="${ROOT_CA_PATH}" \
```
jgwest: As above, curious about purpose.

hack/dev-env/start-e2e.sh
```diff
@@ -47,27 +47,75 @@ getExternalLoadBalancerIP() {
...
-K8S_CONTEXT="--context=vcluster-agent-managed"
-K8S_NAMESPACE="-n argocd"
-getExternalLoadBalancerIP "argocd-redis"
-export MANAGED_AGENT_REDIS_ADDR="$EXTERNAL_IP:6379"
+# Dual-mode setup: Port-forwards (local) vs LoadBalancer IPs
+# Can be overridden with E2E_USE_PORT_FORWARD=true/false
+if [[ "${E2E_USE_PORT_FORWARD:-auto}" == "auto" ]]; then
+  # Auto-detect based on OS
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    E2E_USE_PORT_FORWARD=true
```
jgwest: It doesn't seem right to hardcode this to a particular OS. 

Is there another solution we could use? For example, just outputing an error message that suggests that people can enable this ENV AR? This might be useful for non-mac users too (for example linux users who don't want to install microk8s).

hack/dev-env/start-principal.sh
```diff
@@ -21,18 +21,11 @@ if ! kubectl config get-contexts | tail -n +2 | awk '{ print $2 }' | grep -qE '^
...
-       if test "$ipaddr" != ""; then
-               ARGOCD_PRINCIPAL_REDIS_SERVER_ADDRESS=$ipaddr:6379
-       elif test "$hostname" != ""; then
-               ARGOCD_PRINCIPAL_REDIS_SERVER_ADDRESS=$hostname:6379
-       else
-               echo "Could not determine Redis server address." >&2
-               echo "Please set ARGOCD_PRINCIPAL_REDIS_SERVER_ADDRESS manually" >&2
-               exit 1
-       fi
+       # For TLS to work with proper certificate validation in dev/E2E, we expect
```
jgwest: With this change, we are removing logic that would previously look up the loadbalancer IP, and instead assumes that proxy is enabled.

When running via GH E2E PRs (or from local machine), we still want to use the IP address from load balancer. Can you confirm this is still the case?

If this is still the case, then I presume the comment here needs to be updated because it asserts that in dev/E2E, port-forward is always enabled.

install/helm-repo/argocd-agent-agent/values.yaml
```diff
@@ -133,7 +133,34 @@ tlsClientKeyPath: ""
...
+networkPolicy:
+  enabled: true 
+  redis:
+    enabled: true 
+    name: allow-agent-to-redis
+    namespace: "" 
+    redisSelector:
+      app.kubernetes.io/name: argocd-redis
+    agentSelector:
+      app.kubernetes.io/name: argocd-agent-agent
```
jgwest: Are these values used? I couldn't find them reference in the helm resources.

install/kubernetes/principal/principal-params-cm.yaml
```diff
@@ -140,3 +140,27 @@ data:
...
+  # principal.redis.upstream.ca-path: Path to CA certificate for verifying upstream Redis TLS certificate.
+  # This should match the mount path in the Deployment for upstream CA.
+  # Default: "/app/config/redis-upstream-tls/ca.crt"
+  principal.redis.upstream.ca-path: "/app/config/redis-upstream-tls/ca.crt"
+  # principal.redis.upstream.ca-secret-name: Secret name containing CA certificate for verifying upstream Redis TLS certificate.
+  # Default: "argocd-redis-tls"
+  principal.redis.upstream.ca-secret-name: "argocd-redis-tls"
+  # principal.redis.upstream.tls.insecure: INSECURE: Do not verify upstream Redis TLS certificate.
+  # Default: false
+  principal.redis.upstream.tls.insecure: "false"
```
jgwest: As above, my vote would be to remove the term `upstream`

principal/options.go
```diff
@@ -77,6 +77,15 @@ type ServerOptions struct {
...
 	informerSyncTimeout    time.Duration
+	// Redis TLS configuration
+	redisTLSEnabled          bool
+	redisServerTLSCert       *x509.Certificate
+	redisServerTLSKey        crypto.PrivateKey
+	redisServerTLSCertPath   string
+	redisServerTLSKeyPath    string
+	redisUpstreamTLSCA       *x509.CertPool
+	redisUpstreamTLSCAPath   string
+	redisUpstreamTLSInsecure bool
```
jgwest: As above, `upstream`

principal/redisproxy/redisproxy.go
```diff
@@ -57,6 +61,18 @@ type RedisProxy struct {
...
+	tlsEnabled        bool
+	tlsServerCert     *x509.Certificate
+	tlsServerKey      crypto.PrivateKey
+	tlsServerCertPath string
+	tlsServerKeyPath  string
+
+	// TLS configuration for upstream Redis (connections to principal's argocd-redis)
+	upstreamTLSCA       *x509.CertPool
+	upstreamTLSCAPath   string
+	upstreamTLSInsecure bool
```
jgwest: As above `upstream`

test/e2e/fixture/cluster.go
```diff
@@ -167,23 +180,142 @@ func getCacheInstance(source string, clusterDetails *ClusterDetails) *appstateca
...
+				MinVersion: tls.VersionTLS12,
+			}
+
+			// Load CA certificate for proper verification if available
+			if clusterDetails.PrincipalRedisTLSCAPath != "" {
+				// Check if CA cert file exists
+				if _, err := os.Stat(clusterDetails.PrincipalRedisTLSCAPath); err == nil {
+					caCertPEM, err := os.ReadFile(clusterDetails.PrincipalRedisTLSCAPath)
+					if err != nil {
+						panic(fmt.Sprintf("failed to read principal Redis CA certificate: %v", err))
```
jgwest: `Fail()` would be Ginkgo alternative to panic, can replace here and other uses.

test/e2e/redis_proxy_test.go
```diff
@@ -117,6 +117,11 @@ func (suite *RedisProxyTestSuite) Test_RedisProxy_ManagedAgent_Argo() {
 
 	requires.NotNil(msgChan)
 
+	// Wait for SSE stream to fully establish and Redis SUBSCRIBE to propagate
+	// This prevents a race condition where the pod is deleted before the subscription is active
+	t.Log("Waiting for SSE stream to fully establish...")
+	time.Sleep(5 * time.Second)
```
jgwest: Sleep statements in E2E test code are nearly always bad, and can nearly be always replaced with a better alternative. If you attempt to replace these, and if not able, document what you tried?

test/e2e/redis_proxy_test.go
Rizwana777: The "infinite loop" concern isn't really an issue:
The select with default makes the loop non-blocking
When the channel is empty, we return immediately and wait 5 seconds before the next iteration
The Eventually timeout (120s) prevents the test from hanging forever

hack/dev-env/Procfile.e2e
Rizwana777: The port-forwards are required for "local development only" (macOS host running `go run`).

These were originally in `Procfile.e2e` but have been moved to `Procfile.e2e.local` to make it clear they’re specifically for local development, not CI.

Why Port-Forwards Are Needed:

When running agents/principal locally via `go run`:
1. The processes run on the “host machine” (outside the cluster)
2. They cannot access in-cluster DNS names like `argocd-redis.argocd.svc`
3. Port-forward exposes the in-cluster Redis instances to `localhost` on the host

The Setup in `Procfile.e2e.local`:
-  pf-control-plane: Exposes control-plane Redis to `localhost:6380`
-  pf-managed: Exposes managed agent Redis to `localhost:6381`
-  pf-autonomous: Exposes autonomous agent Redis to `localhost:6382`

This allows the local `go run` processes to connect to Redis via localhost addresses.

In Production:
Real Kubernetes/Helm deployments don’t need this - agents and principal run in-cluster and connect directly to Redis service DNS.

docs/getting-started/kubernetes/index.md
Rizwana777: I have updated it to reuse the existing agent CA. gen-redis-tls-certs.sh now uses the agent CA from hack/dev-env/creds/ or the argocd-agent-ca secret instead of generating a separate one

hack/dev-env/configure-redis-tls.sh
Rizwana777: Should I wait for #676 to merge first before integrating Redis TLS into argocd-agentctl? 

My thinking is:
- #676 adds external certificate support to argocd-agentctl
- I could reuse/extend those same functions for Redis TLS certificate handling
- This would ensure a consistent approach between agent mTLS and Redis TLS in the CLI

Options:
1.  Wait for #676 → Then integrate Redis TLS into argocd-agentctl using the new certificate functions
2. Merge this PR with shell scripts → Create follow-up to integrate into argocd-agentctl after #676
3. Integrate now without #676 → Add Redis TLS to argocd-agentctl immediately

Which approach would you prefer?

hack/dev-env/start-agent-autonomous.sh
Rizwana777: Just to make sure I understand correctly: are you referring to the mTLS certificate extraction that was in the original version?

If so, I have removed that! You’re absolutely right that was unnecessary complexity. The mTLS certificates are loaded from Kubernetes secrets in production.

What I removed:
- All the `kubectl get secret argocd-agent-tls` extraction code
- The `--tls-cert-path` and `--tls-key-path` flags in the dev scripts

What remains:
-  Redis TLS uses the same agent CA (consolidated based on your other comment about reusing the agent CA)
-  For local dev, the script extracts the agent CA from the Kubernetes secret if needed

Why the agent CA extraction is still there:
These dev scripts run agents/principal on the host (via `go run`) with port-forward to Redis. The local process needs the CA file to verify the Redis TLS connection since it can’t access Kubernetes secrets directly.

The script now:
1. Checks if `creds/ca.{crt,key}` exist on disk
2. If not -  extracts them from `argocd-agent-ca` Kubernetes secret
3. Uses that CA to sign all Redis TLS certificates

 Let me know if I’m still missing something or this is not a correct approach?

hack/dev-env/start-agent-autonomous.sh
Rizwana777: This is specifically for local development when running agents on the host machine (via go run), not in-cluster deployment.
Here’s why we use localhost:
- Redis runs in the vcluster (in-cluster)
- But the agent process runs on the host machine (outside the cluster) via go run
- Host machine cannot resolve in-cluster DNS names like “argocd-redis.argocd.svc”

So, We use `kubectl port-forward` to expose the in-cluster Redis to localhost:
kubectl port-forward svc/argocd-redis -n argocd 6381:6379 --context vcluster-agent-managed. This makes the in-cluster Redis accessible at `localhost:6381` from the host machine.

In real Kubernetes/Helm deployments, agents run in-cluster and use the service DNS name (`argocd-redis:6379`) directly - no localhost, no port-forward needed.

This script is purely for local E2E testing convenience. Please correct me if this approach is incorrect

principal/tracker/tracking.go
Rizwana777: You’re right, this change is not related to Redis TLS.
This was a concurrency fix (buffered channel to prevent potential deadlocks) that  got included in this PR. I’ve reverted it to keep the PR focused solely on Redis TLS functionality.
If the deadlock issue needs to be addressed, I’ll file it as a separate issue with proper reproduction steps and handle it.

principal/resource.go
Rizwana777: this change is not related to Redis TLS.
I’ve tested locally and confirmed that tests pass fine with the original 10s timeout. This was included in the PR prematurely - I thought it might be needed but it turns out it’s not.
I’ve reverted this back to 10s to keep the PR focused on Redis TLS changes only.
Thank you for pointing this out

agent/agent.go
Rizwana777: Changed to return an error instead of falling back to system CAs.

cmd/argocd-agent/agent.go
Rizwana777: Changed to return an error instead of falling back to system CAs.

hack/dev-env/configure-argocd-redis-tls.sh
Rizwana777: You're right, replicas are always 1 

I will simplify this to just use `--replicas=1`

internal/argocd/cluster/cluster.go
Rizwana777: this change is not related to Redis TLS. I've removed it.

This was an unrelated improvement to connection state initialization that accidentally got included in this PR.

test/e2e/redis_proxy_test.go
Rizwana777: this is for the Argo CD API HTTPS connection. we have two options:

1. Keep `InsecureSkipVerify: true`(current) - Simple, works for test environment
2. Load Argo CD server CA - Proper TLS verification but more complex setup
Happy to to change to (2) if you prefer proper certificate validation

Redis TLS is ALWAYS enabled for E2E tests - configured in `test/e2e/fixture/cluster.go:370`:
// Redis TLS is always enabled for E2E tests
clusterDetails.ManagedAgentRedisTLSEnabled = true
clusterDetails.PrincipalRedisTLSEnabled = true The `InsecureSkipVerify: true` on line 653 is for the HTTP client connecting to the Argo CD API server's HTTPS endpoint (which uses self-signed certs in the test environment). Without it, tests would fail with "x509: certificate signed by unknown authority".

test/e2e/fixture/argoclient.go
Rizwana777: You are right, it's not directly related to Redis TLS functionality. I've removed the sinceSeconds=300 change - it was a test stabilization attempt but isn't necessary for the Redis TLS feature itself.

test/e2e/fixture/cluster.go
Rizwana777: You’re right. I have removed the unnecessary caching complexity.
The caching was intended to prevent connection leaks when multiple E2E tests access Redis. However, this concern was misplaced.
Now -
The Redis clients will now be created as needed, and the underlying go-redis connection pool will handle reuse efficiently.

test/e2e/fixture/cluster.go
Rizwana777: I made the Redis TLS CA path configurable via REDIS_TLS_CA_PATH environment variable. run-e2e.sh sets it to an absolute path, and cluster.go reads it with a relative path fallback. This avoids hardcoding while allowing overrides for different environments.

test/e2e/fixture/cluster.go
Rizwana777: Fixed. Moved the env var check before the error check in both `getManagedAgentRedisConfig()` and `getPrincipalRedisConfig()`.

test/e2e/fixture/fixture.go
Rizwana777: This changes the cleanup behavior from "fail fast" to "best effort" - log warnings and continue cleaning up even if some deletions fail.

Purpose: Make test cleanup more resilient, especially useful when port-forwards or Redis connections die during cleanup.

I reverted the changes now, I will raise a seperate PR for this if this change is required and valid

cmd/argocd-agent/principal.go
Rizwana777: This change is related to Redis TLS and was needed to fix e2e test failures.
The Problem:
When Redis TLS was enabled, E2E tests started failing because the principal couldn’t sync Kubernetes informers within the default 60s timeout. The additional TLS handshake overhead and slower startup caused informer sync to take longer.
Evidence from Test Failures:
In commit bf0d4f8 (before the timeout fix), tests consistently failed with:
Error Trace: toxyproxy.go:119 (CheckReadiness)
Error: Condition never satisfied
Test: TestResyncTestSuite/Test_ResyncUpdatesOnPrincipalStartupAutonomous
Duration: 121 seconds (timed out waiting for principal to be ready)
The principal never became healthy because informers couldn’t sync in time.
The Fix:
Three coordinated changes were needed:
cmd/argocd-agent/principal.go (lines 259-261, 434-436):
Added --informer-sync-timeout flag with clear documentation
Default remains 60s (0 = internal default)
Can be overridden for slow environments
hack/dev-env/start-principal.sh (line 43):
export ARGOCD_PRINCIPAL_INFORMER_SYNC_TIMEOUT=120s # E2E needs longer timeout
test/e2e/fixture/toxyproxy.go (line 121):
timeout := 180 * time.Second // Test must wait longer than informer timeout
Why These Values:
60s = Default for production (no Redis TLS overhead)
120s = E2E tests with Redis TLS (2x default for safety)
180s = Test timeout (120s informer + 60s buffer)
After these changes, tests pass reliably in both local and CI environments. The flag documentation makes it clear that users can adjust this for their environment if needed.

hack/dev-env/start-principal.sh
Rizwana777: The logic is working correctly  and loadbalancer IP lookup is working correctly and but the comment was misleading, I have updated this

install/helm-repo/argocd-agent-agent/values.yaml
Rizwana777: These `networkPolicy` values are not used anywhere in the Helm chart. Removed it

hack/dev-env/start-agent-autonomous.sh
Rizwana777: https://github.com/argoproj-labs/argocd-agent/pull/664#discussion_r2642619699 removed this based on this comment

hack/dev-env/configure-redis-tls.sh
Rizwana777: removed it

principal/tracker/tracking.go
Rizwana777: https://github.com/argoproj-labs/argocd-agent/pull/664#discussion_r2642622024 based on this comment removed it

test/e2e/redis_proxy_test.go
Rizwana777: I kept the time.Sleep(5 * time.Second) in SSE stream tests because replacing it seemed problematic - polling would consume messages from the channel that the test needs later. I added a comment explaining this: "We use sleep here because polling would consume messages from the channel that the test logic below needs to process." Does this reasoning make sense, or is there a better approach I'm missing? The other sleeps are in standard polling loops with timeouts.

test/e2e/redis_proxy_test.go
Rizwana777: I kept the time.Sleep(5 * time.Second) in SSE stream tests because replacing it seemed problematic - polling would consume messages from the channel that the test needs later. I added a comment explaining this: "We use sleep here because polling would consume messages from the channel that the test logic below needs to process." Does this reasoning make sense, or is there a better approach I'm missing? The other sleeps are in standard polling loops with timeouts.

test/e2e/redis_proxy_test.go
Rizwana777: You are right removed the buffer size . Tests pass without it. Thank you for pointing this out

test/e2e/fixture/cluster.go
Rizwana777: I have updated the code to use Ginkgo's Fail() instead of panic().
Quick question though - since we're using testify/suite rather than Ginkgo's style tests, is there a specific reason to prefer Fail() over panic() here?

agent/agent.go
Rizwana777: As I have updated the code to return an error if WithCacheRefreshInterval didn't set the value , I have modified few of the unit tests to set the value

Rizwana777:
And also I think test server certificate has expired 
=== RUN   Test_TransportFromConfig/Valid_client_certificate_data_in_file
    kubernetes_test.go:238: 
        	Error Trace:	/home/runner/work/argocd-agent/argocd-agent/internal/tlsutil/kubernetes_test.go:238
        	Error:      	Received unexpected error:
        	            	server certificate has expired on Wed, 17 Dec 2025 14:19:08 +0000
        	Test:       	Test_TransportFromConfig/Valid_client_certificate_data_in_file

agent/agent.go
Rizwana777: This isn't related to Redis TLS. I added this while troubleshooting e2e failures and forgot to remove it. I'll revert this change.

hack/dev-env/Procfile.e2e
jannfis: We are already testing Redis connectivity as part of the redis-proxy functionality using a Kubernetes service that can be accessed from outside of the cluster. Why can't we reuse that?

hack/dev-env/configure-redis-tls.sh
jannfis: #676 is merged. It would be great to generate (or reuse) Redis TLS certificates using the argocd-agentctl instead of using the shell scripts. This way, it is platform agnostic (sometimes there are subtle differences between Mac and Linux, for example), it can be used by people in their own development environments (using argocd-agentctl PKI) or in production (external managed CA, but configuration/integration using argocd-agentctl).

hack/dev-env/start-agent-autonomous.sh
jannfis: As mentioned earlier, the redis-proxy functionality should already have all the mechanisms in place to access Redis without port forwarding and dirty hacks.

test/e2e/fixture/cluster.go
jgwest: Sorry, this was a mistake on my part, I am reviewing other PRs in other repos that use Ginkgo. We should not use Ginkgo in this repo :sweat_smile:. Feel free to revert any Ginkgo changes you added.

test/e2e/redis_proxy_test.go
jgwest: Rather than using time.Sleep, you can instead wait for the old pod to appear in the resource tree.
- It will only appear in the resource tree if redis is fully available.

You can do that, for example, by converting this to a `requires.Eventually`:
```go
	// Ensure that the pod appears in the resource tree value returned by Argo CD server (this will only be true if redis proxy is working)
	tree, err := appClient.ResourceTree(suite.Ctx, &application.ResourcesQuery{
		ApplicationName: &appOnPrincipal.Name,
		AppNamespace:    &appOnPrincipal.Namespace,
		Project:         &appOnPrincipal.Spec.Project,
	})
	requires.NoError(err)

	requires.NotNil(tree)

	matchFound := false
	for _, node := range tree.Nodes {
		if node.Kind == "Pod" && node.Name == oldPod.Name {
			matchFound = true
			break
		}
	}
	requires.True(matchFound)
```

test/e2e/redis_proxy_test.go
jgwest: Same proposed solution as above re: using requires.Eventually

cmd/ctl/redis.go
```diff
@@ -0,0 +1,563 @@
...
+			}
+
+			var certData, keyData, caData string
+
+			if certFromSecret != "" && caFromSecret != "" {
+				// Use external certificates from existing secrets
+				certData, keyData, caData, err = readTLSFromExistingSecrets(ctx, clt, certFromSecret, caFromSecret)
+				if err != nil {
+					cmdutil.Fatal("%v", err)
+				}
```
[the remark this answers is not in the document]
Rizwana777: This function is present here https://github.com/argoproj-labs/argocd-agent/blob/main/cmd/ctl/agent.go#L118

test/e2e/README.md
```diff
@@ -24,6 +24,16 @@ To setup the test environment on the cluster, execute the following command from
 make setup-e2e
 ```
 
+**Note:** Redis proxy and resource proxy tests require a reverse tunnel because vcluster pods cannot reach the host machine's IP address where the principal's Redis Proxy runs. The tunnel bridges this gap by routing traffic from the vcluster to your local machine.
```
Rizwana777, line 27: This I observed during local testing, please let me know if this is valid

test/e2e/redis_proxy_test.go
Rizwana777: This test was flaky in my local and also got fixed after this change

cmd/argocd-agent/principal.go
```diff
@@ -265,6 +273,49 @@ func NewPrincipalRunCommand() *cobra.Command {
...
+				if redisTLSCAPath != "" {
+					modesSet++
+				}
+				// Only count non-default secret name to allow default value
+				if redisTLSCASecretName != "" && redisTLSCASecretName != "argocd-redis-tls" {
+					modesSet++
+				}
+				if modesSet > 1 {
+					cmdutil.Fatal("Only one Redis TLS mode can be specified: --redis-tls-insecure, --redis-ca-path, or --redis-ca-secret-name")
+				}
```
jgwest, line 404: @Rizwana777 can you confirm if this one is correct?

cmd/ctl/redis.go
```diff
@@ -0,0 +1,568 @@
...
+			if err != nil {
+				cmdutil.Fatal("Could not create/update secret: %v", err)
+			}
+
+			fmt.Printf("Secret %s created in namespace %s\n", secretName, globalOpts.principalNamespace)
+			fmt.Printf("\nNext: argocd-agentctl redis configure-tls --principal-context %s --principal-namespace %s\n", globalOpts.principalContext, globalOpts.principalNamespace)
+		},
+	}
+	command.Flags().StringSliceVar(&ips, "ip", []string{"127.0.0.1"}, "IP addresses for the certificate's SAN")
+	command.Flags().StringSliceVar(&dns, "dns", []string{"localhost", "argocd-redis", "argocd-redis.argocd.svc.cluster.local"}, "DNS names for the certificate's SAN")
```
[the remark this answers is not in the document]
jgwest: @Rizwana777 can you verify if this is correct?

install/helm-repo/argocd-agent-agent/README.md
```diff
@@ -82,7 +87,7 @@ Kubernetes: `>=1.24.0-0`
 | tlsClientCertPath | string | `""` | Path to the TLS client certificate. |
 | tlsClientInSecure | string | `"false"` | Whether to skip TLS verification for client connections. |
 | tlsClientKeyPath | string | `""` | Path to the TLS client key. |
-| tlsRootCAPath | string | `""` | Path to the TLS root CA certificate. |
+| tlsRootCAPath | string | `"/app/config/tls/ca.crt"` | Path to the TLS root CA certificate. |
```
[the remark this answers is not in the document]
jgwest: @Rizwana777 is this correct?

install/kubernetes/principal/principal-params-cm.yaml
```diff
@@ -140,3 +140,27 @@ data:
...
+  # principal.redis.proxy.server.tls.secret-name: Secret name containing TLS certificate and key for Redis proxy server.
+  # Default: "argocd-redis-tls"
+  principal.redis.proxy.server.tls.secret-name: "argocd-redis-tls"
+  # principal.redis.ca-path: Path to CA certificate for verifying Redis TLS certificate.
+  # This should match the mount path in the Deployment for Redis CA.
+  # Default: "/app/config/redis-tls/ca.crt"
+  principal.redis.ca-path: "/app/config/redis-tls/ca.crt"
+  # principal.redis.ca-secret-name: Secret name containing CA certificate for verifying Redis TLS certificate.
+  # Default: "argocd-redis-tls"
+  principal.redis.ca-secret-name: "argocd-redis-tls"
```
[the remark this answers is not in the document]
jgwest: @Rizwana777 can you verify?

principal/options.go
```diff
@@ -479,3 +488,61 @@ func WithHealthzPort(port int) ServerOption {
...
+	return func(o *Server) error {
+		c, err := tlsutil.TLSCertFromSecret(context.Background(), kube, namespace, name)
+		if err != nil {
+			return err
+		}
+		o.options.redisProxyServerTLSCert = c.Leaf
+		o.options.redisProxyServerTLSKey = c.PrivateKey
+		return nil
+	}
+}
```
jgwest, line 660: @Rizwana777 can you verify?

principal/server.go
```diff
@@ -378,7 +405,45 @@ func NewServer(ctx context.Context, kubeClient *kube.KubernetesClient, namespace
...
+			if !caCertPool.AppendCertsFromPEM(caCert) {
+				return nil, fmt.Errorf("failed to parse Redis CA certificate from %s", s.options.redisTLSCAPath)
+			}
+			clusterMgrRedisTLSConfig.RootCAs = caCertPool
+			log().WithField("caPath", s.options.redisTLSCAPath).Info("Loaded Redis CA certificate for cluster manager")
+		} else {
+			// No CA specified - require explicit configuration
+			return nil, fmt.Errorf("redis TLS enabled but no CA certificate configured for cluster manager: use --redis-ca-path, --redis-ca-secret-name, or --redis-tls-insecure")
+		}
+	}
```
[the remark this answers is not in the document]
jgwest: @Rizwana777 is this do-able?

cmd/ctl/redis.go
```diff
@@ -0,0 +1,568 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
+)
+
+const (
+	redisTLSCertPath     = "/app/tls"
+	argoCDRedisTLSCAPath = "/app/config/redis-tls"
+)
+
+func NewRedisCommand() *cobra.Command {
```
jgwest: I'm not 100% sure I understand the purpose of this, so I will defer to @jannfis on what he wants to do here.

install/kubernetes/principal/principal-params-cm.yaml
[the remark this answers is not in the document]
Rizwana777: verified this, all the naming conventions are correct

principal/options.go
Rizwana777: Updated this to used similar pattern as in other files to be consistent , but c.Leaf also works fine with no issues

install/helm-repo/argocd-agent-agent/README.md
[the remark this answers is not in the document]
Rizwana777: earlier I was using CA path to extract the tls certs, now extracting through secret, so this is not needed, fixed it

cmd/argocd-agent/principal.go
Rizwana777: yes this is correct , whenever use specify 2 flags(--redis-tls-insecure and  --redis-ca-secret-name=argocd-redis-tls ) like this , it will ingore the secret and --redis-tls-insecure mode will run
go run github.com/argoproj-labs/argocd-agent/cmd/argocd-agent principal \
	--allowed-namespaces '*' \
	--kubecontext vcluster-control-plane \
	--log-level ${ARGOCD_AGENT_LOG_LEVEL:-trace} \
	--namespace argocd \
	--auth "mtls:CN=([^,]+)" \
        --redis-tls-insecure \
         --redis-ca-secret-name=argocd-redis-tls
	$ARGS
Fixed it now to use only one mode

cmd/ctl/redis.go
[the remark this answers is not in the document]
Rizwana777: verified this, this is correct  I have made the changes

./dist/argocd-agentctl redis generate-certs \
  --principal-context vcluster-control-plane \
  --principal-namespace <namespace> \. --- > use argocd
  --upsert
Secret argocd-redis-tls created in namespace argocd

then
argocd-agentctl redis configure-tls --principal-context vcluster-control-plane --principal-namespace argocd

rnaaz@rnaaz-mac argocd-agent % kubectl --context vcluster-control-plane -n argocd get secret argocd-redis-tls -o jsonpath='{.data.tls\.crt}' | \
  base64 -d | \
  openssl x509 -noout -text | \
  grep -A5 "Subject Alternative Name"
            X509v3 Subject Alternative Name: 
                DNS:localhost, DNS:argocd-redis, DNS:argocd-redis.argocd.svc.cluster.local, IP Address:127.0.0.1

principal/server.go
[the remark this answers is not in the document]
Rizwana777: moved this to the common function called CreateRedisTLSConfig

hack/dev-env/start-agent-autonomous.sh
Rizwana777: Removed all unnecessary changes

cmd/ctl/redis.go
Rizwana777: https://github.com/argoproj-labs/argocd-agent/pull/664#discussion_r2662325628 As per Jann's comment I am using argocd-agentctl to generate certs, configure-tls  and configure-argocd  , now this will work like this -

Generate Redis TLS Certificates
argocd-agentctl redis generate-certs \
  --principal-context <context> \
  --principal-namespace argocd

Configure Redis Deployment for TLS
argocd-agentctl redis configure-tls \
  --principal-context <context> \
  --principal-namespace argocd

Configure Argo CD for Redis TLS
argocd-agentctl redis configure-argocd \
  --principal-context <context> \
  --principal-namespace argocd

earlier I was using scripts to do this

I am open for suggestions and am happy to make the changes

cmd/ctl/redis.go
```diff
@@ -0,0 +1,583 @@
...
+		},
+		GroupID: "config",
+	}
+	command.AddCommand(NewRedisGenerateCertsCommand())
+	command.AddCommand(NewRedisConfigureTLSCommand())
+	command.AddCommand(NewRedisConfigureArgoCDCommand())
+	return command
+}
+
+func NewRedisGenerateCertsCommand() *cobra.Command {
```
jannfis: I think this command should really go to `pki issue redis` instead. 

This would make it also clear that, for production purposes, people should get the certificates from elsewhere.

Rizwana777:
> Thanks for all the effort you put into this PR, @Rizwana777 !
> 
> I have some concerns about the complexity around the integration part, so I would propose to remove all the code regarding integration from this PR (i.e. the shell scripts, the changes to argocd-agentctl, etc).
> 
> We can then merge the PR, and follow up on how the integration will look like and how deep we actually want to have the integration go.
> 
> WDYT?

@jannfis Thanks for the feedback. I also somewhat felt it was getting complex while working on it, I will remove the integration part for now and later work on this 

cc: @jgwest

principal/redisproxy/redisproxy.go
```diff
@@ -106,8 +201,6 @@ func (rp *RedisProxy) Start() error {
 		}
 	}()
 
-	rp.log().Infof("Redis proxy started on %s", rp.listenAddress)
```
jannfis, line 133: Instead of removing the statement here, and moving it to more "specialized" statements above, how about instead:

```go
rp.log().WithField("tls", rp.tlsEnabled).Infof("Redis proxy started on %s", rp.listenAddress)
```

principal/redisproxy/redisproxy.go
```diff
@@ -85,13 +101,92 @@ func New(listenAddress string, principalRedisAddress string, sendSyncMessageToAg
...
+			return err
+		}
+		rp.log().Infof("Redis proxy started on %s with TLS", rp.listenAddress)
+	} else {
+		l, err = net.Listen("tcp", rp.listenAddress)
+		if err != nil {
+			rp.log().WithError(err).Error("error occurred on listening to addr: " + rp.listenAddress)
+			return err
+		}
+		rp.log().Infof("Redis proxy started on %s without TLS", rp.listenAddress)
```
jannfis: Together with the suggestion a bit below, this can be simplified as:

```suggestion
	if rp.tlsEnabled {
		tlsConfig, err := rp.createServerTLSConfig()
		if err != nil {
			rp.log().WithError(err).Error("error creating TLS config for Redis proxy server")
			return err
		}
		l, err = tls.Listen("tcp", rp.listenAddress, tlsConfig)
	} else {
		l, err = net.Listen("tcp", rp.listenAddress)
	}
	if err != nil {
	    rp.log().WithError(err).WithField("tls", rp.tlsEnabled).Error("Unable to listen on " + rp.listenAddress)
	}
```

principal/redisproxy/redisproxy.go
```diff
@@ -724,24 +817,99 @@ func (are *argoCDRedisWriterInternal) writeToArgoCDRedisSocket(logCtx *logrus.En
...
+		logCtx.WithError(err).WithField("redisAddress", rp.principalRedisAddress).Error("Resolution error")
 		return nil, fmt.Errorf("unable to resolve address: %w", err)
 	}
 
-	// Dial the resolved address
-	redisConn, err = net.DialTCP("tcp", nil, addr)
+	// Dial the resolved address with timeout to prevent indefinite hangs
+	dialer := &net.Dialer{
+		Timeout: 30 * time.Second,
+	}
```
jannfis, line 863: Does adding the timeout here have a specific reason?

principal/redisproxy/redisproxy.go
```diff
@@ -724,24 +817,99 @@ func (are *argoCDRedisWriterInternal) writeToArgoCDRedisSocket(logCtx *logrus.En
...
+	hasUpstreamTLSConfig := rp.redisTLSCA != nil || rp.redisTLSCAPath != "" || rp.redisTLSInsecure
+
+	// Fail if server TLS is enabled but upstream TLS is not configured
+	// This prevents a security gap: agent→proxy is encrypted, but proxy→redis is not
+	if rp.tlsEnabled && !hasUpstreamTLSConfig {
+		conn.Close()
+		return nil, fmt.Errorf("SECURITY ERROR: Redis proxy server has TLS enabled, but no upstream TLS configuration provided. " +
+			"This would create an unencrypted connection to principal Redis. " +
+			"Configure upstream TLS using --redis-ca-path, --redis-ca-secret-name, or --redis-tls-insecure")
+	}
```
jannfis: IMHO, it should not fail at this point. Please print a warning instead and continue.

principal/redisproxy/redisproxy.go
```diff
@@ -724,24 +817,99 @@ func (are *argoCDRedisWriterInternal) writeToArgoCDRedisSocket(logCtx *logrus.En
...
+
+	// If upstream TLS is configured, wrap the connection with TLS
+	// This is independent of server TLS configuration
+	if hasUpstreamTLSConfig {
+		tlsConfig := &tls.Config{
+			MinVersion: tls.VersionTLS12,
+		}
```
jannfis: We should not hardcode the TLS client configuration wrt versions or ciphers for now.

It is safe to use Go defaults for the client.

principal/redisproxy/redisproxy.go
```diff
@@ -724,24 +817,99 @@ func (are *argoCDRedisWriterInternal) writeToArgoCDRedisSocket(logCtx *logrus.En
...
+
+		if rp.redisTLSInsecure {
+			logCtx.Warn("INSECURE: Not verifying upstream Redis TLS certificate")
+			tlsConfig.InsecureSkipVerify = true
+			// Warn if CA configuration is provided but will be ignored
+			if rp.redisTLSCA != nil || rp.redisTLSCAPath != "" {
+				logCtx.Warn("CA configuration provided but ignored due to InsecureSkipVerify=true")
+			}
```
jannfis: I think this warning is superfluous at this point.

internal/tlsutil/redis.go
```diff
@@ -0,0 +1,93 @@
...
+//   - redisAddress: Redis server address (used to extract ServerName for TLS)
+//   - insecure: Whether to skip TLS verification (insecure mode)
+//   - caPool: Pre-loaded CA certificate pool (from Kubernetes secret)
+//   - caPath: Path to CA certificate file
+//   - componentName: Name of the component using Redis (for logging, e.g., "cluster manager", "cluster cache")
+//
+// Returns:
+//   - *tls.Config: Configured TLS config, or nil if TLS is not enabled
+//   - error: Error if TLS is enabled but configuration is invalid
+func CreateRedisTLSConfig(
```
jannfis: This code doesn't belong at all into package `tlsutil`, which provides low-level functions to create TLS artifacts (certificates, cert pools, etc) from input sources. The functions in this package are not supposed to be specific to any other package.

Also, it duplicates other code in this package.

principal/redisproxy/redisproxy.go
```diff
@@ -85,13 +101,92 @@ func New(listenAddress string, principalRedisAddress string, sendSyncMessageToAg
...
+		cert.Certificate = [][]byte{rp.tlsServerCert.Raw}
+		cert.PrivateKey = rp.tlsServerKey
+		cert.Leaf = rp.tlsServerCert
+	} else {
+		return nil, fmt.Errorf("no TLS certificate configured")
+	}
+
+	return &tls.Config{
+		Certificates: []tls.Certificate{cert},
+		MinVersion:   tls.VersionTLS12,
```
jannfis: Please do not hardcode TLS configuration for the server. We do have global TLS server configuration, which should also be used for the Redis proxy's TLS server.

Currently, they are available from the principal's server options: https://github.com/argoproj-labs/argocd-agent/blob/9172314358f232cf57168c2da8896f2c886062aa/principal/options.go#L50-L52

principal/redisproxy/redisproxy.go
```diff
@@ -724,24 +817,99 @@ func (are *argoCDRedisWriterInternal) writeToArgoCDRedisSocket(logCtx *logrus.En
...
+				conn.Close()
+				return nil, fmt.Errorf("failed to read CA certificate: %w", err)
+			}
+			caCertPool := x509.NewCertPool()
+			if !caCertPool.AppendCertsFromPEM(caCert) {
+				conn.Close()
+				return nil, fmt.Errorf("failed to append CA certificate")
+			}
+			tlsConfig.RootCAs = caCertPool
+			logCtx.Debugf("Using CA certificate from %s for upstream Redis TLS", rp.redisTLSCAPath)
```
jannfis: https://github.com/argoproj-labs/argocd-agent/blob/9172314358f232cf57168c2da8896f2c886062aa/internal/tlsutil/tlsutil.go#L91

cmd/ctl/redis.go
Rizwana777: Remove this while removing integration part, I will make a note while adding integration

principal/redisproxy/redisproxy.go
Rizwana777: make sense, thanks

principal/redisproxy/redisproxy.go
Rizwana777: yes, it prevents redis connection dialer to prevent indefinite blocking when redis is unreachable. Timeout make sures failed connections fails fast instead of hanging forever , open for suggestions on timeout value

internal/tlsutil/redis.go
Rizwana777: make sense, remove and updated code

principal/redisproxy/redisproxy.go
Rizwana777: removed duplication and used X509CertPoolFromFile

PR #676, opened 2025-12-05
feat(cli): agent create can now use existing TLS certificates

### LINKED ISSUE #622: argocd-agentctl create Agent from existing TLS secret

When using cert-manager instead of `argocd-agentctl` to manage PKI it's a pain to formulate the required cluster secret manually since you need to extract the CA, TLS cert and TLS key from the generated Certificate secret and then create the cluster secret using the right JSON format for config.

See the steps in the cert-manager PR here: https://github.com/argoproj-labs/argocd-agent/pull/620/files#diff-9232676f0184124a604022337fed7719340a566c7806114a08fd1aa7750c7d18R194 

The `argocd-agentctl` should have a `create-from-tls` command to create the cluster secret based on an existing TLS secret. Maybe something like:

```
argocd-agentctl pki cluster-secret <tls-secret-name> --principal-context <context> --principal-context-namespace <namespace> 
```

* Creating cluster secret manually which is a multi-step process.
* Using ESO's ability to template secrets. Basically have an ESO Kubernetes SecretStore to read the TLS secret and then create an ExternalSecret that templates out the cluster secret correctly. It has more moving parts and requires ESO be installed.

jannfis:
Assisted-by: Cursor

The `argocd-agentctl agent create` command generates its own set of certificates for authenticating Argo CD to the resource proxy, using the CLI maintained PKI. 

However, the `argocd-agentctl pki` command is not intended for production use, so users would have to roll their own cluster secrets for the use with the argocd-agent. This is error prone and not a good user experience.

This change adds the `--tls-from-secret` and `--ca-from-secret` switches to the  `argocd-agentctl agent create` command so users will be able to use the CLI to create required configuration, while using certificates managed externally (e.g. using cert-manager or other tooling).

Fixes #622 

**Checklist**

## Release Notes

* **New Features**
  * Added `--tls-from-secret` and `--ca-from-secret` CLI flags for agent creation, enabling TLS certificates and keys to be loaded from existing Kubernetes secrets.
  * Supports namespace-aware secret references (format: `namespace/name` or `name`).
  * Validates both TLS and CA secrets are provided together.

### DIFF
--- a/cmd/ctl/agent_test.go
+++ b/cmd/ctl/agent_test.go
@@ -281,7 +281,7 @@ func Test_parseSecretRef(t *testing.T) {
 			name:              "empty namespace prefix",
 			secretRef:         "/my-secret",
 			defaultNamespace:  "default-ns",
-			expectedNamespace: "",
+			expectedNamespace: "default-ns",
 			expectedName:      "my-secret",
 		},
 	}


gnunn1:
I'm testing this out, a couple of things I'm running into:

1. Even when you pass in the --principal-context it still assumes the namespace is argocd, in my IMHO it should use the namespace associated with the context

```
$ ./argocd-agentctl agent create managed-cluster --tls-from-secret=managed-cluster-principal --ca-from-secret=argocd-agent-ca --principal-context=principal
[FATAL]: could not read TLS secret argocd/managed-cluster-principal: secrets "managed-cluster-principal" not found
```
Prefixing the secret name with the namespace, i.e. `gitops-control-plane/managed-cluster-principal` works fine.

2. I'm using cert-manager to generate certificates as per the docs here: 

https://argocd-agent.readthedocs.io/latest/configuration/principal/pki-certificates/#step-1-initialize-the-pki_1

The docs create a secret for the CA which does not include the field `ca.crt` as creating a TLS certificate with kubectl doesn't allow for creating a CA but instead you get a tls.crt and tls.key fields. This command is expecting a `ca.crt` field.

```
$ ./argocd-agentctl agent create managed-cluster --tls-from-secret=gitops-control-plane/managed-cluster-principal --ca-from-secret=gitops-control-plane/argocd-agent-ca --principal-context=principal
[FATAL]: CA secret gitops-control-plane/argocd-agent-ca does not contain key 'ca.crt'
```
Should the command be able to work with a tls.crt or should I update the docs to create a secret with ca.crt and move the signing CA into a standalone secret? Updating the docs makes sense because for the agent I have a step to strip the key out of the CA secret so doing this as a standalone secret makes sense to me.

gnunn1:
Like I said I don't mind updating the docs to have a separate secret with ca.crt and if the underlying code expects that then even more reason to do it. I'll work on an update to the cert-manager docs over the weekend.

PR #688, opened 2026-01-02
fix: log streaming reliability and flaky test stability

Mangaal:
This PR fixes flaky tests and refactors event handling and log streaming code.

### **Event Handling Refactor**

1. **Introduced safeSendCloudEvent** - Prevents principal crash in `principal/event.go` when a channel is closed concurrently during response handling. Converts panic into a nil error since the response is no longer needed.
2. **Fixed duplicate event sends** - Moved scheduleRetry() in` internal/event/event.go` to call BEFORE it publishes to sentEvents map. Previously, it was in a defer, which allowed concurrent SendWaitingEvents loops to observe retryAfter==nil and immediately retry, causing duplicate sends within the same second.

### **Log Streaming Refactor**

1. **Fixed gRPC stream not properly closed** - Agent was sending EOF message but not calling CloseAndRecv(), causing intermittent "EOF" errors in the Argo CD UI.
2. **Fixed race condition in EOF handling** - Watchdog could cancel stream after WaitForCompletion returned. Fixed by closing doneCh before signalling completeCh.
3. **Fixed session cleanup** - Added defer RemoveSession() in HTTP handler to ensure cleanup on timeout/disconnect.
4. **Increased log fetch timeout in e2e test** - Changed from previous value to 90*time.Second with 1*time.Second poll interval to handle slow environments and transient failures during log streaming.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable log streaming: guaranteed terminal signaling, enforced stream closure on EOF/errors, ensured cleanup and watchdog termination, and safer shutdown ordering.
  * Safer event delivery to avoid panics and surface send/cancel errors.
  * Non-stream log timeout behavior improved (longer wait, optional timestamped timeout message) and detection/reporting of upstream or timeout indicators.

* **New Features**
  * Public session removal endpoint to explicitly finalize sessions.

* **Tests**
  * Improved log-stream tests and e2e log decoding, increased polling timeouts, and clearer test error logging.

### DIFF
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -757,6 +757,8 @@ type EventWriter struct {
 }
 
 type eventMessage struct {
+	// when this lock is owned, never attempt to THEN acquire `eventWriter.mu`, as this will lead to a deadlock.
+	// If you require `eventWriter.mu`, you must acquire that lock FIRST, before attempting to acquiring `mu` (to avoid deadlock)
 	mu sync.RWMutex
 
 	// latest event for a resource
--- a/principal/apis/logstreamapi/logstream.go
+++ b/principal/apis/logstreamapi/logstream.go
@@ -61,11 +61,20 @@ type httpWriter struct {
 }
 
 type logClient struct {
+	mu           sync.Mutex
 	ctx          context.Context
 	cancelFn     context.CancelFunc
 	logCtx       *logrus.Entry
 	requestID    string
 	terminateErr error // returned as stream status when set
+}
+
+func (c *logClient) setTerminateErr(err error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.terminateErr == nil {
+		c.terminateErr = err
+	}
 }
 
 func (s *Server) newLogClient(ctx context.Context) *logClient {
@@ -147,7 +156,7 @@ func (s *Server) RegisterHTTP(requestUUID string, w http.ResponseWriter, r *http
 	return nil
 }
 
-// StreamLogs receives log data from agent and forwards to HTTP writer
+// StreamLogs receives log data from agent
 func (s *Server) StreamLogs(stream logstreamapi.LogStreamService_StreamLogsServer) error {
 	c := s.newLogClient(stream.Context())
 	// One background pump goroutine per stream.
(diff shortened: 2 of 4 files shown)


Mangaal:
Hi @jgwest, can I get a review?

principal/event.go
```diff
@@ -588,14 +582,30 @@ func (s *Server) processResourceEventResponse(ctx context.Context, agentName str
...
+	return err
+}
+
+// safeSendCloudEvent attempts to send a CloudEvent on a channel.
+// If the channel is closed concurrently, it converts the resulting panic into
+// a benign outcome (nil error), because the response is no longer needed.
+func safeSendCloudEvent(ctx context.Context, ch chan *cloudevents.Event, ev *cloudevents.Event) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			// Channel was closed by StopTracking() while the response was in-flight.
```
jgwest, line 598: If there is a panic, can we log it, for example, at error or warn log level?

principal/apis/logstreamapi/logstream.go
```diff
@@ -96,104 +109,129 @@ func (s *Server) RegisterHTTP(requestUUID string, w http.ResponseWriter, r *http
...
-		if c.requestID == "" {
-			c.requestID = msg.GetRequestUuid()
-			c.logCtx = c.logCtx.WithField("request_id", c.requestID)
-			c.logCtx.Info("LogStream started")
-
-			s.mu.Lock()
-			if sess, ok := s.sessions[c.requestID]; ok {
-				sess.cancelFn = func() {
-					// tag this stream as terminated due to client detach TTL
+			goto done
```
jgwest: Rather than using `goto`, could we instead add a label to `(label): for` and use `break (label)`, or move the logic into a function and use `return`? I'm not reflexively against `goto`, but AFAICT there are better options in this case (correct me if I'm wrong).

internal/event/event.go
```diff
@@ -1061,22 +1061,22 @@ func (ew *EventWriter) sendUnsentEvent(resID string) {
 		return
 	}
 
-	// Move to sentEvents BEFORE unlocking to prevent ACK race
-	// Fire-and-forget events (ACKs and heartbeats) don't need retry tracking
 	target := Target(eventMsg.event)
 	isFireAndForget := target == TargetEventAck || target == TargetHeartbeat
 	if !isFireAndForget {
+		// IMPORTANT: Set retryAfter *before* publishing into sentEvents.
```
jgwest, line 1069: Given that the lock ordering is changed here, can you add this comment to eventMessage struct earlier in the file (presuming you agree):
```go
type eventMessage struct {

	// when this lock is owned, never attempt to THEN acquire `eventWriter.mu`, as this will lead to a deadlock. 
	// If you require `eventWriter.mu`, you must acquire that lock FIRST, before attempting to acquiring `mu` (to avoid deadlock) 
	mu sync.RWMutex

// (...)
}
```

jgwest (approved):
LGTM, thanks @Mangaal!

Issue #479, opened 2025-07-29
Live resource view should take resource tracking configuration into account


jannfis:
**Is your feature request related to a problem? Please describe.**

The live resource view will only retrieve resources that are deemed to be managed by Argo CD. In order to make a decision whether a resource is managed by Argo CD, the agent 

* looks at the resource tracking information on the requested resource or
* if no tracking information is found, tries to resolve owner references to a resource with tracking information

Right now, only the default resource tracking labels/annotations are being considered.

**Describe the solution you'd like**

The agent should inspect the underlying Argo CD installation to figure out which resource tracking method (label, annotation) is being used and use that information for determining whether a resource is actually managed by the Argo CD instance or not. It should take into account the method, and the actual name of the label or annotation used.

**Describe alternatives you've considered**

**Additional context**

PR #690, opened 2026-01-05
feat: add configurable resource tracking support

Rizwana777:
JIRA - https://issues.redhat.com/browse/GITOPS-8291

Fixes: https://github.com/argoproj-labs/argocd-agent/issues/479

## Summary by CodeRabbit

* **New Features**
  * Added runtime resource tracking reader that evaluates the current tracking method (annotation, label, or both) and applies it when determining managed resources.

* **Tests**
  * Added unit tests covering all tracking methods.
  * Added end-to-end tests validating tracking behavior and verification of tracking metadata in real deployments.
  * Updated resource proxy tests to exercise tracking annotation handling.

### DIFF
--- a/agent/tracking.go
+++ b/agent/tracking.go
@@ -1,0 +1,79 @@
+// Copyright 2026 The argocd-agent Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package agent
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/argoproj-labs/argocd-agent/internal/kube"
+	"github.com/argoproj/argo-cd/v3/pkg/apis/application/v1alpha1"
+	"github.com/argoproj/argo-cd/v3/util/settings"
+)
+
+const (
+	trackingAnnotationKey = "argocd.argoproj.io/tracking-id"
+	trackingLabelKey      = "app.kubernetes.io/instance"
+)
+
+// ResourceTrackingReader reads the resource tracking configuration
+// from Argo CD's SettingsManager using an informer-based approach
+type ResourceTrackingReader struct {
+	settingsManager *settings.SettingsManager
+}
+
+// NewResourceTrackingReader creates a new ResourceTrackingReader
+// It uses Argo CD's SettingsManager which maintains a shared informer cache
+// for near real-time updates when the argocd-cm ConfigMap changes
+func NewResourceTrackingReader(ctx context.Context, kubeClient *kube.KubernetesClient, namespace string) *ResourceTrackingReader {
+	// Create SettingsManager which uses an informer for the argocd-cm ConfigMap
+	// The context is used to control the lifecycle of the informer
+	settingsMgr := settings.NewSettingsManager(ctx, kubeClient.Clientset, namespace)
+
+	return &ResourceTrackingReader{
+		settingsManager: settingsMgr,
+	}
+}
+
+// IsResourceTracked checks if a resource has Argo CD tracking information
+// based on the current tracking configuration from the SettingsManager
+func (r *ResourceTrackingReader) IsResourceTracked(labels map[string]string, annotations map[string]string) (bool, error) {
+	// Get tracking method from SettingsManager
+	// This reads from the informer cache, so it's efficient and up-to-date
+	trackingMethodStr, err := r.settingsManager.GetTrackingMethod()
+	if err != nil {
+		return false, err
+	}
+
+	// Handle different tracking methods
+	switch trackingMethodStr {
+	case string(v1alpha1.TrackingMethodLabel):
+		_, hasLabel := labels[trackingLabelKey]
+		return hasLabel, nil
+
+	case string(v1alpha1.TrackingMethodAnnotationAndLabel):
+		_, hasAnnotation := annotations[trackingAnnotationKey]
+		_, hasLabel := labels[trackingLabelKey]
+		return hasAnnotation && hasLabel, nil
+
+	case string(v1alpha1.TrackingMethodAnnotation), "":
+		// Default to annotation-based tracking (when empty or explicitly set)
+		_, hasAnnotation := annotations[trackingAnnotationKey]
+		return hasAnnotation, nil
+
+	default:
+		return false, fmt.Errorf("unknown tracking method: %s", trackingMethodStr)
+	}
+}
--- a/agent/tracking_test.go
+++ b/agent/tracking_test.go
@@ -1,0 +1,147 @@
+// Copyright 2026 The argocd-agent Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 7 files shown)


agent/tracking.go
```diff
@@ -0,0 +1,218 @@
...
+// DefaultResourceTrackingConfig returns the default tracking configuration.
+// We default to annotation+label to support all tracking methods until the
+// actual Argo CD configuration is read from the ConfigMap.
+func DefaultResourceTrackingConfig() *ResourceTrackingConfig {
+	return &ResourceTrackingConfig{
+		Method:        TrackingMethodAnnotationAndLabel,
+		AnnotationKey: "argocd.argoproj.io/tracking-id",
+		LabelKey:      "app.kubernetes.io/instance",
+	}
+}
```
Rizwana777: Using annotation+label (instead of argocd annotation default) so that agent recognizes any resources tracked by any method during startup.

Shall I keep this as it is now or shall I switch back to argocd annotation default ?

agent/tracking.go
```diff
@@ -0,0 +1,217 @@
...
+type TrackingMethod string
+
+const (
+	// TrackingMethodAnnotation uses only annotation for tracking
+	TrackingMethodAnnotation TrackingMethod = "annotation"
+	// TrackingMethodLabel uses only label for tracking
+	TrackingMethodLabel TrackingMethod = "label"
+	// TrackingMethodAnnotationAndLabel uses both annotation and label for tracking
+	TrackingMethodAnnotationAndLabel TrackingMethod = "annotation+label"
+)
```
chetan-rns, line 29: Nit: We can reuse the constants from Argo CD

https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L117

agent/tracking.go
```diff
@@ -0,0 +1,217 @@
...
+	case TrackingMethodLabel:
+		// Only check label
+		_, hasLabel := labels[r.LabelKey]
+		return hasLabel
+
+	case TrackingMethodAnnotationAndLabel:
+		// Check both annotation and label
+		_, hasAnnotation := annotations[r.AnnotationKey]
+		_, hasLabel := labels[r.LabelKey]
+		return hasAnnotation || hasLabel
```
chetan-rns: With `annotation+label` method, Argo CD adds both the label and annotation to resources. Shouldn't we check for the existence of both?

```suggestion
		return hasAnnotation && hasLabel
```

agent/tracking.go
chetan-rns: IMO, the annotation-based tracking should be the default method, which aligns with upstream Argo CD.

agent/tracking.go
```diff
@@ -0,0 +1,217 @@
...
+	})
+
+	// Do an initial refresh
+	if err := r.RefreshConfig(ctx); err != nil {
+		logCtx.WithError(err).Error("Failed to perform initial config refresh")
+	}
+
+	go func() {
+		ticker := time.NewTicker(r.refreshInterval)
+		defer ticker.Stop()
```
chetan-rns: Instead of periodically polling the API server, should we consider using Argo CD's `SettingsManager`? It maintains a shared informer cache to track the config map data. The changes will be updated almost instantly, and the users don't have to wait for the next refresh cycle. WDYT?

https://github.com/argoproj/argo-cd/blob/master/util/settings/settings.go#L752

agent/tracking.go
```diff
@@ -0,0 +1,217 @@
...
+
+	case TrackingMethodAnnotationAndLabel:
+		// Check both annotation and label
+		_, hasAnnotation := annotations[r.AnnotationKey]
+		_, hasLabel := labels[r.LabelKey]
+		return hasAnnotation || hasLabel
+
+	default:
+		// Unknown method, default to checking both
+		_, hasAnnotation := annotations[r.AnnotationKey]
```
chetan-rns: Annotation based tracking should be the default tracking method.

test/e2e/resource_tracking_test.go
```diff
@@ -0,0 +1,389 @@
...
+	err = suite.ManagedAgentClient.Get(suite.Ctx, types.NamespacedName{
+		Name:      "argocd-cm",
+		Namespace: "argocd",
+	}, cm, metav1.GetOptions{})
+	if err == nil {
+		if cm.Data == nil {
+			cm.Data = make(map[string]string)
+		}
+		cm.Data["application.resourceTrackingMethod"] = "annotation+label"
+		err = suite.ManagedAgentClient.Update(suite.Ctx, cm, metav1.UpdateOptions{})
```
chetan-rns: Should we use `fixture.EnsureUpdate` to handle update conflicts?

test/e2e/resource_tracking_test.go
chetan-rns, line 1: The e2e tests only verify if the correct annotations/labels are set by Argo CD. I think we should check if the resource proxy on the control plane can actually fetch live resources with different tracking methods. 

Example:
https://github.com/argoproj-labs/argocd-agent/blob/main/test/e2e/rp_test.go#L232

Rizwana777: I have updated the logic to check resource proxy, PTAL

Rizwana777:
@chetan-rns addressed all the comments, PTAL

agent/tracking.go
```diff
@@ -0,0 +1,123 @@
...
+// The configuration is read from SettingsManager which uses an informer,
+// so it reflects near real-time state of the argocd-cm ConfigMap
+func (r *ResourceTrackingReader) GetConfig() *ResourceTrackingConfig {
+
+	// Get tracking method from SettingsManager
+	// This reads from the informer cache, so it's efficient and up-to-date
+	trackingMethodStr, err := r.settingsManager.GetTrackingMethod()
+	if err != nil {
+		// If there's an error reading the tracking method, use the default (annotation)
+		return &ResourceTrackingConfig{
```
chetan-rns: We should handle this error properly. Can we return an error?

agent/tracking.go
```diff
@@ -0,0 +1,123 @@
...
+	case string(v1alpha1.TrackingMethodLabel):
+		trackingMethod = v1alpha1.TrackingMethodLabel
+	case string(v1alpha1.TrackingMethodAnnotationAndLabel):
+		trackingMethod = v1alpha1.TrackingMethodAnnotationAndLabel
+	default:
+		// Unknown or empty, use default (annotation-based)
+		trackingMethod = v1alpha1.TrackingMethodAnnotation
+	}
+
+	return &ResourceTrackingConfig{
```
chetan-rns: `r.settingsManager.GetTrackingMethod()` already returns the tracking method set in the `argocd-cm`. Why do we need to abstract it again within `ResourceTrackingConfig`?

agent/tracking.go
```diff
@@ -0,0 +1,123 @@
...
+	"context"
+
+	"github.com/argoproj-labs/argocd-agent/internal/kube"
+	"github.com/argoproj/argo-cd/v3/pkg/apis/application/v1alpha1"
+	"github.com/argoproj/argo-cd/v3/util/settings"
+)
+
+// ResourceTrackingConfig holds the configuration for how Argo CD tracks resources
+type ResourceTrackingConfig struct {
+	// Method is the tracking method used by Argo CD
```
chetan-rns: I'm not sure if we still need this struct. We obtain the method from the settings manager, and the AnnotationKey and the LabelKeys are constants

https://github.com/argoproj/argo-cd/blob/master/common/common.go#L201

We can refactor to make this simple: Remove the ResourceTrackingConfig and GetConfig methods and directly fetch the tracking method in the IsResourceTrackedMethod. WDYT?

agent/tracking_test.go
```diff
@@ -0,0 +1,210 @@
+// Copyright 2025 The argocd-agent Authors
```
chetan-rns:
```suggestion
// Copyright 2026 The argocd-agent Authors
```

chetan-rns (approved):
Thanks @Rizwana777 

LGTM

PR #698, opened 2026-01-09
fix: detect RST_STREAM errors and trigger reconnection

drewbailey:
The agent was logging "stream terminated by RST_STREAM" errors but not triggering reconnection because these errors come as gRPC codes.Internal rather than one of the expected reconnectable error codes.

- Add detection for "stream terminated" in codes.Internal messages
- Trigger reconnection when RST_STREAM errors occur
- Fix log message in receiver goroutine (was "sending")

## Bug Fixes
* Improved connection handling for stream termination scenarios. The system now enhances its ability to detect stream termination errors and automatically triggers reconnection procedures to maintain service stability during network disruptions.
* Refined error logging to accurately report whether errors occur during stream receiving operations or stream sending operations, providing improved diagnostic clarity.

### DIFF
--- a/agent/connection.go
+++ b/agent/connection.go
@@ -195,7 +195,7 @@ func (a *Agent) handleStreamEvents() error {
 				if grpcutil.NeedReconnectOnError(err) {
 					a.SetConnected(false)
 				} else {
-					logCtx.Errorf("Error while sending to stream: %v", err)
+					logCtx.Errorf("Error while receiving from stream: %v", err)
 				}
 			}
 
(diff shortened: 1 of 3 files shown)


agent/connection.go
```diff
@@ -195,7 +195,7 @@ func (a *Agent) handleStreamEvents() error {
 				if grpcutil.NeedReconnectOnError(err) {
 					a.SetConnected(false)
 				} else {
-					logCtx.Errorf("Error while sending to stream: %v", err)
+					logCtx.Errorf("Error while receiving from stream: %v", err)
```
jannfis, line 198: 👍

PR #701, opened 2026-01-09
build: add multi arch image for amd64 + arm64

### LINKED ISSUE #624: Add ARM64 + AMD64 MultiArch build
Currently, the ArgoCD Agent and Principal images are built only for **AMD64** architecture.
Please add support for multi-architecture (**ARM64 + AMD64**) builds to ensure compatibility with heterogeneous clusters and modern ARM-based cloud environments (e.g., Graviton, Apple Silicon).

### Proposed solution:

Update the build pipeline to produce and push multi-arch Docker images.

Ensure manifest-list creation for automatic architecture resolution.

Validate builds on both **ARM64** and **AMD64** environments.

### Benefits:

Enables usage in mixed-architecture Kubernetes clusters.

Improves support for **ARM**-based cloud nodes and local development setups.

### Example:

`docker buildx build --platform linux/amd64,linux/arm64 -t argocd-agent:multiarch `

cjcocokrisp:

At the current moment, the image target of the Makefile only builds for amd64. This PR adds support for multiarch builds using `docker buildx`. This is accomplished by adding a new Makefile target `image-multiarch`. This GitHub actions CI for pushing an image is also edited to make use of this new target. The container tool binary is switched from podman to docker due to podman not having buildx support.

Fixes #624 

**Checklist**

## Summary by CodeRabbit

* **Chores**
  * Official multi-architecture image builds added for linux/amd64 and linux/arm64, improving hardware compatibility.
  * CI now prepares container runtime and QEMU emulation to enable cross-architecture builds.
  * CI introduces a conditional multi-arch build path for appropriate events and contexts.
  * Tagging switched to a portable Docker binary; push logic and registry pushes remain intact.
  * Build tooling exposes multi-arch platforms, buildx flags, and a dedicated multi-arch image target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,9 @@ ifeq ($(DOCKER_BIN),docker)
 		$(DOCKER_BIN) buildx rm argocd-agent-builder
 else
 	MULTIARCH_BUILD_CMDS?= \
+		if $(DOCKER_BIN) manifest exists $(IMAGE_REPOSITORY)/$(IMAGE_NAME):$(IMAGE_TAG); then \
+			$(DOCKER_BIN) manifest rm $(IMAGE_REPOSITORY)/$(IMAGE_NAME):$(IMAGE_TAG); \
+		fi && \
 		$(DOCKER_BIN) manifest create $(IMAGE_REPOSITORY)/$(IMAGE_NAME):$(IMAGE_TAG) && \
 		$(DOCKER_BIN) build -f Dockerfile --platform $(IMAGE_MULTIARCH_PLATFORMS) --manifest $(IMAGE_REPOSITORY)/$(IMAGE_NAME):$(IMAGE_TAG)
 endif


Makefile
```diff
@@ -1,10 +1,12 @@
...
 
-# Image names
+# Image name and options
 IMAGE_REPOSITORY?=ghcr.io/argoproj-labs/argocd-agent
 IMAGE_NAME=argocd-agent
-IMAGE_PLATFORMS?=linux/amd64
+IMAGE_PLATFORM?=linux/amd64
+IMAGE_MULTIARCH_PLATFORMS?=linux/amd64,linux/arm64
 IMAGE_TAG?=latest
+BUILDX_FLAGS?=--load
```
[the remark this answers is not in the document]
cjcocokrisp: The `--load` flag will fail on the default docker daemon due to it not supporting multiarch image store. However, I enabled the containerd image store in the CI which does support them. It will fail for the users if they do not have that set which is imo out of our control.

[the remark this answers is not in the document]
jannfis: @cjcocokrisp We could also set the BUILDX_FLAG in the CI to `--load`, and just leave them empty by default in the Makefile, i.e.

```
BUILDX_FLAGS?=
```

and then set BUILDX_FLAGS as environment in the CI when running the workflow.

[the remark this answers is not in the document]
cjcocokrisp: I agree, I think that's a good idea. Forces a user to select what they want to do when building an image locally and will skip the error case if they are using the default daemon. Will fix this right now.

Makefile
```diff
@@ -172,7 +174,13 @@ cli:
...
 .PHONY: image
 image:
-	$(DOCKER_BIN) build -f Dockerfile --platform $(IMAGE_PLATFORMS) -t $(IMAGE_REPOSITORY)/$(IMAGE_NAME):$(IMAGE_TAG) .
+	$(DOCKER_BIN) build -f Dockerfile --platform $(IMAGE_PLATFORM) -t $(IMAGE_REPOSITORY)/$(IMAGE_NAME):$(IMAGE_TAG) .
+
+.PHONY: image-multiarch
+image-multiarch:
+	$(DOCKER_BIN) buildx create --use --name argocd-agent-builder
+	$(DOCKER_BIN) buildx build $(BUILDX_FLAGS) --platform $(IMAGE_MULTIARCH_PLATFORMS) -t $(IMAGE_REPOSITORY)/$(IMAGE_NAME):$(IMAGE_TAG) .
+	$(DOCKER_BIN) buildx rm argocd-agent-builder
```
jannfis: Do we want to support both, Docker and Podman? I feel as Red Hat, we should make sure that everything works with Podman too. Podman's `buildx build` is just an alias for `build`, not supporting multi-arch builds to my knowledge.

cjcocokrisp: I agree, for Podman, it seems the way to build a multi arch image would be to use something like `podman create manifest <name>` then build for each platform and specify the `--manifest` flag with that manifest. Docker should have the same functionality so it doesn't matter which tool you use. 

Changing the `image-multiarch` target to have what's described above is probably best. How do you feel about having another target for `docker buildx` as well?

jannfis: We could have a conditional in there, e.g.

```
image-multiarch:
ifeq ($(DOCKER_BIN), podman)
  # podman multi-arch instructions
else
  # docker buildx instructions
endif
```

It's not perfect, but maybe a good start?

cjcocokrisp: Yeah I think that will be the best option for the time being. Just played around with doing the method of creating a manifest and the functionality of the two varies between the two tools. 

With podman, you can create the manifest and have it be empty and do as I mentioned above. However, with Docker you can only create a manifest with images that have been pushed to a registry. Such an odd design decision on Dockers part lol.

jannfis:
Testing this with podman gives me an error:

```
$ DOCKER_BIN=podman make image-multiarch 
podman manifest create ghcr.io/argoproj-labs/argocd-agent/argocd-agent:latest && podman build -f Dockerfile --platform linux/amd64,linux/arm64 --manifest ghcr.io/argoproj-labs/argocd-agent/argocd-agent:latest
Error: image name "ghcr.io/argoproj-labs/argocd-agent/argocd-agent:latest" is already associated with image "1c5471b04d931cfa8390d2a90f7506987ca9a01c0aba01a4a00c995cae9ca53d": that name is already in use, or use --replace to instruct Podman to do so.
make: *** [Makefile:192: image-multiarch] Error 125
```

jannfis (approved):
LGTM

PR #716, opened 2026-01-21
fix: fix intermittent update retry conflict E2E test failures

jgwest:

In a few places, our E2E tests are making unprotected K8s Update API calls, which will fail if the K8s resource is updated between Get and Update. 

This PR add Update retry-on-conflicts as needed.

**Checklist**

## Release Notes

* **Tests**
  * Improved test reliability by adding retry-aware, atomic update helpers to avoid conflicts during resource updates.
  * Refactored end-to-end tests to use standardized, safer update patterns for deployments and application resources.

* **Chores**
  * Consolidated test utilities to centralize update-and-retry logic and simplify cleanup paths.

### DIFF
--- a/test/e2e/delete_test.go
+++ b/test/e2e/delete_test.go
@@ -74,16 +74,22 @@ func (suite *DeleteTestSuite) Test_CascadeDeleteOnManagedAgent() {
 	}, 60*time.Second, 1*time.Second)
 
 	t.Log("Add a finalizer to the Deployment to keep it from being deleted")
-	depl.Finalizers = append(depl.Finalizers, "test-e2e/my-test-finalizer")
-	err = suite.ManagedAgentClient.Update(suite.Ctx, depl, metav1.UpdateOptions{})
+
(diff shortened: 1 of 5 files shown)


test/e2e/delete_test.go
```diff
@@ -114,8 +120,10 @@ func (suite *DeleteTestSuite) Test_CascadeDeleteOnManagedAgent() {
 	err = suite.ManagedAgentClient.Get(suite.Ctx, fixture.ToNamespacedName(depl), depl, metav1.GetOptions{})
 	requires.NoError(err)
 
-	depl.Finalizers = nil
-	err = suite.ManagedAgentClient.Update(suite.Ctx, depl, metav1.UpdateOptions{})
+	err = suite.ManagedAgentClient.EnsureDeploymentUpdate(suite.Ctx, fixture.ToNamespacedName(depl), func(d *appsv1.Deployment) error {
```
chetan-rns, line 123: Nit: We could use the existing generic fixture.EnsureUpdate

jgwest: Ah, I see, looks like EnsureUpdate requires an object cast, the advantage to the Ensure*Update functions are you don't need the cast (but arguably this is a very minor advantage)

Issue #720, opened 2026-01-25
Multi-Region HA / Hot-stand by support


drewbailey:
**Is your feature request related to a problem? Please describe.**

Support for a hot-stand by principle that is able to run in a separate region. In the event of a regional outage, a Global LB should be able to route traffic to the stand-by instance for agents to 

Slightly different problem than what #186 aims to solve.

## ArgoCD Agent HA Failover Design

_ai disclosure: diagrams and tables were formatted/generatd with ai_

## Overview

This document outlines the High Availability (HA) failover strategy for the argocd-agent principal component, enabling cross-region disaster recovery.

## Design Decisions

| Decision | Choice | Rationale |
|----------|--------|-----------|
| **Recovery Target** | 1-5 minutes | Acceptable for DR scenario, enables simpler DNS-based failover |
| **State Replication** | Principal-to-Principal streaming | Uses existing event streaming mechanisms, Replica stays in sync |
| **Agent Connectivity** | Single GSLB endpoint | Transparent to agents, no agent code changes |

## Architecture

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                           Global DNS / GSLB                                 │
│                      principal.argocd.example.com                           │
│                                                                             │
│   Health Checks: /healthz on each principal                                 │
│   Failover: Route to Region B when Region A unhealthy                       │
└─────────────────────┬───────────────────────────────┬───────────────────────┘
                      │                               │
                      ▼                               ▼
    ┌─────────────────────────────────────┐   ┌─────────────────────────────────┐
    │         REGION A (Primary)          │   │       REGION B (Replica)        │
    │                                     │   │                                 │
    │  ┌───────────────────────────┐      │   │  ┌───────────────────────────┐  │
    │  │     Principal Server      │      │   │  │     Principal Server      │  │
    │  │                           │      │   │  │                           │  │
    │  │  - gRPC :8403             │◄─────┼───┼──│  - Replication Client     │  │
    │  │  - Health :8080/healthz   │      │   │  │  - gRPC :8403             │  │
    │  │  - In-memory state        │──────┼───┼─►│  - Mirrors all state      │  │
    │  │  - Replication Server     │      │   │  │  - Health :8080/healthz   │  │
    │  └─────────────┬─────────────┘      │   │  └─────────────┬─────────────┘  │
    │                │                    │   │                │                │
    │  ┌─────────────▼─────────────┐      │   │  ┌─────────────▼─────────────┐  │
[...]

PR #723, opened 2026-01-27
proposal for ha failover design

drewbailey:
PR form of #720 for easier commenting / collaboration

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive HA failover guide describing an active/passive cross-region architecture, controller state model, replication and recovery flows, failover/failback scenarios and timing, replication protocol and gap recovery, security and observability recommendations, configuration options and ports, GSLB/DNS routing and health checks, operational limitations and edge cases, and proposals for future enhancements.

### DIFF
--- a/docs/proposals/ha-failover.md
+++ b/docs/proposals/ha-failover.md
@@ -139,6 +139,8 @@ The replication server uses mTLS for authentication. The replica presents a clie
 
 Both streaming (`Subscribe`) and unary (`GetSnapshot`, `Status`) RPCs are protected via separate gRPC interceptors. A stream interceptor alone is insufficient — unary RPCs require a dedicated unary interceptor.
 
+The HA admin server (port 8405) has no TLS and binds `127.0.0.1` only. Operators access it via `kubectl port-forward`; Kubernetes RBAC is the auth boundary.
+
 ```
 Replica → primary replication server (port 8404)
   TLS: mutual, replica presents cert signed by shared CA
@@ -210,7 +212,7 @@ Because the replica has been continuously replicating and writing resources to i
 
 ## CLI
 
-The `argocd-agentctl ha` subcommand manages HA state. It auto port-forwards to the principal pod via `--principal-context`, or accepts `--address` for direct gRPC.
+The `argocd-agentctl ha` subcommand manages HA state. It auto port-forwards to the principal pod's admin port (8405) via `--principal-context`, or accepts `--address` for a direct connection. The connection is plain gRPC (no TLS).
 
 | Command | Description |
 |---------|-------------|
@@ -237,11 +239,14 @@ For environments that only have simple DNS (no GSLB health checks), the operator
 
 ### Ports
 
-| Port | Purpose |
-|------|---------|
-| 8443 | Agent gRPC (mTLS) |
-| 8404 | Replication server (mTLS, primary listens) |
-| 8003 | Health check HTTP (`/healthz`) |
+| Port | Bind | TLS | Purpose |
+|------|------|-----|---------|
+| 8443 | `0.0.0.0` | mTLS | Agent gRPC |
+| 8404 | `0.0.0.0` | mTLS | Principal-to-principal replication |
+| 8405 | `127.0.0.1` | None | HAAdmin gRPC (`ha status/promote/demote`) |
+| 8003 | `0.0.0.0` | None | Health check HTTP (`/healthz`) |
+
+The admin port (8405) binds localhost only — no TLS. `argocd-agentctl` establishes a `kubectl port-forward` automatically; Kubernetes RBAC is the auth boundary.
 
 ---
 


docs/proposals/ha-failover.md
```diff
@@ -0,0 +1,708 @@
...
+## Operational Considerations
+
+### Replication Lag Monitoring
+
+### Split-Brain Prevention
+
+The replication model inherently prevents split-brain:
+
+1. **Replica knows it's a replica** - Won't accept agents while replication stream is healthy
+2. **Failover requires replication stream break** - Only then does replica become active
```
jannfis: For the split-brain problem, I can imagine the following scenario between Principal 1 (PR1) and Principal 2 (PR2):

* PR1 is primary, PR2 is replica
* PR1 isn't able to communicate to PR2 due to network failure/issue/hiccup
* Replication stream breaks
* PR2, not receiving any replication data, assumes primary role
* PR2, still able to send to (but not receive from) PR1 sends replication data to PR1
* PR1 still assumes it is primary - split brain

I do not have any idea currently on how to solve this without a quorum (i.e. a third principal or external observer), but that would inherently increase complexity.

drewbailey: Yeah, this is something I have been trying to grapple with. I think we can prevent split-brain (at the expense of availability) using self-fencing & requiring peer ACKs on replication events.

Primary must continuously prove replica is listening, if there are no ACKs, the principal self-fences.

 PR1 (primary):
    - Sends heartbeats every 5s on replication stream
    - Expects ACKs from PR2
    - No ACK for 15s → FENCED state

  PR2 (replica):
    - No heartbeat/events for 30s → promotes to STANDBY
    - Trusts that PR1 will self-fence

I also think there could be a pluggable interface to allow the use of vendor solutions like [route53 ARC](https://docs.aws.amazon.com/r53recovery/latest/dg/routing-control.html) to be an external witness & help with fencing and coordination.

A new state, fenced would allow for something like
```
  ACTIVE ──(no ACKs for ackTimeout)──► FENCED
                                          │
                                          ├─ /healthz: UNHEALTHY
                                          ├─ reject new agent connections
                                          ├─ terminate existing agent sessions
                                          │
                                          ▼
                                (peer reconnects as primary)
                                          │
                                          ▼
                                      SYNCING
```

  Failure Scenarios
```
Scenario 1: PR2 Crashes (Clean Failure)

Time    PR1 (primary)              PR2 (replica)         Agents
─────────────────────────────────────────────────────────────────
T+0     Serving agents             Dies                 Connected to PR1
T+5s    Heartbeat fails            -                    Still on PR1
T+10s   No ACKs...                 -                    Still on PR1
T+15s   FENCED                     -                    Disconnected
        /healthz UNHEALTHY

outage - no healthy principal
```

I believe this is unavoidable with two nodes & no external witness, PR1 cant distinguish “PR2 is dead” from “PR2 is partitioned”. To prevent split-brain, PR1 must always fence on lost ACKs, even if that causes a brief outage. External witness / something like ARC can help in this failure mode


  Scenario 2: Asymmetric Partition (PR1 → PR2 Broken)
```
  Time    PR1 (primary)              PR2 (replica)         DNS/GSLB
  ─────────────────────────────────────────────────────────────────
  T+0     Can't reach PR2            Replication breaks    PR1 healthy
          Heartbeats fail            Starts 30s timer
  T+15s   No ACKs → FENCED           Waiting...            PR1 UNHEALTHY
          Terminates agents                                PR2 healthy
  T+30s   -                          Promotes → STANDBY    Routes to PR2
  T+45s   -                          Agent connects
          -                          → ACTIVE              Stable on PR2

  Result: Clean failover, no split-brain. PR1 fenced before PR2 promoted.
```

```
  Scenario 3: Full Partition (Neither Can Reach Other)

  Time    PR1 (primary)              PR2 (replica)         DNS/GSLB
  ─────────────────────────────────────────────────────────────────
  T+0     Can't reach PR2            Can't receive         Depends on
          No ACKs                    from PR1              GSLB's view
  T+15s   Probes PR2: unreachable    Waiting...
          Stays ACTIVE
  T+30s   Still ACTIVE               Probes PR1:
                                     unreachable
                                     Promotes → STANDBY

            Both think peer is dead - potential dual-active
```

  Mitigation relies on DNS/GSLB. Even if both are ACTIVE-capable:
  - GSLB health checks determine who gets traffic
  - Term/epoch rejects stale updates
  - When partition heals, one demotes based on term comparison

```
  Scenario 4: Primary Network Blip (Transient)

  Time    PR1 (primary)              PR2 (replica)
  ─────────────────────────────────────────────────────────
  T+0     Network hiccup             Replication pauses
  T+5s    Still no ACKs              Waiting...
  T+10s   Network recovers           Receives heartbeat
          Receives ACK               Sends ACK
          Stays ACTIVE               Stays REPLICATING

  Result: No state change. ackTimeout (15s) > typical network blip.
```

docs/proposals/ha-failover.md
```diff
@@ -0,0 +1,1340 @@
...
+| Scenario | Self-Fencing Only | With External Authority |
+|----------|-------------------|-------------------------|
+| Typical DR | ✅ Sufficient | Overkill |
+| Regulatory/compliance | ⚠️ May not satisfy auditors | ✅ External witness documented |
+| Zero split-brain tolerance | ⚠️ Edge cases exist | ✅ Belt + suspenders |
+| Multi-cloud | ✅ Works | Need equivalent per cloud |
+
+---
+
+## Option B: Simplified Operations Model (Recommended)
```
drewbailey: @jannfis I think self-fencing, terms/epochs etc got too complex.. If we rely on external promotion decisions (manual or something like route53 ARC) the principle logic is simpler

PR #725, opened 2026-01-28
chore: update build configs

jparsai:
This PR is modify build configs, so that we can build FIPS-compliant binary from konflux. If additional env variables added in this PR are not set, then there is no effect in binary and binary works as expected in both cases.

Here is gitops release PR that will utilize these changes for konflux https://github.com/rh-gitops-midstream/release/pull/620

## Summary by CodeRabbit

* **Chores**
  * Exposed additional build-time settings to allow configurable compilation and experiment controls.
  * Added derived flag handling for tag/module selection and made the agent binary name configurable.
  * Updated build targets to honor the new settings and flags, improving flexibility of builds across environments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,11 +27,25 @@ MKDOCS_DOCKER_IMAGE?=squidfunk/mkdocs-material:9
 MKDOCS_RUN_ARGS?=
 
 # Binary names
-BIN_NAME_ARGOCD_AGENT=argocd-agent
+BIN_NAME_AGENT?=
 BIN_NAME_CLI?=argocd-agentctl
 BIN_ARCH?=$(shell go env GOARCH)
 BIN_OS?=$(shell go env GOOS)
+CGO_ENABLED?=0
+GOEXPERIMENT?=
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -28,11 +28,32 @@ MKDOCS_RUN_ARGS?=
 
 # Binary names
 BIN_NAME_ARGOCD_AGENT=argocd-agent
+BIN_NAME_AGENT?=
```
jannfis: What's the difference between BIN_NAME_ARGOCD_AGENT and BIN_NAME_AGENT? It's a bit confusing to me.

jparsai: Yes, in existing implementation we are only using `BIN_NAME_AGENT` and in this PR also I used this only, `BIN_NAME_ARGOCD_AGENT` seems to be unused which was also pointed out by Coderabbit and now I removed it. PTAL

jparsai:
> We probably should change the `cli` target to use these flags, too.

I updated `cli` target as well, PTAL.

PR #726, opened 2026-01-28
feat: support destination based agent mapping

### LINKED ISSUE #593: Consider alternative to apps-in-any-namespace for cluster mapping

The current architecture of the Agent uses apps-in-any-namespace to map applications to clusters. This works well in single-tenant use cases like cluster configuration but is potentially problematic in multi-tenant cases where users may want to use apps-in-any-namespace as intended to isolate multiple tenants.

As it is now all Applications targetting a specific cluster must go in the same namespace and the namespace must be named the same as used to register the cluster agent. This means that multiple tenants using the same cluster can no longer be isolated to a tenant specific namespace as was IMHO the original intent of apps-in-any-namespace.

I believe this is mostly an issue with managed mode and is less of a problem in autonomous since tenants would deploy their applications on the target clusters in autonomous. There may be issues for autonomous but most of my experience is with managed mode so I may not be fully considering all the implications.

We could map Applications to Clusters by using the Application.spec.destination.name instead. The target cluster would be resolved by the name matching the name of the target cluster. This change would enable Applications to be in any namespace on the control plane.

Note that you should still be able to restrict target clusters using the AppProject's ability to limit destinations so this change should not cause unrestricted access to target clusters.

I haven't thought of a good alternative.

**Additional context**

Slack Thread: https://cloud-native.slack.com/archives/C07L5SX6A9J/p1758289030402449

chetan-rns:
Support mapping applications to target agents based on the `.spec.destination.name` instead of the app namespace.

1. Added `ARGOCD_PRINCIPAL_DESTINATION_BASED_MAPPING` and `ARGOCD_AGENT_DESTINATION_BASED_MAPPING` to toggle the feature
2. Added `ARGOCD_AGENT_CREATE_NAMESPACE` to create namespaces on the fly
3. Added logic to route the apps based on the destination name
4. Track the app to agent mapping, which the redis proxy component will use.
5. Only managed mode is supported
6. Added unit tests, e2e tests, and docs.

Fixes #593 

1. Start the principal with `--destination-based-mapping` and `--allowed-namespaces` configured
2. Start the agent with `destination-based-mapping`, `--create-namespace`, and `--allowed-namespaces` configured
3. Create a new namespace on the principal called `test`
4. Create an Argo CD app in this test namespace with the `spec.destination.name` pointing to the agent name (`agent-managed`)
5. The managed agent should create a new namespace `test` and sync the app to this test namespace instead of the agent's namespace. 

**Checklist**

* [✅ ] Documentation update is required by this PR (and has been updated) OR no documentation update is required.

## Summary by CodeRabbit

* **New Features**
  * Destination-based mapping with CLI/env flags, allowed-namespaces support, and optional automatic creation of target namespaces; Redis proxy and server paths now route by app destination when enabled.

* **Behavior Changes**
  * When enabled, apps are created/managed in their original destination namespaces; autonomous agents remain namespace-bound. Invalid flag combinations are rejected.

* **Documentation**
  * New concept guide with migration guidance and examples.

* **Tests**
  * Extensive unit and end-to-end coverage added.

* **Chores**
  * Helm/Kubernetes manifests and RBAC updated to expose new config keys and namespace permissions.

### DIFF
--- a/cmd/ctl/ca.go
+++ b/cmd/ctl/ca.go
@@ -453,6 +453,53 @@ func NewPKIIssueAgentClientCert() *cobra.Command {
 			issueAndSaveSecret(globalOpts.agentContext, config.SecretNameAgentClientCert, agentNamespace, upsert, func(c *x509.Certificate, pk crypto.PrivateKey) (string, string, error) {
 				return tlsutil.GenerateClientCertificate(agentName, c, pk)
 			})
+			ctx := context.TODO()
+
+			principalClt, err := kube.NewKubernetesClientFromConfig(ctx, globalOpts.principalNamespace, "", globalOpts.principalContext)
+			if err != nil {
+				cmdutil.Fatal("Error creating Kubernetes client: %v", err)
+			}
+
+			agentClt, err := kube.NewKubernetesClientFromConfig(ctx, agentNamespace, "", globalOpts.agentContext)
+			if err != nil {
+				cmdutil.Fatal("Error creating Kubernetes client: %v", err)
+			}
+
+			caSecret, err := principalClt.Clientset.CoreV1().Secrets(globalOpts.principalNamespace).Get(ctx, config.SecretNamePrincipalCA, v1.GetOptions{})
+			if err != nil {
+				cmdutil.Fatal("Error getting CA secret from principal: %v", err)
+			}
+
+			agentSecret := &corev1.Secret{
+				ObjectMeta: v1.ObjectMeta{
+					Name:      config.SecretNameAgentCA,
+					Namespace: agentNamespace,
+				},
+				Type: corev1.SecretTypeOpaque,
+				Data: map[string][]byte{
+					"ca.crt": caSecret.Data["tls.crt"],
+				},
+			}
+
+			exists := false
+			_, err = agentClt.Clientset.CoreV1().Secrets(agentNamespace).Get(ctx, config.SecretNameAgentCA, v1.GetOptions{})
+			if err != nil {
+				if !errors.IsNotFound(err) {
+					cmdutil.Fatal("Error getting agent CA secret: %v", err)
+				}
+			} else {
+				exists = true
+			}
+
+			if !exists {
+				_, err = agentClt.Clientset.CoreV1().Secrets(agentNamespace).Create(ctx, agentSecret, v1.CreateOptions{})
+			} else {
+				_, err = agentClt.Clientset.CoreV1().Secrets(agentNamespace).Update(ctx, agentSecret, v1.UpdateOptions{})
+			}
+			if err != nil {
+				cmdutil.Fatal("Error updating agent CA secret: %v", err)
+			}
+			fmt.Printf("Agent CA secret %s/%s created\n", agentNamespace, config.SecretNameAgentCA)
 		},
 	}
 
--- a/docs/configuration/tls-certificates.md
+++ b/docs/configuration/tls-certificates.md
@@ -125,7 +125,7 @@ argocd-agentctl jwt create-key \
 
 ### Step 5: Issue Agent Client Certificate
 
-For each agent, generate and deploy a client certificate:
+For each agent, generate and deploy a client certificate. This command automatically propagates the CA certificate to the agent cluster:
 
 ```bash
 argocd-agentctl pki issue agent <agent-name> \
@@ -135,9 +135,11 @@ argocd-agentctl pki issue agent <agent-name> \
   --upsert
 ```
 
-### Step 6: Propagate CA Certificate to Agents
+This command creates both:
+- `argocd-agent-client-tls` secret (client certificate) in the agent cluster
+- `argocd-agent-ca` secret (CA certificate) in the agent cluster
 
-Copy the CA certificate to agent clusters (without the private key):
+**Note**: If you need to propagate the CA certificate separately (e.g., for existing agents), you can still use:
 
 ```bash
 argocd-agentctl pki propagate \
--- a/hack/dev-env/deploy.sh
+++ b/hack/dev-env/deploy.sh
@@ -29,6 +29,13 @@ test -x ${AGENTCTL} || (echo "Please build the CLI" && exit 1)
 
 source ${SCRIPTPATH}/utility.sh
 
+DEFAULT_IMAGE_NAME=ghcr.io/argoproj-labs/argocd-agent/argocd-agent:latest
+IMAGE_NAME=${DEFAULT_IMAGE_NAME}
+if test "$2" != ""; then
+	IMAGE_NAME="$2"
+fi
+
+
 ARGOCD_AGENT_PRINCIPAL_CONTEXT=vcluster-control-plane
 ARGOCD_AGENT_MANAGED_CONTEXT=vcluster-agent-managed
 ARGOCD_AGENT_AUTONOMOUS_CONTEXT=vcluster-agent-autonomous
@@ -46,7 +53,10 @@ cp -a ${BASEPATH}/install/kubernetes/* ${TMPDIR}
 
 deploy_principal() {
 	(
-		cd ${TMPDIR}/principal && kustomize edit set namespace argocd
+		cd ${TMPDIR}/principal && (
+			kustomize edit set namespace argocd
+			kustomize edit set image argocd-agent=${IMAGE_NAME}
+		)
 		sed -i'' \
 			-e "s/  principal.allowed-namespaces:.*/  principal.allowed-namespaces: \"agent-*\"/" \
 			principal-params-cm.yaml
@@ -58,7 +68,10 @@ deploy_principal() {
 deploy_agent_managed() {
 	(
 		principal_addr=$(getExternalLoadBalancerIP ${ARGOCD_AGENT_PRINCIPAL_CONTEXT} argocd argocd-agent-principal)
-		cd ${TMPDIR}/agent && kustomize edit set namespace argocd
+		cd ${TMPDIR}/agent && (
+			kustomize edit set namespace argocd
+			kustomize edit set image argocd-agent=${IMAGE_NAME}
+		)
 		sed -i'' \
 		        -e "s/  agent.mode:.*/  agent.mode: \"managed\"/" \
 			-e "s/  agent.creds:.*/  agent.creds: \"mtls:any\"/" \
@@ -71,7 +84,10 @@ deploy_agent_managed() {
 deploy_agent_autonomous() {
 	(
 		principal_addr=$(getExternalLoadBalancerIP ${ARGOCD_AGENT_PRINCIPAL_CONTEXT} argocd argocd-agent-principal)
-		cd ${TMPDIR}/agent && kustomize edit set namespace argocd
+		cd ${TMPDIR}/agent && (
+			kustomize edit set namespace argocd
+			kustomize edit set image argocd-agent=${IMAGE_NAME}
+		)
 		sed -i'' \
 		        -e "s/  agent.mode:.*/  agent.mode: \"autonomous\"/" \
 			-e "s/  agent.creds:.*/  agent.creds: \"mtls:any\"/" \
@@ -114,7 +130,8 @@ case "$1" in
 	undeploy_agent_autonomous || true
 	;;
 *)
-	echo "USAGE: $0 (deploy|undeploy)" >&2
+	echo "USAGE: $0 <deploy|undeploy> [image]" >&2
+	echo "[image] is optional and defaults to ${DEFAULT_IMAGE_NAME}" >&2
 	exit 1
 esac
 
--- a/install/kubernetes/argo-cd/principal/kustomization.yaml
+++ b/install/kubernetes/argo-cd/principal/kustomization.yaml
@@ -106,3 +106,15 @@ patches:
         apiGroups: ["argoproj.io"]
         resources: ["applications"]
         verbs: ["update"]
+- target:
+    group: networking.k8s.io
+    version: v1
+    kind: NetworkPolicy
+    name: argocd-redis-network-policy
+  patch: |-
+    - op: add
+      path: /spec/ingress/0/from/-
+      value:
+        podSelector:
+          matchLabels:
+            app.kubernetes.io/name: argocd-agent-principal


agent/inbound.go
```diff
@@ -876,3 +896,47 @@ func (a *Agent) deleteRepository(repo *corev1.Secret) error {
...
+	_, err = a.kubeClient.Clientset.CoreV1().Namespaces().Create(a.context, ns, metav1.CreateOptions{})
+	if err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to create namespace %s: %w", namespace, err)
+	}
+
+	log().Infof("Created namespace %s", namespace)
+	return nil
```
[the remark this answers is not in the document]
chetan-rns: Updated

docs/concepts/agent-mapping.md
```diff
@@ -0,0 +1,255 @@
...
+When an Application is created on the control plane (principal), argocd-agent needs to determine which agent should manage it. The mapping mode controls how this routing decision is made.
+
+| Feature | Namespace-Based Mapping | Destination-Based Mapping |
+|---------|------------------------|---------------------------|
+| Routing mechanism | Application namespace | `spec.destination.name` |
+| Apps per namespace | Single agent | Multiple agents |
+| Multi-tenancy | Limited | Full support |
+| Configuration | Default (no config) | Requires flag |
+| App namespace on agent | Agent's namespace (typically `argocd`) | Preserves original namespace |
+| Supported agent modes | Managed and Autonomous | Managed only |
```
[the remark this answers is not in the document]
chetan-rns: Updated

test/e2e/dest_mapping_test.go
```diff
@@ -0,0 +1,820 @@
...
+	autonomousDestMapAgentName = "agent-autonomous"
+)
+
+var (
+	argoClient *fixture.ArgoRestClient
+)
+
+type DestinationMappingTestSuite struct {
+	fixture.BaseSuite
+}
```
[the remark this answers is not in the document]
chetan-rns: Updated the cleanup

principal/redisproxy/redisproxy.go
```diff
@@ -802,10 +818,22 @@ func extractAgentNameFromRedisCommandKey(redisKey string, logCtx *logrus.Entry)
...
+	// Use lookup function if available (for destination-based mapping)
+	// Otherwise fall back to namespace (which equals agent name in namespace-based mapping)
+	if rp.agentLookupFn != nil {
+		if agent := rp.agentLookupFn(namespace, appName); agent != "" {
+			return agent, nil
+		}
+	}
+
+	// Namespace is the agent name for the default namespace-based mapping
+	return namespace, nil
```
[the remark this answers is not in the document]
chetan-rns: Done

test/e2e/dest_mapping_test.go
```diff
@@ -0,0 +1,862 @@
...
+	requires.NoError(err)
+
+	defer func() {
+		err = fixture.EnsureDeletion(suite.Ctx, suite.PrincipalClient, app)
+		requires.NoError(err)
+		// Also clean up namespace on agent if it was created
+		agentNs := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: appNs}}
+		err = fixture.EnsureDeletion(suite.Ctx, suite.ManagedAgentClient, agentNs)
+		requires.NoError(err)
+	}()
```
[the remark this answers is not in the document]
chetan-rns: I don't think this is required. The test-namespace is not created/used in any other e2e test. So, it's safe to remove it with defer.

test/e2e/dest_mapping_test.go
```diff
@@ -0,0 +1,863 @@
...
+		configMap.Data["application.namespaces"] = "*"
+	})
+	requires.NoError(err)
+
+	suite.restartApplicationController()
+}
+
+func (suite *DestinationMappingTestSuite) TearDownSuite() {
+	suite.BaseSuite.TearDownTest()
+	requires := suite.Require()
```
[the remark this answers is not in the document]
chetan-rns: TearDownTest() cleans up any remaining resources created by the e2e tests. So, calling it in the TearDownSuite should be okay. TearDownSuite will be invoked after running all the tests for that suite.

principal/callbacks_test.go
```diff
@@ -1472,4 +1720,526 @@ func TestServer_updateAppCallback(t *testing.T) {
...
+			if !tt.shouldTrackApp {
+				trackedAgent := s.appToAgent.Get(tt.newApp.QualifiedName())
+				assert.Empty(t, trackedAgent, "App should not be tracked")
+			}
+
+			// Verify resources added/removed
+			if tt.expectedDeleteAgent != "" {
+				resourceKey := resources.NewResourceKeyFromApp(tt.oldApp)
+				assert.NotContainsf(t, s.resources.Get(tt.expectedDeleteAgent).GetAll(), resourceKey, "Resource should be added to new agent")
+			}
```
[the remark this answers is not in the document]
chetan-rns: Updated

cmd/argocd-agent/agent.go
```diff
@@ -350,6 +360,16 @@ func NewAgentRunCommand() *cobra.Command {
...
 
+	command.Flags().BoolVar(&destinationBasedMapping, "destination-based-mapping",
+		env.BoolWithDefault("ARGOCD_AGENT_DESTINATION_BASED_MAPPING", false),
+		"Enable destination-based mapping. When enabled, applications are synced to their original namespace and the agent watches all namespaces")
+	command.Flags().BoolVar(&createNamespace, "create-namespace",
+		env.BoolWithDefault("ARGOCD_AGENT_CREATE_NAMESPACE", false),
+		"Create target namespace if it doesn't exist when syncing applications (used with destination-based-mapping)")
+	command.Flags().StringSliceVar(&allowedNamespaces, "allowed-namespaces",
+		env.StringSliceWithDefault("ARGOCD_AGENT_ALLOWED_NAMESPACES", nil, []string{}),
+		"List of additional namespaces the agent is allowed to manage applications in")
```
jgwest: Nit, up to you if worth doing: 
- Could modify the line to  'List of additional namespaces the agent is allowed to manage applications in _via applications in any namespace feature_'

docs/concepts/agent-mapping.md
```diff
@@ -0,0 +1,250 @@
...
+argocd-agent supports two modes for mapping applications to agents: **namespace-based mapping** (default) and **destination-based mapping**. This document explains both approaches, their trade-offs, and when to use each.
+
+## Overview
+
+When an Application is created on the control plane (principal), argocd-agent needs to determine which agent should manage it. The mapping mode controls how this routing decision is made.
+
+| Feature                | Namespace-Based Mapping                  | Destination-Based Mapping    |
+|------------------------|------------------------------------------|------------------------------|
+| Routing mechanism      | Application namespace                    | `spec.destination.name`      |
+| Apps per namespace     | Single agent                             | Multiple agents              |
```
jgwest: Personally I found the 'Apps per namespace' row confusing, specifically what 'apps per namespace' feature name meant. For example, with the namespace-based mapping you can have many apps per namespace :sweat_smile: .

Perhaps an alternative feature name like 'application namespaces per agent'? 
- Namespace-based mapping: single namespace
- Destination-based mapping: multiple namespaces

Or perhaps you feel this is already covered by 'App namespace on agent' row.

WDYT?

principal/callbacks.go
```diff
@@ -779,3 +898,30 @@ func (s *Server) startSpan(operation, resourceKind string, obj metav1.Object) (c
 		),
 	)
 }
+
+type concurrentMap[K comparable, V any] struct {
```
jgwest, line 904: :+1:  for using a concurrent map here, this is a good solution for how to synchronize this map across threads without cumbersome locking.

This should be obvious to anyone using this API, but can you add a comment to concurrentMap something like this:
```go 
// concurrentMap should not be accessed via `m` field. To consume this map, use the `Get`, `Set`, and `Delete` functions.
// Likewise, especially do not range over `m` field, as this is an unsynchronized operation.
```

install/kubernetes/agent/agent-clusterrole.yaml
```diff
@@ -12,5 +12,19 @@ rules:
...
+  resources:
+  - applications
+  verbs:
+  - create
+  - get
+  - list
+  - watch
+  - update
+  - delete
+  - patch
```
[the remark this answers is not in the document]
chetan-rns: We cannot determine the namespaces the agent is allowed to operate in beforehand. Users are expected to tighten the RBAC based on whether they are using namespace/destination-based mapping. I have included a small note in the docs.

docs/concepts/agent-mapping.md
```diff
@@ -0,0 +1,253 @@
+# Agent Mapping Modes
+
+argocd-agent supports two modes for mapping applications to agents: **namespace-based mapping** (default) and **destination-based mapping**. This document explains both approaches, their trade-offs, and when to use each.
+
```
jgwest, line 4: You have a great section below  `important "Managed Mode Only"` that indicates this feature only applies to managed agents. 

I think it would be useful to mention that here, right at the top of the document. That would make that fact clear from the start. (You could either move the existing note from below, or just add another sentence up here, up to you, presuming you agree)

docs/concepts/agent-mapping.md
```diff
@@ -0,0 +1,253 @@
...
+3. **Limited Multi-Tenancy**: Teams sharing an agent must either share a namespace or use separate agents.
+
+4. **App Namespace Rewriting**: Applications are always stored in the agent's configured namespace (typically `argocd`), regardless of their original namespace on the principal.
+
+## Destination-Based Mapping
+
+Destination-based mapping uses the Application's `spec.destination.name` field to determine which agent handles it. This enables more flexible routing and better multi-tenancy support.
+
+!!! important "Managed Mode Only"
+    Destination-based mapping only applies to **managed agents**. Autonomous agents do not require this feature because the agent is the source of truth and controls where applications are stored locally. On the principal, autonomous apps are always stored in the `{agentName}` namespace for clear ownership.
```
jgwest: > Autonomous agents do not require this feature because the agent is the source of truth and controls where applications are stored locally.

This may not be related to this PR/feature, but do we currently support 'apps in any namespace' in the autonomous agent case?

That is, with an automonous agent I can do something like this:
- Namespace 'ns1':
    - Application 'a1'
- Namespace 'ns2':
    - Application 'a2'
    
And argocd-agent autonomous mode would watch namespaces ns1/ns2, and synchronize those Applications back to principal.
- (This would require configuring the AppProject and agent settings to allow this, of course.)

Looking at [the relevant code](https://github.com/argoproj-labs/argocd-agent/blob/773d3fe297b0864152dae11ddce642f760550a5f/agent/agent.go#L230), it seems like autonomous mode agent will only see Applications in its own namespace.

docs/concepts/agent-mapping.md
```diff
@@ -0,0 +1,253 @@
...
+
+argocd-agent supports two modes for mapping applications to agents: **namespace-based mapping** (default) and **destination-based mapping**. This document explains both approaches, their trade-offs, and when to use each.
+
+## Overview
+
+When an Application is created on the control plane (principal), argocd-agent needs to determine which agent should manage it. The mapping mode controls how this routing decision is made.
+
+| Feature                | Namespace-Based Mapping                  | Destination-Based Mapping    |
+|------------------------|------------------------------------------|------------------------------|
+| Routing mechanism      | Application namespace                    | `spec.destination.name`      |
```
jgwest: Perhaps "`spec.destination.name` on principal", to make it explicit which Application CR is modified. (This is somewhat obvious as it is in the context of managed mode, but still may be helpful here)

docs/concepts/agent-mapping.md
```diff
@@ -0,0 +1,253 @@
...
+
+2. **Namespace Management Overhead**: Requires creating and managing a namespace for each agent on the principal cluster.
+
+3. **Limited Multi-Tenancy**: Teams sharing an agent must either share a namespace or use separate agents.
+
+4. **App Namespace Rewriting**: Applications are always stored in the agent's configured namespace (typically `argocd`), regardless of their original namespace on the principal.
+
+## Destination-Based Mapping
+
+Destination-based mapping uses the Application's `spec.destination.name` field to determine which agent handles it. This enables more flexible routing and better multi-tenancy support.
```
jgwest: As above, perhaps "`.spec.destination.name` on principal" to make this point explicit.

docs/concepts/agent-mapping.md
```diff
@@ -0,0 +1,253 @@
...
+    Users can manually create this namespace on the agent or use the `--create-namespace` flag to allow the agent to create namespaces automatically when applications target namespaces that don't exist.
+
+!!! note "Allowed Namespaces"
+    With destination-based mapping, applications can exist in any namespace. Both principal and agent have `--allowed-namespaces` flags to control which namespaces they will manage. The flag supports glob patterns for flexible namespace matching.
+
+!!! note "RBAC Permissions"
+    The default installation grants cluster-wide Application permissions. For tighter security, use RoleBindings per namespace instead of ClusterRoleBinding.
+
+### Example
+
```
jgwest, line 162: As above, perhaps something like 'Application defined on principal', either above the YAML block, or inside as a comment

docs/concepts/agent-mapping.md
```diff
@@ -0,0 +1,253 @@
...
+2. **Destination-Based**: Verify `spec.destination.name` matches the agent name and both principal and agent have destination-based mapping enabled.
+
+### Redis Cache Misses
+
+If you see "error getting cached app managed resources" in the ArgoCD UI:
+
+1. **Check Mapping Mode Consistency**: Ensure both principal and agent use the same mapping mode.
+
+2. **Verify Key Format**: In destination-based mode, Redis keys should include the application namespace (e.g., `team-alpha_guestbook`).
```
jgwest: You can likely remove `2. **Verify Key Format**` as this is not something that is actionable by the user AFAICT. If redis proxy is not properly translating keys, this is a bug for a developer to fix.

docs/concepts/agent-mapping.md
```diff
@@ -0,0 +1,253 @@
...
+  namespace: agent-managed  # This determines the target agent
+spec:
+  project: default
+  source:
+    repoURL: https://github.com/argoproj/argocd-example-apps
+    path: guestbook
+  destination:
+    server: https://kubernetes.default.svc  # Ignored for routing
+    namespace: guestbook
+```
```
jgwest, line 52: A simple diagram might be useful here which shows how multiple Applications on different namespaces on principal are translated to namespaces on managed agent. WDYT?

(I always say documentation never has enough examples or diagrams, but perhaps that is specific to the way my brain builds a mental model :smile:)

docs/concepts/agent-mapping.md
```diff
@@ -0,0 +1,253 @@
+# Agent Mapping Modes
```
jgwest, line 1: Thanks for putting this doc page together, this provides a nice, detailed summary of the feature and how it works. :tada: 

I noticed there were a couple other doc pages that may need to be updated, as they are still exclusively mentioning the old mechanism.
- Even if it is just adding a reference on those pages to this one.

From skimming through the docs, these were 2 I noticed:
- docs/concepts/sync-protocol.md (I noticed, for example, Namespace Mapping section)
- docs/configuration/namespaces.md

test/e2e/dest_mapping_test.go
```diff
@@ -0,0 +1,863 @@
...
+	err = argoClient.Login()
+	requires.NoError(err)
+
+	envVar := fmt.Sprintf(`ARGOCD_AGENT_DESTINATION_BASED_MAPPING=true
+ARGOCD_AGENT_CREATE_NAMESPACE=true
+ARGOCD_PRINCIPAL_DESTINATION_BASED_MAPPING=true`)
+	err = os.WriteFile(fixture.EnvVariablesFromE2EFile, []byte(envVar+"\n"), 0644)
+	requires.NoError(err)
+
+	fixture.RestartAgent(suite.T(), "agent-managed")
```
jgwest, line 84: :+1:, it's still great we have the ability to reconfigure agent modes/setting via our E2E tests

test/e2e/dest_mapping_test.go
```diff
@@ -0,0 +1,863 @@
...
+			}
+		}
+	})
+	requires.NoError(err, "failed to restore original ConfigMap state")
+
+	suite.restartApplicationController()
+}
+
+// createTestApp creates a standard test application for destination mapping tests (managed mode)
+func createTestApp(name, namespace string, autoSync bool) *v1alpha1.Application {
```
jgwest: Nit: `createTestApp` -> `createTestAppForDestinationMappingTests`

Why? This function name is in scope with other functions in the package (and thus will be callable from other tests), but this function is really only useful for this file AFAICT.

(Unless you intend for this function to be reused by other test files, but I'm guessing not)

test/e2e/dest_mapping_test.go
```diff
@@ -0,0 +1,863 @@
...
+		},
+	}
+	if autoSync {
+		app.Spec.SyncPolicy.Automated = &v1alpha1.SyncPolicyAutomated{}
+	}
+	return app
+}
+
+// createAutonomousTestApp creates a test application for the autonomous agent
+func createAutonomousTestApp(name, namespace, targetNamespace string) *v1alpha1.Application {
```
jgwest: Nit: Same as above, `createAutonomousTestAppForDestinationMappingTests`, presuming you agree with my logic above

test/e2e/dest_mapping_test.go
```diff
@@ -0,0 +1,863 @@
...
+		if resource.Kind == "Deployment" && resource.Name == destMapGuestbookUIDepl {
+			foundDeployment = true
+			break
+		}
+	}
+	requires.True(foundDeployment, "Status should include the kustomize-guestbook-ui Deployment")
+}
+
+// TestRefreshPropagation verifies that refresh requests from the
+// principal are correctly propagated to the agent via the redis proxy.
```
jgwest: `via the redis proxy` does this agent behaviour use the redis proxy functionality?

test/e2e/dest_mapping_test.go
```diff
@@ -0,0 +1,863 @@
...
+	requires.Eventually(func() bool {
+		principalApp := argoapp.Application{}
+		err := suite.PrincipalClient.Get(suite.Ctx, fixture.ToNamespacedName(app), &principalApp, metav1.GetOptions{})
+		return err != nil && errors.IsNotFound(err)
+	}, 60*time.Second, 1*time.Second)
+}
+
+// TestRedisProxy verifies that redis proxy correctly routes
+// requests for destination-mapped applications.
+func (suite *DestinationMappingTestSuite) TestRedisProxy() {
```
jgwest, line 559: Correct me if I'm wrong, but this test is testing resource proxy, not redis proxy. 

To test redis proxy, you would ask the Argo CD API to tell you the child resources of a parent `Deployment` resource, for example, via the Argo CD resourcetree api (see redis proxy e2e test for example).

jgwest (changes requested):

docs/concepts/agent-mapping.md
chetan-rns: I agree. I find diagrams useful to build a mental model. Added diagrams for both namespace and destination-based mapping

docs/concepts/agent-mapping.md
chetan-rns: Updated the other docs pages.

test/e2e/dest_mapping_test.go
chetan-rns: It doesn't use redis proxy. Updated the comment.

test/e2e/dest_mapping_test.go
chetan-rns: You are correct. I have updated the test to use the ResourceTree API. Thanks for the example.

docs/concepts/agent-mapping.md
chetan-rns: I don't think Autonomous agents currently support apps in any namespace. I didn't include it because there is no mapping involved i.e the apps already exist on the target workload cluster/agent. If we do want to support it, the principal must make sure the app names are unique on the control plane. Example: there could be an app named 'a1' on different agent namespaces 'ns1' and 'ns2'. 

The principal should probably transform the app name on the control plane to avoid conflicts. Something like `<ns>_<appname>`. Example: the principal creates `ns1_a1` and `ns2_a2` on the control plane namespace `autonomous-agent`.

jannfis: I think for now, apps-in-any-namespace for autonomous agents is very low on the priority list.

chetan-rns:
> unexpected lack of '_' namespace/name separate: 'app|managed-resources|guestbook|1.8.3

@jannfis The Redis keys for apps created in the argocd installation namespace will only have the app name. i.e. `app|resources-tree|my-app|1.8.3.gz` instead of `app|resources-tree|<agent-name>_my-app|1.8.3.gz` I've updated the redis proxy code to handle this case and I don't see this error anymore.

jgwest (approved):
LGTM too!

PR #739, opened 2026-02-03
fix: Handle malformed cluster secrets gracefully

### LINKED ISSUE #738: Principal panics on malformed cluster secrets
**Describe the bug**

When the principal encounters a malformed cluster secret, it will panic.

1. Create a malformed cluster secret representing an agent, i.e. one that has malformed JSON in `.data.config`. Make sure the secret has both required labels on it.
2. Observe the principal panic when trying to add the agent configuration

**Expected behavior**
Principal shouldn't panic, but handle the error gracefully (i.e. print a log message, do not handle the agent)

**Additional context**

jannfis:

The principal will panic when handling malformed cluster secrets.

Fixes #738 

1. Create a malformed cluster secret (both labels on it, but random data in `.data.config`
2. Observe principal not panicing, but instead printing an error message

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed cluster secrets: parsing failures now abort processing early and surface clearer error messages.
  * Updating cluster mappings is more tolerant of missing or unparsable prior data, preventing unnecessary failures while preserving successful updates.

* **Tests**
  * Added tests covering malformed cluster secret scenarios to ensure mapping and update resilience.

### DIFF
--- a/internal/argocd/cluster/informer.go
+++ b/internal/argocd/cluster/informer.go
@@ -91,10 +91,9 @@ func (m *Manager) onClusterUpdated(old *v1.Secret, new *v1.Secret) {
 		}
 	}
 
-	// Unmap cluster from old agent. Cluster could not be mapped yet due to
-	// potential errors when adding the cluster to the manager (i.e. not being
-	// able to unmarshal the cluster from the secret). In this case, we don't
-	// want to return an error but add the cluster to the manager anyway.
+	// Unmap cluster from old agent. If the cluster isn't mapped yet (that
+	// can happen if onClusterAdded found a malformed secret), we treat the
+	// operation as upsert instead of unmapping the old cluster.
 	log().Tracef("Unmapping cluster %s from agent %s", c.Name, oldAgent)
 	if err = m.unmapCluster(oldAgent); err != nil && err != ErrNotMapped {
 		log().WithError(err).Errorf("Could not unmap cluster %s from agent %s", c.Name, oldAgent)


internal/argocd/cluster/informer.go
```diff
@@ -90,9 +91,12 @@ func (m *Manager) onClusterUpdated(old *v1.Secret, new *v1.Secret) {
 		}
 	}
 
-	// Unmap cluster from old agent
+	// Unmap cluster from old agent. Cluster could not be mapped yet due to
```
mikeshng:
```suggestion
	// Unmap cluster from old agent. Cluster could not be unmapped yet due to
```

internal/argocd/cluster/informer.go
```diff
@@ -90,9 +91,12 @@ func (m *Manager) onClusterUpdated(old *v1.Secret, new *v1.Secret) {
...
+	// potential errors when adding the cluster to the manager (i.e. not being
```
mikeshng:
```suggestion
	// potential errors when removing the cluster to the manager (i.e. not being
```

mikeshng (commented):
Nit. I don't understand the new comment. The line I am reading is about unmapping the old agent. My understanding is that we want to unmap the old agent, and if it doesn't already exist in the map, thats fine because the goal is to remove it anyway. If it's not in the map, we can safely proceed with mapping the new agent. Maybe I am not getting something.

internal/argocd/cluster/informer.go
jannfis: Oh, I think it is missing an `If`, i.e. `If cluster could not be mapped yet`. That happens when the cluster secret is malformed and passing through `onClusterAdded`, which will return before mapping the cluster. When the user now updates the secret, it will pass through `onClusterUpdated` - but if it wasn't mapped before, the update would never be picked up. Does that make sense?

jannfis: (i.e., treat it as upsert, not as pure update)

Issue #751, opened 2026-02-11
"Unexpected redis key seen" warning logs


gnunn1:
**Describe the bug**

The principal is logging a bunch of warnings about unexpected keys in OpenShift GitOps 1.19:

```
time="2026-02-11T00:53:21Z" level=warning msg="Unexpected redis key seen: 'git-refs|https://github.com/gnunn-gitops/cluster-config-v2%7C1.8.3.gz'. Redirecting to principal by default." connUUID=b6f58615-b23f-4c11-a46a-3b8fdf1d2780 function=redisFxn module=redisProxy
time="2026-02-11T00:53:21Z" level=warning msg="Unexpected redis key seen: 'git-refs|https://github.com/gnunn-gitops/cluster-config-v2%7C1.8.3.gz'. Redirecting to principal by default." connUUID=b6f58615-b23f-4c11-a46a-3b8fdf1d2780 function=redisFxn module=redisProxy
time="2026-02-11T00:53:22Z" level=warning msg="Unexpected redis key seen: 'gitdirs|https://github.com/gnunn-gitops/cluster-config-v2%7C232c92db490001cb7f239ba70a2a14ff42b7e2d0%7C1.8.3.gz'. Redirecting to principal by default." connUUID=b6f58615-b23f-4c11-a46a-3b8fdf1d2780 function=redisFxn module=redisProxy
```

After speaking to Jann it doesn't like anything to be concerned about but somethat that should be handled in code.

**Steps to reproduce the behaviour**
1. Deploy agent with OpenShift GitOps 1.19, check logs for these warnings
2. (...)
3. (...)

**Expected behavior**

No warnings in the logs

**Additional context**

Does the Agent need a mechanism to handle these more seamlessly when Argo CD adds new keys? Should this be an info or debug level message rather then a warning?

jgwest:
My philosophy for how to handle this warning is:
- For any redis keys that Argo CD Agent does not recognize, it should default to forwarding them to principal (this is currently true)
- BUT, Argo CD Agent should log any redis keys which it does not currently natively recognize 
	- I went with logging at 'WARNING' so it gets attention and prevents us (argo cd agent developers et al) from pretending they doesn't exist 😁 
- When a new warning is reported, Argo CD agent developers should investigate the warning and identify the correct behaviour:
	- Correct behaviour is 
		- A) forward it to agent redis, or
		- B) handle it locally on principal redis
[...]

PR #765, opened 2026-02-24
fix(resource-proxy): Use list options for listing resources

### LINKED ISSUE #766: Events not being filtered to specific resource
**Describe the bug**

The agent is not filtering events for the specific resource being viewed, rather it returns all the events that it has access to. This results in thousands of events when using the agent with cluster-admin rights and potentially impacts performance.

Image

In this example the resource is showing thousands of events but if you get the events for the resource using kubectl you can see there are only two events for the specific resource.

```
$ kubectl events --for=externalsecret/thanos-object-storage
LAST SEEN             TYPE      REASON         OBJECT                                 MESSAGE
21m (x8 over 24m)     Warning   UpdateFailed   ExternalSecret/thanos-object-storage   could not apply template: could not execute template: could not execute template: unable to execute template at key thanos.yaml: unable to parse template at key thanos.yaml: template: thanos.yaml:6: unexpected "\\" in operand
5m51s (x5 over 21m)   Warning   UpdateFailed   ExternalSecret/thanos-object-storage   error processing spec.data[0] (key: acm-observability), err: property access_key does not exist in data of secret "acm-observability"
```

1. Give the argocd-agent cluster-admin permissions 
2. View the events for a specific resource and note that it is showing all events.
3. (...)

**Expected behavior**
It should only show events for the specific resource.

**Additional context**
This may be what is causing the large messages in #748

jannfis:

The resource proxy on the agent would ignore any query params when listing resources. In the case of Events, this would lead to it returning all events that it had access to (potentially cluster-wide), not the ones that were requested.

Fixes #766

**Checklist**

## Summary by CodeRabbit

* **New Features**
  * Resource listing now supports query parameters: fieldSelector, labelSelector, resourceVersion, and pagination (limit, continue) for filtered and paged views.

* **Bug Fixes**
  * Unknown query keys are ignored and emitted as warnings; nil or missing params no longer cause failures.

* **Tests**
  * Added tests for param parsing, propagation of filters/pagination, invalid values handling, and nil/edge cases.

### DIFF
--- a/agent/resource.go
+++ b/agent/resource.go
@@ -383,6 +383,8 @@ func listOptionsFromParams(params map[string]string) v1.ListOptions {
 	if v, ok := params["limit"]; ok {
 		if limit, err := strconv.ParseInt(v, 10, 64); err == nil {
 			opts.Limit = limit
+		} else {
+			logCtx.Warnf("Invalid limit value in resource proxy request: %s", v)
 		}
 	}
 	return opts


agent/resource.go
```diff
@@ -331,10 +331,37 @@ func (a *Agent) getManagedResource(ctx context.Context, gvr schema.GroupVersionR
 	return res, err
 }
 
+// listOptionsFromParams builds a v1.ListOptions from the given HTTP query
+// parameter map. Unknown keys are silently ignored.
```
jgwest: > Unknown keys are silently ignored. 

WDYT about logging at 'warning' level for unknown keys? This would allow us to better detect missing support for any other argocd api resource request calls. (This behaviour is also what I went with for redis proxy for [reasons](https://github.com/argoproj-labs/argocd-agent/issues/751#issuecomment-3926613903) )

jgwest (changes requested):

agent/resource.go
jannfis: Good thinking. I think it cannot hurt to validate the incoming request params to see if something is fishy. Usually, it shouldn't happen that an unknown request parameter is being sent - so warning level makes sense to me.

PR #768, opened 2026-02-24
feat: add active/passive HA support to argocd-agent principal

drewbailey:

**feat: add active/passive HA to argocd-agent principal**

Adds cross-region active/passive High Availability to the principal server. A primary principal replicates all state to a replica in real time; an
operator CLI command promotes/demotes principals. Agents connect through a single GSLB/DNS endpoint and require zero changes.

**What's included**

- HA state machine (pkg/ha) — five states (RECOVERING → ACTIVE/SYNCING → REPLICATING ↔ DISCONNECTED). Only ACTIVE returns 200 from /healthz. Principals never autonomously promote; all role transitions are operator-driven, eliminating split-brain without a coordinator.

- Replication framework (pkg/replication, principal/apis/replication) — bidirectional gRPC stream (port 8404) where the replica subscribes to all events across all agents. Subscribe opens before GetSnapshot so the primary buffers events during the snapshot transfer — preventing event loss at the snapshot boundary. Sequence numbers and periodic reconciliation (1m) bound drift to at most one reconciliation interval in the worst case.

- Principal integration (principal/ha_integration.go, principal/callbacks.go) — wires the HA controller into the principal server. On snapshot receipt the replica upserts Applications, AppProjects, and ApplicationSets directly into its K8s cluster. Informer callbacks skip resource tracking when not ACTIVE to prevent stale-cleanup false positives on the replica.

- mTLS + allowlist auth — the replication server enforces mTLS on both streaming (Subscribe) and unary (GetSnapshot, Status) RPCs via separate interceptors. Identity extracted from SPIFFE URI SAN and checked against --ha-allowed-replication-clients.

- HAAdmin server — localhost-only gRPC server (port 8405, no TLS). Exposes promote/demote/status RPCs. Intended to be accessed exclusively via kubectl port-forward; Kubernetes RBAC is the auth boundary.

- CLI (cmd/ctl/ha.go) — argocd-agentctl ha status/promote/demote. Auto-discovers the principal pod and establishes a port-forward; accepts --address for direct access.

- ApplicationSet replication (for principal replication only) — ApplicationSets replicated alongside Applications and AppProjects, with ownerRef remapping on the replica so ArgoCD doesn't re-reconcile them as orphans.

Ports
| Port | Bind | TLS | Purpose |
|------|------|-----|---------|
| 8443 | `0.0.0.0` | mTLS | Agent gRPC (unchanged) |
| 8404 | `0.0.0.0` | mTLS | Replication stream |
| 8405 | `127.0.0.1` | None | HAAdmin (CLI) |
| 8003 | `0.0.0.0` | None | `/healthz` |

Scope / non-goals

- No autonomous promotion — operator must run ha promote
- No shared storage — replica writes all resources to its own K8s cluster
- Agent configuration unchanged — agents see a single DNS name
- Secrets not replicated — operator provisions independently per region

## Summary by CodeRabbit

* **New Features**
  * High Availability (HA): controller, metrics, admin API, promote/demote, health/status endpoints, CLI commands, and HA integration into server lifecycle.
  * Replication: forwarder, client, gRPC replication service with snapshot/ACK flows and auth controls.
  * ApplicationSet support: backend, manager, events, and event forwarding.

* **Bug Fixes**
  * Improved per-stream connection/heartbeat lifecycle and connection cleanup to avoid leaks.

* **Documentation**
  * Added HA failover design proposal.

* **Tests**
  * Extensive unit/integration tests for HA, replication, and ApplicationSet flows.

* **Chores**
  * Added .claude/settings.local.json to .gitignore.

### DIFF
--- a/internal/manager/application/application.go
+++ b/internal/manager/application/application.go
@@ -196,12 +196,11 @@ func (m *ApplicationManager) Upsert(ctx context.Context, app *v1alpha1.Applicati
 		// rejects the update with a storage precondition error (Code 4).
 		app.ResourceVersion = existing.ResourceVersion
 		app.UID = existing.UID
-		if v, ok := existing.Annotations[manager.SourceUIDAnnotation]; ok {
-			if app.Annotations == nil {
-				app.Annotations = make(map[string]string)
-			}
-			app.Annotations[manager.SourceUIDAnnotation] = v
-		}
+		// Do NOT preserve the existing source-uid here. Create() already stamped
+		// source-uid = primary's UID before the AlreadyExists error. Preserving
+		// the existing value would allow a stale/wrong source-uid (e.g. from a
+		// previous failover where the AppSet created the app with a replica UID)
+		// to permanently override the correct primary UID from the snapshot.
 		if m.role == manager.ManagerRolePrincipal {
 			app.Operation = nil
 			stampLastUpdated(app)
--- a/internal/manager/applicationset/applicationset.go
+++ b/internal/manager/applicationset/applicationset.go
@@ -83,12 +83,9 @@ func (m *ApplicationSetManager) Upsert(ctx context.Context, appSet *v1alpha1.App
 		// rejects the update with a storage precondition error (Code 4).
 		appSet.ResourceVersion = existing.ResourceVersion
 		appSet.UID = existing.UID
-		if v, ok := existing.Annotations[manager.SourceUIDAnnotation]; ok {
-			if appSet.Annotations == nil {
-				appSet.Annotations = make(map[string]string)
-			}
-			appSet.Annotations[manager.SourceUIDAnnotation] = v
-		}
+		// Do NOT preserve the existing source-uid here. Create() already stamped
+		// source-uid = primary's UID; preserving the existing value would lock in
+		// a stale/wrong source-uid from a previous failover.
 		updated, ierr = m.appSetBackend.Update(ctx, appSet)
 		return ierr
 	})
--- a/internal/manager/applicationset/applicationset_test.go
+++ b/internal/manager/applicationset/applicationset_test.go
@@ -128,30 +128,31 @@ func Test_Create_ClearsResourceVersion(t *testing.T) {
 	be.AssertExpectations(t)
 }
 
-func Test_Upsert_PreservesSourceUID(t *testing.T) {
+func Test_Upsert_IncomingSourceUIDWins(t *testing.T) {
 	be := new(mockBackend)
 	m := NewApplicationSetManager(be, "argocd")
 	ctx := context.Background()
 
-	existing := newTestAppSet("test", "argocd", "existing-uid")
+	// existing has a stale source-uid (e.g. set during a previous failover)
+	existing := newTestAppSet("test", "argocd", "replica-uid")
 	existing.ResourceVersion = "rv-1"
 	existing.Annotations = map[string]string{
-		manager.SourceUIDAnnotation: "primary-uid",
+		manager.SourceUIDAnnotation: "stale-uid",
 	}
 
-	incoming := newTestAppSet("test", "argocd", "some-uid")
+	// incoming carries the primary's UID; Create() will stamp source-uid = "primary-uid"
+	incoming := newTestAppSet("test", "argocd", "primary-uid")
 
-	// Create fails with AlreadyExists
 	alreadyExists := k8serrors.NewAlreadyExists(schema.GroupResource{}, "test")
 	be.On("Create", ctx, mock.Anything).Return(nil, alreadyExists)
 	be.On("Get", ctx, "test", "argocd").Return(existing, nil)
+	// Update must be called with source-uid = "primary-uid" (incoming wins, stale dropped)
 	be.On("Update", ctx, mock.MatchedBy(func(a *v1alpha1.ApplicationSet) bool {
 		return a.Annotations[manager.SourceUIDAnnotation] == "primary-uid"
 	})).Return(existing, nil)
 
-	result, err := m.Upsert(ctx, incoming)
+	_, err := m.Upsert(ctx, incoming)
 	require.NoError(t, err)
-	require.Equal(t, "primary-uid", result.Annotations[manager.SourceUIDAnnotation])
 	be.AssertExpectations(t)
 }
 
--- a/internal/manager/appproject/appproject.go
+++ b/internal/manager/appproject/appproject.go
@@ -185,12 +185,9 @@ func (m *AppProjectManager) Upsert(ctx context.Context, project *v1alpha1.AppPro
 	project.UID = existing.UID
 	project.ResourceVersion = existing.ResourceVersion
 	project.Namespace = m.namespace
-	if v, ok := existing.Annotations[manager.SourceUIDAnnotation]; ok {
-		if project.Annotations == nil {
-			project.Annotations = make(map[string]string)
-		}
-		project.Annotations[manager.SourceUIDAnnotation] = v
-	}
+	// Do NOT preserve the existing source-uid here. Create() already stamped
+	// source-uid = primary's UID; preserving the existing value would lock in
+	// a stale/wrong source-uid from a previous failover.
 	if m.role == manager.ManagerRolePrincipal {
 		stampLastUpdated(project)
 	}


cmd/argocd-agent/principal.go
```diff
@@ -599,6 +701,29 @@ func getResourceProxyTLSConfigFromFiles(certPath, keyPath, caPath string) (*tls.
...
+	if caPath != "" {
+		caPool, err := tlsutil.X509CertPoolFromFile(caPath)
+		if err != nil {
+			return nil, fmt.Errorf("error loading replication CA: %w", err)
+		}
+		cfg.RootCAs = caPool
+		cfg.ClientCAs = caPool
+	}
+	return cfg, nil
+}
```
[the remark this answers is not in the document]
jannfis: Please use [tlsutil.SetTLSConfigFromFlags()](https://github.com/argoproj-labs/argocd-agent/blob/33793feb39b60e9dc813cf756b15163fbfb0498d/internal/tlsutil/tlsutil.go#L228) with values from the principal's `--tls-min-version`, `--tls-max-version` and `--tls-ciphersuites` to augment the `tls.Config` object with proper settings.

agent/connection.go
```diff
@@ -169,14 +170,13 @@ func (a *Agent) handleStreamEvents() error {
...
-		a.eventWriter = event.NewEventWriter(stream)
-		go a.eventWriter.SendWaitingEvents(a.context)
-	}
+	// Per-stream context: cancelled when this stream dies so all child
+	// goroutines (recv, send, heartbeat) exit and don't leak across reconnects.
+	streamCtx, streamCancel := context.WithCancel(a.context)
+	defer streamCancel()
+
+	a.eventWriter = event.NewEventWriter(stream)
+	go a.eventWriter.SendWaitingEvents(streamCtx)
```
jannfis, line 183: This reverted a fix we made in https://github.com/argoproj-labs/argocd-agent/commit/eea6ee4d5a7c1c5df4561f1e820cd6920d213bbf. Any particular reason for that?

internal/manager/application/application.go
```diff
@@ -175,6 +175,43 @@ func (m *ApplicationManager) Create(ctx context.Context, app *v1alpha1.Applicati
...
+	if !errors.IsAlreadyExists(err) {
+		return nil, err
+	}
+	existing, err := m.applicationBackend.Get(ctx, app.Name, app.Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("get existing application for upsert: %w", err)
+	}
+	app.UID = existing.UID
+	app.ResourceVersion = existing.ResourceVersion
+	app.Generation = existing.Generation
```
jannfis: Does this actually work when any of the values are different?

Last time I checked, the Kubernetes API would silently ignore any changes to these fields.

internal/manager/application/application_test.go
```diff
@@ -755,6 +755,8 @@ func Test_CompareSourceUIDForApp(t *testing.T) {
 
 	t.Run("should return false if the UID doesn't match", func(t *testing.T) {
 		incoming := oldApp.DeepCopy()
+		// Clear source-uid so comparison falls back to incoming.UID (normal operation path)
+		delete(incoming.Annotations, manager.SourceUIDAnnotation)
```
jannfis, line 759: Hm, what's the reason for this change to an existing test?

Would it fail without this change now?

internal/manager/applicationset/applicationset.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	existing, err := m.appSetBackend.Get(ctx, appSet.Name, appSet.Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("get existing applicationset for upsert: %w", err)
+	}
+
+	appSet.UID = existing.UID
+	appSet.ResourceVersion = existing.ResourceVersion
+	appSet.Generation = existing.Generation
```
jannfis: As with the comment in application.go, I was wondering if this actually works?

internal/manager/applicationset/applicationset.go
```diff
@@ -0,0 +1,125 @@
...
+	return m.appSetBackend.Get(ctx, name, namespace)
+}
+
+// Delete deletes an ApplicationSet.
+func (m *ApplicationSetManager) Delete(ctx context.Context, namespace string, appSet *v1alpha1.ApplicationSet, opts *backend.DeletionPropagation) error {
+	return m.appSetBackend.Delete(ctx, appSet.Name, namespace, opts)
+}
+
+// List lists all ApplicationSets across all namespaces.
+func (m *ApplicationSetManager) List(ctx context.Context) ([]v1alpha1.ApplicationSet, error) {
```
jannfis: I think this should either

* have an optional list selector, similar to what is in the [Application manager's implementation](https://github.com/argoproj-labs/argocd-agent/blob/5991957e36c5201de191a7d67ff192c65d63aeeb/internal/manager/application/application.go#L709)
* or set the namespace to the manager's (principal's) namespace, like the code rabbit comment suggested.

Probably it boils down to whether we want to support applicationsets-in-any-namespace or not.

internal/backend/kubernetes/applicationset/applicationset.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/argoproj-labs/argocd-agent/internal/backend"
+	"github.com/argoproj-labs/argocd-agent/internal/informer"
+	"github.com/argoproj/argo-cd/v3/pkg/apis/application/v1alpha1"
+	appclientset "github.com/argoproj/argo-cd/v3/pkg/client/clientset/versioned"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ backend.ApplicationSet = &KubernetesBackend{}
+
+// KubernetesBackend implements backend.ApplicationSet using the Argo CD client.
```
jannfis: Nit: I think this comment is a bit wrong, or misleading - to me, the term Argo CD client would suggest that we use the Argo CD API client, which we do not.

pkg/client/remote.go
```diff
@@ -432,6 +432,11 @@ func connectBackoff() wait.Backoff {
 // When Connect returns nil, the connection was successfully established and an
 // authentication token has been received.
 func (r *Remote) Connect(ctx context.Context, forceReauth bool) error {
+	if r.conn != nil {
+		r.conn.Close()
+		r.conn = nil
+	}
```
jannfis, line 452: I wonder if that should instead result in an error, instead of silently closing and re-opening the connection, and let the caller handle this situation appropriately?

hack/ha-test/setup.sh
```diff
@@ -0,0 +1,106 @@
...
+kind create cluster --name ha-test 2>/dev/null || echo "Cluster already exists"
+
+# Wait for cluster to be ready
+kubectl --context kind-ha-test wait --for=condition=Ready nodes --all --timeout=60s
+
+echo "=== Creating namespace and installing Argo CD CRDs ==="
+kubectl --context kind-ha-test create namespace argocd 2>/dev/null || true
+
+# Install Argo CD CRDs
+kubectl --context kind-ha-test apply -f https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/crds/application-crd.yaml
```
mikeshng: Need the ApplicationSet CRD after recent changes.

hack/ha-test/run-primary.sh
```diff
@@ -0,0 +1,21 @@
...
+    --log-level debug \
+    --namespace argocd \
+    --insecure-tls-generate \
+    --insecure-jwt-generate \
+    --auth "mtls:" \
+    --enable-resource-proxy=false \
+    --ha-enabled \
+    --ha-preferred-role primary \
+    --ha-peer-address "localhost:8444" \
+    --ha-failover-timeout 10s
```
mikeshng, line 24: Need `--ha-replication-auth`

hack/ha-test/run-replica.sh
```diff
@@ -0,0 +1,21 @@
...
+    --log-level debug \
+    --namespace argocd \
+    --insecure-tls-generate \
+    --insecure-jwt-generate \
+    --auth "mtls:" \
+    --enable-resource-proxy=false \
+    --ha-enabled \
+    --ha-preferred-role replica \
+    --ha-peer-address "localhost:8443" \
+    --ha-failover-timeout 10s
```
mikeshng, line 25: Need `--ha-replication-auth`

hack/ha-test/run-replica.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+# Run the replica principal for HA testing
+
+SCRIPT_DIR="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 || exit 1; pwd -P )"
+
+exec "$SCRIPT_DIR/argocd-agent" principal \
+    --listen-port 8444 \
+    --metrics-port 8001 \
+    --healthz-port 8004 \
```
mikeshng, line 9: For KinD cluster testing need `--ha-admin-port` not 8405 or it conflicts with the primary. Maybe a comment will be suffice.

hack/ha-test/run-primary.sh
```diff
@@ -0,0 +1,21 @@
...
+
+exec "$SCRIPT_DIR/argocd-agent" principal \
+    --listen-port 8443 \
+    --metrics-port 8000 \
+    --healthz-port 8003 \
+    --allowed-namespaces '*' \
+    --kubecontext kind-ha-test \
+    --log-level debug \
+    --namespace argocd \
+    --insecure-tls-generate \
```
mikeshng: This didn't work for me because the replica client failed to connect due to CA issue. Most likely introduced by recent change?

hack/dev-env/test-ha-local.sh
```diff
@@ -0,0 +1,129 @@
...
+    --metrics-port $REPLICA_METRICS \
+    --healthz-port $REPLICA_HEALTH \
+    --allowed-namespaces '*' \
+    --kubecontext vcluster-control-plane \
+    --log-level debug \
+    --namespace argocd \
+    --auth "mtls:CN=([^,]+)" \
+    --ha-enabled \
+    --ha-preferred-role replica \
+    --ha-peer-address "localhost:$PRIMARY_HEALTH" \
```
mikeshng:
```suggestion
    --ha-peer-address "localhost:$PRIMARY_PORT" \
```

wrong port?

hack/dev-env/test-ha-local.sh
```diff
@@ -0,0 +1,129 @@
...
+    --metrics-port $PRIMARY_METRICS \
+    --healthz-port $PRIMARY_HEALTH \
+    --allowed-namespaces '*' \
+    --kubecontext vcluster-control-plane \
+    --log-level debug \
+    --namespace argocd \
+    --auth "mtls:CN=([^,]+)" \
+    --ha-enabled \
+    --ha-preferred-role primary \
+    --ha-peer-address "localhost:$REPLICA_HEALTH" \
```
mikeshng:
```suggestion
    --ha-peer-address "localhost:$PRIMARY_PORT" \
```

wrong port?

drewbailey:
@mikeshng fixed up local script - apologies for the oversight I have been testing in a cloud environment

agent/connection.go
drewbailey: The reuse is still preserved (nil check + UpdateTarget). The if/else branches are flipped but functionally identical. SendWaitingEvents is called every reconnect (necessary since old goroutine exits when old streamCtx is cancelled).

```go
	// Per-stream context: cancelled when this stream dies so all child
	// goroutines (recv, send, heartbeat) exit and don't leak across reconnects.
	streamCtx, streamCancel := context.WithCancel(a.context)
	defer streamCancel()

	if a.eventWriter == nil {
		a.eventWriter = event.NewEventWriter(stream)
	} else {
		a.eventWriter.UpdateTarget(stream)
	}
	go a.eventWriter.SendWaitingEvents(streamCtx)
```

internal/manager/application/application_test.go
drewbailey: Yes, it would fail without the change. CompareSourceUID now checks the incoming app's source-uid annotation first (the HA failover path). Without the delete(), the DeepCopy'd annotation would still be present on the incoming object and match the existing one, causing the "UID doesn't match" test case to incorrectly pass.

```
     759 -    delete(incoming.Annotations, manager.SourceUIDAnnotation)
      759 +    // delete(incoming.Annotations, manager.SourceUIDAnnotation) // temporarily removed to verify
      760      incoming.UID = ktypes.UID("new_uid")
      761
      762      exists, uidMatch, err := m.CompareSourceUID(ctx, incoming)

Test_CompareSourceUIDForApp/should_return_false_if_the_UID' -v 2>&1)
  ⎿  Error: Exit code 1
     === RUN   Test_CompareSourceUIDForApp
     === RUN   Test_CompareSourceUIDForApp/should_return_false_if_the_UID_doesn't_match
         application_test.go:765:
                Error Trace:    /Users/drew_bailey/external/argocd-agent/internal/manager/application/application_test.go:765
                Error:          Should be false
                Test:           Test_CompareSourceUIDForApp/should_return_false_if_the_UID_doesn't_match
     --- FAIL: Test_CompareSourceUIDForApp (0.00s)
         --- FAIL: Test_CompareSourceUIDForApp/should_return_false_if_the_UID_doesn't_match (0.00s)
     FAIL
```

internal/manager/applicationset/applicationset.go
drewbailey: you are right, they were being silently ignored

pkg/client/remote.go
drewbailey: @jannfis currently nobody explicitly cleans up the connection when the stream dies. The agent's `handleStreamEvents` returns, `maintainConnection` loops back, and `Connect()` finds a dangling conn. Rather than making Connect() error out on that state (which would cause the reconnect loop to log the error and call Connect() again), we could address the lifecycle gap with a new  `Disconnect()` that is called when the stream dies, so conn is already nil before Connect() runs. wdyt?

internal/backend/kubernetes/applicationset/applicationset.go
```diff
@@ -0,0 +1,109 @@
+// Copyright 2024 The argocd-agent Authors
```
chetan-rns: Minor catch:
```suggestion
// Copyright 2026 The argocd-agent Authors
```

internal/manager/application/application.go
```diff
@@ -175,6 +175,41 @@ func (m *ApplicationManager) Create(ctx context.Context, app *v1alpha1.Applicati
...
+		if app.Annotations == nil {
+			app.Annotations = make(map[string]string)
+		}
+		app.Annotations[manager.SourceUIDAnnotation] = v
+	}
+	if m.role == manager.ManagerRolePrincipal {
+		app.Operation = nil
+		stampLastUpdated(app)
+	}
+	updated, err := m.applicationBackend.Update(ctx, app)
```
chetan-rns: Should we retry on update conflicts? 

https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict

internal/manager/applicationset/applicationset.go
```diff
@@ -0,0 +1,123 @@
...
+	appSet.ResourceVersion = existing.ResourceVersion
+
+	if v, ok := existing.Annotations[manager.SourceUIDAnnotation]; ok {
+		if appSet.Annotations == nil {
+			appSet.Annotations = make(map[string]string)
+		}
+		appSet.Annotations[manager.SourceUIDAnnotation] = v
+	}
+
+	updated, err := m.appSetBackend.Update(ctx, appSet)
```
chetan-rns: Same as above. Should we retry on conflict errors?

cmd/ctl/ha.go
```diff
@@ -0,0 +1,353 @@
...
+			defer cancel()
+
+			client, cleanup, err := getHAAdminClient(ctx, address)
+			if err != nil {
+				return err
+			}
+			defer cleanup()
+
+			if !force {
+				fmt.Println("WARNING: This will promote the principal to ACTIVE.")
```
chetan-rns, line 107: The timeout keeps ticking while waiting for the user confirmation. Is this intentional? Or should we prompt the confirmation before creating the context?

principal/callbacks.go
```diff
@@ -424,11 +472,40 @@ func (s *Server) deleteAppProjectCallback(outbound *v1alpha1.AppProject) {
...
+func (s *Server) deleteAppSetCallback(outbound *v1alpha1.ApplicationSet) {
+	if !s.IsActive() {
+		return
+	}
+	ev := s.events.ApplicationSetEvent(event.Delete, outbound)
+	s.ha.ForwardEventForReplication(event.New(ev, event.TargetApplicationSet), "", replication.DirectionOutbound)
+}
+
 func (s *Server) newRepositoryCallback(outbound *corev1.Secret) {
 	ctx, span := s.startSpan(operationcreate, "Repository", outbound)
```
chetan-rns, line 510: Should we also forward the repository events to the replica?

principal/callbacks.go
```diff
@@ -766,15 +843,18 @@ func (s *Server) syncRepositoriesForProject(ctx context.Context, projectName, ns
...
 	annotations := resource.GetAnnotations()
 	if annotations == nil {
 		return false
 	}
-	_, ok := annotations[manager.SourceUIDAnnotation]
-	return ok
+	if _, ok := annotations[manager.SourceUIDAnnotation]; !ok {
+		return false
+	}
+	return s.agentMode(resource.GetNamespace()) == types.AgentModeAutonomous
```
chetan-rns, line 861: I wonder if this could lead to a race condition. We rely on `namespaceMap` in the principal callback to get the agent mode. When the principal restarts, the informer callbacks are triggered asynchronously, but the `namespaceMap` may not be populated by then. The issue is not specific to this PR, and I think it exists in the main branch as well.

drewbailey: @chetan-rns good call, I'll file an issue to track, last thing this pr needs is more scope :D

principal/apis/replication/server.go
```diff
@@ -0,0 +1,422 @@
...
+	defer ackCancel()
+	go func() {
+		ack, err := client.stream.Recv()
+		ackCh <- ackResult{ack, err}
+	}()
+	var firstAck *replicationapi.ReplicationAck
+	select {
+	case res := <-ackCh:
+		if res.err != nil {
+			return fmt.Errorf("waiting for initial ACK: %w", res.err)
```
jannfis: Should be `status.Errorf` instead of `fmt.Errorf`

principal/apis/replication/server.go
```diff
@@ -0,0 +1,422 @@
...
+		}
+		firstAck = res.ack
```
jannfis, line 180: Could `res.ack` be `nil`, even when `res.err` is nil?

principal/apis/replication/server.go
```diff
@@ -0,0 +1,422 @@
...
+	case <-ackCtx.Done():
+		return fmt.Errorf("waiting for initial ACK: timed out after %s", s.options.InitialAckTimeout)
```
jannfis: It would probably make sense to use `status.Errorf` to signal deadline exceeded here

principal/server.go
```diff
@@ -553,6 +605,15 @@ func (s *Server) Start(ctx context.Context, errch chan error) error {
...
 	}()
 
+	// The applicationset informer lives in its own go routine
+	go func() {
+		if err := s.appSetManager.StartBackend(s.ctx); err != nil {
+			log().WithError(err).Error("ApplicationSet backend has exited non-successfully")
+		} else {
+			log().Info("ApplicationSet backend has exited")
+		}
+	}()
```
jannfis: I think this should only run with replication enabled, because ApplicationSet watches are only required in that context.

PR #789, opened 2026-03-04
fix(install): remove applicationsets CRD to fix kubectl apply annotation size limit

atilsensalduz:
The kustomization overlays for `principal` and `agent-managed` were incomplete: they removed the applicationset controller's runtime resources but left its CRD in the rendered manifest set. Because client-side `kubectl apply` stores the entire rendered manifest in an annotation, the oversized CRD triggered a hard Kubernetes API limit and broke the documented install procedure. Removing the CRD from these two overlays aligns the rendered output with the intended component set and eliminates the annotation size violation. Documenting `--server-side` for all overlays provides a consistent, future-proof install pattern that sidesteps the annotation limit regardless of upstream CRD size growth.

Fixes: 788

## Summary by CodeRabbit

* **Documentation**
  * Updated Kubernetes installation guide with server-side flag for improved configuration application handling.

* **Configuration Changes**
  * Added patch to remove applicationsets CustomResourceDefinition from agent-managed deployment configuration.

### DIFF
--- a/docs/getting-started/kubernetes/index.md
+++ b/docs/getting-started/kubernetes/index.md
@@ -67,6 +67,7 @@ Install a customized Argo CD instance that excludes components that will run on 
 ```bash
 # Apply the principal-specific Argo CD configuration
 kubectl apply -n argocd \
+  --server-side  \
   -k 'https://github.com/argoproj-labs/argocd-agent/install/kubernetes/argo-cd/principal?ref=<release-branch>' \
   --context <control-plane-context>
 ```
@@ -241,11 +242,13 @@ Replace <release-branch> with the release you wish to use:
 ```bash
 # For managed agents
 kubectl apply -n argocd \
+  --server-side \
   -k 'https://github.com/argoproj-labs/argocd-agent/install/kubernetes/argo-cd/agent-managed?ref=<release-branch>' \
   --context <workload-cluster-context>
 
 # For autonomous agents (alternative)
 # kubectl apply -n argocd \
+#   --server-side \
 #   -k 'https://github.com/argoproj-labs/argocd-agent/install/kubernetes/argo-cd/agent-autonomous?ref=<release-branch>' \
 #   --context <workload-cluster-context>
 ```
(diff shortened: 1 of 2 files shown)


install/kubernetes/argo-cd/principal/kustomization.yaml
```diff
@@ -3,6 +3,12 @@ resources:
 
 patches:
 # Get rid of everything related to the application controller
+- patch: |-
+    $patch: delete
+    apiVersion: apiextensions.k8s.io/v1
+    kind: CustomResourceDefinition
+    metadata:
+      name: applicationsets.argoproj.io
```
jannfis: Ever since https://github.com/argoproj-labs/argocd-agent/pull/726 has been merged, there is ApplicationSet support on the control plane. The docs haven't been updated yet across the board, but we shouldn't remove the ApplicationSet CRD from the control plane.

atilsensalduz: thanks for the review @jannfis 
wdyt about server-side apply? should I keep them or should I close the PR?

jannfis: Yes, `--server-side` is probably required. We do use it in the [dev-env installation scripts](https://github.com/argoproj-labs/argocd-agent/blob/d3ca8d2550c51221d6d31d176ed596fce40ce866/hack/dev-env/setup-vcluster-env.sh#L188) too for the very same reason. And we probably forgot to surface this in the docs :)

atilsensalduz: thanks @jannfis , I updated the PR.

jannfis: I think you removed one patch too many :) The ApplicationSet CRD can be safely removed from the [managed agent installation manifests](https://github.com/argoproj-labs/argocd-agent/blob/main/install/kubernetes/argo-cd/agent-managed/kustomization.yaml).

atilsensalduz: ah I thought you wanna remove it from both places 😅 I reverted for agent-managed kustomization.yaml

PR #793, opened 2026-03-06
chore: Enable data race detection in argocd-agent 'dev-env' and E2E tests

jgwest:
We currently enable go data race detection for unit tests, and the E2E test portion of E2E tests, but NOT for the actual argocd agent binaries themselves when running E2E test. (The actual binaries are where we would expect most data races to occur)

This PR adds a new script which will wrap the 'go run' calls we use for executing agents (locally). 

This script will:
- Switch the default to 'data race detection enabled' for argocd-agent when running via `make start-e2e`
- When `make test-e2e` completed, it will check the above log for DATA RACE log, and fail if found.
    - Data race detection can be disable via env var

## Summary by CodeRabbit

* **New Features**
  * Added configurable Go race-detector support for local agent start scripts (enabled by default).
* **Bug Fixes**
  * Improved e2e startup reliability by enabling stricter shell error handling and tee’ing process output to a persistent log.
* **Tests**
  * Increased the e2e test timeout to **45 minutes**.
  * E2e now automatically fails if the captured log contains `WARNING: DATA RACE`.
* **Chores**
  * Updated dev tooling to pass the race-detector setting through the e2e start workflow.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ BIN_NAME_CLI?=argocd-agentctl
 BIN_ARCH?=$(shell go env GOARCH)
 BIN_OS?=$(shell go env GOOS)
 CGO_ENABLED?=0
+ENABLE_DATA_RACE_DETECTOR?=true
 GOEXPERIMENT?=
 GO_TAGS?=
 GO_MOD?=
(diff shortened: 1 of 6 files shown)


jgwest:
Rebased to pick up https://github.com/argoproj-labs/argocd-agent/pull/1004

jannfis (approved):
LGTM, thanks!

PR #833, opened 2026-03-24
feat: multi-principal uid handling

### LINKED ISSUE #831: Better decision making for triggering delete/recreate on source UID mismatch

The argocd-agent will compare a resource's UID against the UID stored in the `source-uid` annotation to determine whether a resource update event actually targets an existing resource, or if it needs to delete and recreate the target resource. This is important, for example, when the events about deletion and recreation of a resource got lost (e.g. the agent wasn't connected, got restarted, etc when the event occured).

This works fine when there is a single source of truth (the principal). But with the recently introduced principal-to-principal sync for DR scenarios, this will result in mass deletion of resources on managed agents once the two principals switch roles. This is because the resources synced to the second principal will get a new UID assigned by Kubernetes, which then will not match the value in the `source-uid` annotation on the managed agent's resource.

We need to augment this mechanism with a way to detect a principal failover scenario where the change of UIDs on applications is expected, without a mass delete/recreate cycle being triggered on the agents.

An idea that has emerged in a [Slack discussion](https://cloud-native.slack.com/archives/C07L5SX6A9J/p1774013478202759?thread_ts=1773937754.632529&cid=C07L5SX6A9J) is to use an additional UID for the principal itself, so that the trigger for delete/recreate will move from `UID != source UID` to `UID != source-uid && principal UID == source-principal-uid` or similar. See that discussion for further details.

**Additional context**

There are a few edge cases, that probably have no to little consideration right now. IMHO, as long as the edge cases are documented properly, we can live with a 90% solution for the time being.

drewbailey:

Addresses problems outlined in #831 for multi-principal source-UID handling

Fixes #831 

I ran a failover exercise and captured application identity snapshots before failover and after reconciliation on the new principal.

### Topology
- Old principal: `m-test-a-ea1t-us`
- New principal: `m-j-a-ea1t-us`
- Agents: `mesh-a-ea1t-us`, `mesh-b-ea1t-us`

### What I captured
For every `Application` in:
- old principal
- new principal
- mesh-a agent
- mesh-b agent

I recorded:
- `metadata.name`
- `metadata.uid`
- `metadata.resourceVersion`
- `metadata.creationTimestamp`
- `metadata.deletionTimestamp`
- `argocd.argoproj.io/source-uid`
- `argocd.argoproj.io/principal-uid`
- destination info
- sync / health status

### Final result
Ownership transferred correctly after failover and reconciliation:

- Agent `principal-uid` moved from the old principal UID to the new principal UID.
- The new principal cleared the inherited `source-uid` annotations from its replicated application objects.
- Application identity remained stable throughout the exercise.

#### 1. No delete+recreate occurred
I compared `name -> uid` across the pre-failover snapshot and the final post-reconciliation snapshot in all four contexts.

Result:
- `m-test-a-ea1t-us`: no UID changes
- `m-j-a-ea1t-us`: no UID changes
- `mesh-a-ea1t-us`: no UID changes
- `mesh-b-ea1t-us`: no UID changes

I also checked:
- no `creationTimestamp` changes
- no `deletionTimestamp` present in the final snapshot

This shows the failover path updated ownership in place rather than deleting and recreating applications.

#### 2. Agent ownership moved to the new principal
Agent `principal-uid` values:

- Before failover:
  - `mesh-a-ea1t-us`: `d47ba09f-cb02-4887-8b7b-ddac0b3be37e`
  - `mesh-b-ea1t-us`: `d47ba09f-cb02-4887-8b7b-ddac0b3be37e`

- After failover and reconciliation:
  - `mesh-a-ea1t-us`: `15fe35f1-a643-418a-8c5c-4b4da266e085`
  - `mesh-b-ea1t-us`: `15fe35f1-a643-418a-8c5c-4b4da266e085`

This confirms both agents adopted the new principal UID.

#### 3. The new principal cleared inherited `source-uid`
`source-uid` counts on the new principal (`m-j-a-ea1t-us`):

- Before failover: `20`
- After failover and reconciliation: `0`

That means the promoted principal no longer retained the replicated `source-uid` values from the old principal after reconciliation.

#### 4. Application counts remained stable
Application counts were unchanged across the exercise:

- `m-test-a-ea1t-us`: `20`
- `m-j-a-ea1t-us`: `20`
- `mesh-a-ea1t-us`: `10`
- `mesh-b-ea1t-us`: `10`

**Checklist**

## Summary by CodeRabbit

* **New Features**
  * Persistent principal identity is generated, stored, and propagated; resync and emitted events are stamped with principal UID and trace context.

* **Bug Fixes / Improvements**
  * Identity-aware create/update/delete flows: in-place managed transitions, consistent spec caching under resolved source identity, safer delete-and-recreate on mismatch, and warnings for unknown events.

* **Tests**
  * Added coverage validating principal UID behavior, identity comparison outcomes, and managed-mode caching/transition paths.

### DIFF
--- a/agent/inbound.go
+++ b/agent/inbound.go
@@ -215,6 +215,15 @@ func (a *Agent) rewriteDestinationForManagedAgent(app *v1alpha1.Application) {
 	app.Spec.Destination.Name = "in-cluster"
 }
 
+func sourceUIDForApp(app *v1alpha1.Application) ktypes.UID {
+	if app.Annotations != nil {
+		if sourceUID := app.Annotations[manager.SourceUIDAnnotation]; sourceUID != "" {
+			return ktypes.UID(sourceUID)
+		}
+	}
+	return app.UID
+}
+
 // handleCreateApp applies the identity decision matrix for Create events.
 func (a *Agent) handleCreateApp(logCtx *logrus.Entry, incomingApp *v1alpha1.Application, identity *application.IdentityCompareResult, principalUID string) error {
 	if identity != nil && identity.Exists {
@@ -230,11 +239,11 @@ func (a *Agent) handleCreateApp(logCtx *logrus.Entry, incomingApp *v1alpha1.Appl
 		case identityActionTransition:
 			logCtx.Info("Received Create during principal transition. Transitioning in-place")
 			a.rewriteDestinationForManagedAgent(incomingApp)
-			_, err := a.appManager.UpdateManagedAppWithTransition(a.context, incomingApp, principalUID, string(incomingApp.UID))
+			_, err := a.appManager.UpdateManagedAppWithTransition(a.context, incomingApp, principalUID, string(sourceUIDForApp(incomingApp)))
 			if err != nil {
 				return fmt.Errorf("could not transition existing app: %w", err)
 			}
-			a.sourceCache.Application.Set(incomingApp.UID, incomingApp.Spec)
+			a.sourceCache.Application.Set(sourceUIDForApp(incomingApp), incomingApp.Spec)
 			return nil
 		case identityActionDeleteRecreate:
 			logCtx.Debug("App exists with different source UID. Deleting before recreate")
@@ -244,11 +253,11 @@ func (a *Agent) handleCreateApp(logCtx *logrus.Entry, incomingApp *v1alpha1.Appl
 		case identityActionUpdateStampUID:
 			logCtx.Info("Received Create with missing source-uid (AppSet wipe). Updating + stamping")
 			a.rewriteDestinationForManagedAgent(incomingApp)
-			_, err := a.appManager.UpdateManagedAppWithTransition(a.context, incomingApp, principalUID, string(incomingApp.UID))
+			_, err := a.appManager.UpdateManagedAppWithTransition(a.context, incomingApp, principalUID, string(sourceUIDForApp(incomingApp)))
 			if err != nil {
 				return fmt.Errorf("could not update app after source-uid wipe: %w", err)
 			}
-			a.sourceCache.Application.Set(incomingApp.UID, incomingApp.Spec)
+			a.sourceCache.Application.Set(sourceUIDForApp(incomingApp), incomingApp.Spec)
 			return nil
 		}
 	}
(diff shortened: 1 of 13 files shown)


agent/inbound.go
```diff
@@ -159,13 +160,24 @@ func (a *Agent) processIncomingApplication(ev *event.Event) error {
...
+
+	var identity *application.IdentityCompareResult
 
 	if a.mode == types.AgentModeManaged {
-		// Source UID annotation is not present for apps on the autonomous agent since it is the source of truth.
-		exists, sourceUIDMatch, err = a.appManager.CompareSourceUID(a.context, incomingApp)
+		identity, err = a.appManager.CompareIdentity(a.context, incomingApp, principalUID)
 		if err != nil {
-			return fmt.Errorf("failed to compare the source UID of app: %w", err)
+			return fmt.Errorf("failed to compare identity of app: %w", err)
```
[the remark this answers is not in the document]
drewbailey: I think valid in theory but low-risk in practice. The race window is narrow (failover + queued delete event not yet processed). The suggestion is reasonable but introduces complexity — the Delete path would need to decide what to do on principal mismatch (silently drop, error, log warning?)

agent/inbound.go
```diff
@@ -159,13 +160,24 @@ func (a *Agent) processIncomingApplication(ev *event.Event) error {
 	targetNamespace := a.getTargetNamespaceForApp(incomingApp)
 	incomingApp.SetNamespace(targetNamespace)
 
-	var exists, sourceUIDMatch bool
+	principalUID := event.PrincipalUID(ev.CloudEvent())
+
+	// Carry the principal identity from the CloudEvent extension into the
+	// Application annotations so that downstream update paths (including the
+	// normal UpdateManagedApp) can adopt it on pre-upgrade resources.
+	if principalUID != "" {
```
chetan-rns: Is the principal UID required for autonomous apps? If not, should we include it within the managed mode check below?

principal/callbacks.go
```diff
@@ -318,7 +325,7 @@ func (s *Server) newAppProjectCallback(outbound *v1alpha1.AppProject) {
 		agentAppProject := appproject.AgentSpecificAppProject(*outbound, agent, s.destinationBasedMapping)
 		ev := s.events.AppProjectEvent(event.Create, &agentAppProject)
 		// Inject trace context into the event for propagation to agent
-		tracing.InjectTraceContext(ctx, ev)
+		s.stampEvent(ctx, ev)
```
chetan-rns, line 328: We are stamping the principal UID for AppProjects and Repositories here. Do we plan to use the `CompareIdentity` for these resources on the agent?

drewbailey: I would like to handle AppProject & Repositories as a follow up to avoid scope in this PR, but I do think it would be preferred to avoid unnecessary delete+recreate churn for app projects & repositories when the principal changes. I can leave a todo + file an issue for now?

agent/inbound.go
drewbailey: 👍 updated

jannfis (approved):
LGTM

PR #834, opened 2026-03-25
fix: Send sync operation in a separate event to prevent overwriting

### LINKED ISSUE #809: User-triggered sync operation is never sent when principal and managed-agent are temporarily disconnected
**Describe the bug**
In managed agent configuration, if a sync operation is initiated by principal, but principal is disconnected from the agent at the time, the sync operation will never be received by the agent (even after reconnect).

The following is currently a bug in the logic which triggers sync operation from principal->agent in the managed agent case:
1) Initially: principal and agent are connected normally. 
    - an application 'A' exists on agent and principal, and is fully synchronized.
2) Principal and agent lose connection
3) User initiates a sync operation from principal (while agent<->principal are disconnected)
    - This is triggered by adding operation data to .operation field of Argo CD Application
4) Principal attempts to send sync operation message (the spec update of Application with non-nil .operation field), message fails.
5) (this step is a guess at what is going on) principal receives another 'application spec update event', which overwrites the previous message from 4. Principal stops trying to send message from 4. 
	- BUT, due to the new logic in 'updateAppCallback', 'Prevent sending operation back on regular spec updates' added by https://github.com/argoproj-labs/argocd-agent/pull/648, the Application's .operation update (from 4) is NOT included in the new message.
6) Agent reconnects to principal
7) Principal sends 'spec update' message to agent from 5, but that does NOT include the Application .operation field change (from 4)

### To reproduce

This will simulate persistent failures from managed agent <-> principal.
```
# Pull branch for bug reproduction
git clone https://github.com/jgwest/argocd-agent
git checkout delme-repro-argocd-agent-715-jan-2026

# setup vcluster and enabled argocd-agent config 
make setup-e2e

# start toxiproxy server (or docker)
podman run --rm --net=host -it ghcr.io/shopify/toxiproxy

# chaos tester tells toxiproxy server to block all connections between principal<->agent every 10-15 seconds, for 5-10 seconds. 
go run hack/chaos-tester/main.go

# Start local argocd-agent (agent uses toxiproxy as intermediary to principal)
make start-e2e

# Run Test_SyncManaged over and over until it fails
# For me, it takes 20-25 minutes to fail.
until-fail.sh go test -count=1 -v -v -timeout 60m -run TestSyncTestSuite/Test_TerminateOperationManaged github.com/argoproj-labs/argocd-agent/test/e2e

# until-fail.sh is simple shell script:
# https://gist.github.com/jgwest/7048a765d398519837f990120cf3fdd0
```

It takes ~40 minutes to reproduce when run on my machine in this fashion. 

When it fails, it will fail on this line of the test that waits for sync operation to start on managed agent:
https://github.com/argoproj-labs/argocd-agent/blob/e7e927667f5dbe2e5ee07465601f417c8c10fddc/test/e2e/sync_test.go#L340

chetan-rns:

Currently, changes to the operation field are sent via the regular spec update events. However, when the agent and the principal are disconnected, incoming spec update events will overwrite the old events in the queue, and the operation update may never reach the agent. In this PR, we send the operation updates in a separate event type so that it won't be overwritten by regular spec updates.

Fixes #809 

Steps to reproduce are provided in the issue

**Checklist**

## Summary by CodeRabbit

* **New Features**
  * Introduced a SetOperation event to deliver Application operations separately from spec updates; trace context is included for both events.
* **Behavior Changes**
  * Spec update events now have operations stripped; operation transitions (nil→non-nil) are emitted as a separate SetOperation event.
  * Managed agents validate incoming operation source identity and reject mismatches to prevent unintended overrides.
* **Tests**
  * Added tests covering SetOperation emission and handling in managed and autonomous modes.

### DIFF
--- a/agent/inbound.go
+++ b/agent/inbound.go
@@ -207,10 +207,10 @@ func (a *Agent) processIncomingApplication(ev *event.Event) error {
 	case event.SetOperation:
 		logCtx.Trace("Received a SetOperation event")
 		if a.mode == types.AgentModeManaged {
-			if !exists {
+			if !identity.Exists {
 				return fmt.Errorf("refusing SetOperation: app %s does not exist", incomingApp.QualifiedName())
 			}
-			if !sourceUIDMatch {
+			if !identity.SourceUIDMatch {
 				return fmt.Errorf("refusing SetOperation: source UID mismatch for app %s", incomingApp.QualifiedName())
 			}
 		}


agent/inbound.go
```diff
@@ -227,6 +227,19 @@ func (a *Agent) processIncomingApplication(ev *event.Event) error {
 		if err != nil {
 			logCtx.Errorf("Error updating application: %v", err)
 		}
+	case event.SetOperation:
+		logCtx.Trace("Received a SetOperation event")
+		switch a.mode {
+		case types.AgentModeManaged:
+			_, err = a.appManager.SetManagedOperation(a.context, incomingApp)
+		case types.AgentModeAutonomous:
+			_, err = a.appManager.UpdateOperation(a.context, incomingApp)
```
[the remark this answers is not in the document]
chetan-rns: Updated

principal/ha_integration.go
```diff
@@ -558,7 +558,7 @@ func (h *HAComponents) handleReplicatedEvent(ev *replication.ReplicatedEvent) er
 		}
 		key := resources.NewResourceKeyFromApp(app)
 		switch evType {
-		case event.Create, event.SpecUpdate, event.TerminateOperation:
+		case event.Create, event.SpecUpdate, event.SetOperation, event.TerminateOperation:
 			h.remapAppSetOwnerRefs(ctx, app)
 			if _, err := server.appManager.Upsert(ctx, app); err != nil {
```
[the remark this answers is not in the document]
chetan-rns: I've removed the event since it will be set to nil in Upsert

agent/inbound.go
```diff
@@ -227,6 +227,12 @@ func (a *Agent) processIncomingApplication(ev *event.Event) error {
 		if err != nil {
 			logCtx.Errorf("Error updating application: %v", err)
 		}
+	case event.SetOperation:
+		logCtx.Trace("Received a SetOperation event")
+		_, err = a.appManager.SetOperation(a.context, incomingApp)
+		if err != nil {
+			logCtx.Errorf("Error setting operation: %v", err)
+		}
```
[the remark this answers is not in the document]
chetan-rns: updated

principal/callbacks.go
```diff
@@ -195,6 +193,16 @@ func (s *Server) updateAppCallback(old *v1alpha1.Application, new *v1alpha1.Appl
 	s.ha.ForwardEventForReplication(event.New(ev, event.TargetApplication), agentName, replication.DirectionOutbound)
 	logCtx.WithField("event_type", ev.Type()).Tracef("Added app to send queue, total length now %d", q.Len())
 
+	// When a new operation appears (nil → non-nil), send it as a separate
```
jgwest, line 203: What about the case where the operation field is set while the principal controller is not running (pod is stopped due to OOM, node is down, operator is being upgraded, etc)?

If the controller is not running, it will not see the watch event which tells it that the `.operation` field of the Application was modified.

We will still reconcile the Application on startup, but the principal has no way of knowing whether or not the `.operation` field change was previously communicated to the agent.

jannfis: Would it be fair to remove any existing `.operation` field from the Application CR on startup?

If it has been transmitted to the agent prior, the sync will proceed on the agent's cluster and the status will be propagated to the principal eventually. For the UI, it would look like no sync in progress though.

chetan-rns: On restart, the principal's informer sends a `Create` event to the agent. So, the operation will be propagated to the agent if it still exists on the principal. The agent will run the operation and clear the field in the next status update.

There can be two cases where the operation field could still exist when the principal restarts:

1. The user triggered the operation when the principal was down. This is handled because the operation will be propagated via the `Create` event once the principal restarts. 
2. The principal sent the operation to the agent and shut down before receiving the status update (which clears the operation field). So, the operation field still exists when the principal restarts. There is a small window, where the principal will send the Create event (with operation) before it receives the status update. So, there is a chance the operation might run again on the agent cluster. This might be okay given it happens rarely?

jgwest: >  This might be okay given it happens rarely?

I'm fine with this.

Brainstorming a possible solution (maybe for another PR, if we like the idea)... If we wanted to keep track of whether we had already sent a specific operation request to agent, we could do something like this:

1) User triggers an operation on principal by modifying operation field, as per usual.
    - When principal (in managed mode) sees that Application `.operation` field has changed (via informer), it injects a randomly generated UUID in `.operation.info` field, something like "agent-operation-uuid: (randomly generated uuid)". 
        - That value is written to _principal_'s Application CR .operation field.
        - AFAICT it is safe for us to inject custom values into this field, but my search was not exhaustive.
    - Next, after that new value is written, agent  sends the full contents of that operation (including the new field) via SetOperation, as per this PR.

2) Agent receives SetOperation from principal, from above.
    - When it receives SetOperation, it compares the `agent-operation-uuid: (...)` value in  `.operation.info` from SetOperation (from above) with what is currently set on the agent's Application CR `.operation.info` field.
        - If the UUIDs match, then the SetOperation is ignored (since we assume it was already sent)
        - If the UUIDs don't match, then SetOperation is processed as usual.

Does this make sense?

chetan-rns: I like the idea of persisting the UUID with the resource since it survives the principal/agent restarts. But we should take care not to overwrite the user-added operation info. One more option is to use a dedicated label/annotation to track this operation ID.

PR #849, opened 2026-03-27
chore: extract service mesh docs from networking into their own doc page

jgwest:
- Moves the service mesh section from networking into its own page
- The mesh section is: A) is not specifically related to general networking (unlike the rest of the page)  B) long C) niche, so it makes sense to move it to its own page
- The service mesh content is otherwise unchanged from original text (I've carefully verified using meld)

If you don't agree with the move, feel free to close the PR as is.

**Checklist**

## Summary by CodeRabbit

* **Documentation**
  * Reorganized service mesh integration guidance into a dedicated configuration guide.
  * Added comprehensive setup instructions for Istio and Linkerd deployments, including authentication, TLS, and component configuration.
  * Included security best practices, troubleshooting sections for common issues, and deployment examples.


jannfis (approved):
LGTM and makes sense. Thanks.

PR #875, opened 2026-04-07
feat: extend label selectors to other agent resources

chetan-rns:
In this PR, we replace the `--app-label-selector` with a generic `--label-selector` that applies to all the resources the agent/principal reconciles. This prevents the agent/principal from syncing any non-agent resources.

The previous `--app-label-selector` is not part of any release yet. So, I think we can replace it without causing any backwards compatibility issues.

Fixes #?

1. Set `ARGOCD_AGENT_LABEL_SELECTOR` on the agent and `ARGOCD_PRINCIPAL_LABEL_SELECTOR` on the principal
2. The agent/principal will only process the resources that match this label selector

**Checklist**

## Summary by CodeRabbit

* **Refactor**
  * Renamed and generalized configuration from app-label-selector → label-selector, expanding scope to filter all watched Kubernetes resources (Applications, Projects, ApplicationSets, Repository Secrets) instead of Applications only.

* **Documentation**
  * Updated CLI flag, env var, ConfigMap keys, Helm values, and docs to reflect the rename and broader resource-filtering semantics.

* **Other**
  * Startup now logs the active label selector when set.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -148,9 +148,9 @@ type Agent struct {
 	// destination-based mapping.
 	createNamespace bool
 
-	// appLabelSelector is an optional Kubernetes label selector that restricts
-	// which Applications the agent watches.
-	appLabelSelector string
+	// labelSelector is an optional Kubernetes label selector that restricts
+	// which resources the agent can process.
+	labelSelector string
 }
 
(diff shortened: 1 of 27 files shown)


jannfis (approved):
LGTM

PR #890, opened 2026-04-10
fix: memory leak at agent component

C4tWithShell:

1. Fix  memory leak - missing `q.Done(ev)` in `sender()` 
The workqueue's Get() moves items into an internal processing set. Without Done(), every event stayed referenced forever. Over 7 days with continuous Application sync events (each ~KB), this accumulates to ~1GB. 

image

Added:
`q.Done(ev)` for the nil-event early return path
`defer q.Done(ev)` for the normal path

This matches the pattern already used on the principal side at [event.go:138](https://github.com/argoproj-labs/argocd-agent/blob/d7ac642cd416d04c6852d8448a3952f866beddaf/principal/event.go#L138).

2. connMap never cleaned up
When a Redis subscription connection times out due to inactivity, the goroutine closed the pubsub but left the connectionEntry in connMap. Over time with many agent reconnections, this map grows unbounded. 

Added `defer pubsub.Close()` and `defer delete(connMap)` at function top, so cleanup runs on every exit path (timeout at line 231, queue-gone at line 252)

N/A

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection lifecycle management to ensure queue completion is always signaled, preventing stuck items when events are absent.
  * Ensured reliable cleanup of subscription resources and removal of stale connection entries on inactivity to prevent resource leaks.

### DIFF
--- a/agent/connection.go
+++ b/agent/connection.go
@@ -85,8 +85,10 @@ func (a *Agent) sender(stream eventstreamapi.EventStream_SubscribeClient) error 
 	logCtx.Trace("Grabbed an item")
 	if ev == nil {
 		// TODO: Is this really the right thing to do?
+		q.Done(ev)
 		return nil
 	}
+	defer q.Done(ev)
 	logCtx = logCtx.WithFields(logrus.Fields{
 		"event_target": ev.DataSchema(),
 		"event_type":   ev.Type(),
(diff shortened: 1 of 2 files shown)


mikeshng (commented):
@jannfis could this be the leak that we saw in the previous scale test?

jannfis (approved):
LGTM, thanks a lot!

PR #892, opened 2026-04-10
fix: agent should remap apps from the principal namespace to the agent namespace

### LINKED ISSUE #847: Control Plane and Agent must be in same namespace when using destination mapping
**Describe the bug**

If you have the Control Plane in one namespace, `argocd`, and the agent in a different namespace, `argocd-agent`, when using destination mapping the application will not get copied over as the Agent will fail complaining it cannot create the Application in the `argocd` namespace.

1. Install the Control Plane and Agent in different namespaces using destination mapping
2. Create an application on the Control Plane
3. Note that it fails to get propagated over as the Agent tries to create the Application in the Control Name namespace rather then it's own.

**Expected behavior**
When the Application is being sourced from the Control Plane's namespace the Agent should be smart enough to change the namespace to it's own namespace.

**Additional context**
This is important because you need to install the Agent in a different namespace when you want to use Argo CD to manage the cluster the Control Plane resides on. Requiring the use of namespace mapping to handle this use case then precludes any oppty to share the Control Plane between cluster-config and multi-tenant apps use cases since users will often want to use Apps-In-Any-Namespace to support tenancy and not for mapping Applications.

chetan-rns:

When destination-based mapping is enabled, and the agent runs in a different namespace than the principal, the agent would create the app in the same namespace as the principal. This will not work when both the agent and the principal are running in the same cluster. The agent should remap the apps from the principal namespace to the agent namespace. So that the app gets created in the same namespace as the agent.

* The agent and the principal will exchange namespaces during authentication
* Agent stamps the original namespace as an annotation
* The principal relies on this annotation to find the original namespace to update the status.

We need the annotation because the principal doesn't know if the agent has remapped the namespace. For example, let's say the app was created in the tenant namespace `agent-managed` and the agent is also running in the same `agent-managed` namespace on the workload cluster. The agent will not remap because the app namespace is different from the principal namespace. The agent will send status updates from the `agent-managed` namespace. Since the status update event has the same namespace as the agent, we don't want to remap it back to the principal namespace. The principal will rely on the annotation to find the original namespace.

Fixes https://github.com/argoproj-labs/argocd-agent/issues/847

1. Install the agent and the principal in the same cluster
2. Create an app in the principal namespace
3. The agent should have created the app in its own agent namespace.

**Checklist**

## Summary by CodeRabbit

* **New Features**
  * Agents and principals now exchange and record Kubernetes namespaces during authentication, enabling conditional destination-based namespace remapping.
  * Applications remapped by the agent are stamped with an annotation so the principal can restore the original namespace.
  * Remapping respects managed vs autonomous modes and tenant vs principal boundaries.

* **Tests**
  * Added unit and integration tests covering remap logic, annotation stamping, auth namespace exchange, resync, and event handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/argoproj-labs/argocd-agent/pull/892)


chetan-rns:
@coderabbitai full review

chetan-rns:
@coderabbitai full review

internal/resync/resync_test.go
```diff
@@ -722,6 +750,204 @@ func Test_ProcessRequestUpdateEvent_GPGKey(t *testing.T) {
...
+
+		ev, shutdown := handler.sendQ.Get()
+		assert.False(t, shutdown)
+		assert.Equal(t, event.EventRequestUpdate.String(), ev.Type())
+
+		got := &event.RequestUpdate{}
+		err = ev.DataAs(got)
+		assert.Nil(t, err)
+		assert.NotEmpty(t, got.Checksum, "should have checksum because app was found after remap")
+	})
```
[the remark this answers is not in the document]
chetan-rns: updated

agent/inbound.go
```diff
@@ -1213,10 +1217,20 @@ func (a *Agent) deleteGPGKey(cm *corev1.ConfigMap) error {
...
+// annotation so that the principal can unambiguously remap it back later.
 func (a *Agent) getTargetNamespaceForApp(app *v1alpha1.Application) string {
 	if a.destinationBasedMapping && a.mode == types.AgentModeManaged {
+		principalNS := a.principalNS()
+		if principalNS != "" && app.Namespace == principalNS {
+			if app.Annotations == nil {
+				app.Annotations = make(map[string]string)
+			}
+			app.Annotations[manager.OriginalNamespaceAnnotation] = principalNS
+			return a.namespace
```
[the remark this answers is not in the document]
chetan-rns: updated

gnunn1:
I'm testing my there seems to be a problem with GPG keys on the agent pod in `argocd-agent` namespace:

```
time="2026-04-10T19:10:56Z" level=warning msg="Could not unmanage GPG key ConfigMap argocd-gpg-keys-cm: resource argocd-gpg-keys-cm is not managed" module=GrpcEvent
time="2026-04-10T19:10:56Z" level=info msg="Creating a new GPG key on behalf of an incoming event" gpgKey=argocd-gpg-keys-cm method=CreateGPGKey module=GrpcEvent
time="2026-04-10T19:10:56Z" level=error msg="Unable to process incoming event" clientAddr="192.168.1.5:443" direction=Recv error="could not create incoming GPG key after deleting existing GPG key: configmaps \"argocd-gpg-keys-cm\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>" event_id=argocd-gpg-keys-cm_f26ec720-9d81-4dbb-a0b2-63907d100874_12014992 module=StreamEvent resource_id=argocd-gpg-keys-cm_f26ec720-9d81-4dbb-a0b2-63907d100874 type=io.argoproj.argocd-agent.event.spec-update
time="2026-04-10T19:10:56Z" level=error msg="Unable to process incoming event" clientAddr="192.168.1.5:443" direction=Recv error="configmaps \"argocd-gpg-keys-cm\" is forbidden: User \"system:serviceaccount:argocd-agent:argocd-agent-agent\" cannot get resource \"configmaps\" in API group \"\" in the namespace \"argocd\"" event_id=ec38f6a2-d8a9-400e-b1bf-c6a0bc7bb688 module=StreamEvent resource_id=ec38f6a2-d8a9-400e-b1bf-c6a0bc7bb688 type=io.argoproj.argocd-agent.event.response-synced-resource
```

It looks like it is trying to interact with the configmap in the `argocd` namespace instead of `argocd-agent` where the agent resides.

agent/inbound.go
```diff
@@ -541,7 +544,8 @@ func (a *Agent) processIncomingResourceResyncEvent(ev *event.Event) error {
 
 	resyncHandler := resync.NewRequestHandler(dynClient, sendQ, a.emitter, a.resources, logCtx, manager.ManagerRoleAgent, a.namespace).
 		WithDestinationBasedMapping(a.destinationBasedMapping).
-		WithIgnoreUnmanagedApps(a.ignoreUnmanagedApps)
+		WithIgnoreUnmanagedApps(a.ignoreUnmanagedApps).
+		WithPeerNamespace(a.principalNS())
```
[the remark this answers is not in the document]
chetan-rns: It should be okay. Destination-based mapping cannot be used in autonomous mode. We check if the destination-based mapping is enabled before accessing peerNamespace.

chetan-rns:
@gnunn1 Thanks for trying out the PR. The issue should be fixed now.

agent/agent.go
```diff
@@ -596,6 +597,14 @@ func (a *Agent) Start(ctx context.Context) error {
 
 	if a.remote != nil {
 		a.remote.SetClientMode(a.mode)
+		a.remote.SetOnAuthenticated(func(principalNs string) {
+			if principalNs == "" {
```
jgwest, line 601: If principalNS is "", this is an error condition, right? A quick skim of what the principal sends agent tells me there isn't a case where this is expected behaviour.

If we're not going to return an error, we could at least log it.

agent/agent.go
```diff
@@ -604,6 +613,12 @@ func (a *Agent) Start(ctx context.Context) error {
 	return nil
 }
 
+func (a *Agent) principalNS() string {
```
jgwest, line 616: I'm curious: do we support the user changing the principal Argo CD namespace after the fact? (For example, some number of months after initial install)? If a user did change their principal Argo CD namespace, what would happen to the Applications that are associated with that namespace on principal? Would they be orphaned?

agent/inbound.go
```diff
@@ -1213,10 +1218,20 @@ func (a *Agent) deleteGPGKey(cm *corev1.ConfigMap) error {
...
-// be created on the agent. If destinationBasedMapping is enabled AND the agent
-// is in managed mode, it returns the original namespace from the principal.
+// be created on the agent. In destination-based mapping + managed mode, apps
+// whose namespace matches the principal's namespace are remapped to the agent's
+// own namespace. When remapping occurs the original namespace is recorded in an
+// annotation so that the principal can unambiguously remap it back later.
 func (a *Agent) getTargetNamespaceForApp(app *v1alpha1.Application) string {
 	if a.destinationBasedMapping && a.mode == types.AgentModeManaged {
+		principalNS := a.principalNS()
+		if principalNS != "" && app.Namespace == principalNS && a.namespace != principalNS {
```
jgwest, line 1228: If principalNS is "" in this code block, it seems like this would be an error (thus we should either log it, or log and return error)

internal/resync/resync.go
```diff
@@ -161,9 +170,25 @@ func (r *RequestHandler) ProcessIncomingSyncedResource(ctx context.Context, inco
...
+	// always live in the Argo CD namespace regardless of what the peer reports.
+	lookupNamespace := incoming.Namespace
+	switch incoming.Kind {
+	case "AppProject", "Repository", "GPGKey":
+		lookupNamespace = r.namespace
+	case "Application":
+		// For Applications with destination-based mapping, remap the namespace
+		// if the app exists in the peer's namespace
+		if r.destinationBasedMapping {
+			if r.peerNamespace != "" && incoming.Namespace == r.peerNamespace {
```
jgwest, line 183: if DBM is enabled, shouldn't peerNamespace always be non-empty (correct me if wrong)? Thus it would be good to have a sanity test here that would log/return an error/reject if namespace was empty.

internal/resync/resync.go
```diff
@@ -522,7 +549,15 @@ func newRequestUpdateFromObject(res *unstructured.Unstructured, kind string) (*e
 		return nil, fmt.Errorf("failed to generate checksum for resource %s/%s: %w", res.GetKind(), res.GetName(), err)
 	}
 
-	reqUpdate := event.NewRequestUpdate(res.GetName(), res.GetNamespace(), kind, sourceUID, checksum[:])
+	// If the agent remapped this app's namespace, the original-namespace
+	// annotation records where the app lives on the peer. Use that so the
+	// peer can look it up in the correct namespace without guessing.
+	namespace := res.GetNamespace()
+	if originalNs, found := annotations[manager.OriginalNamespaceAnnotation]; found && kind == "Application" {
```
jgwest, line 556: In this function (correct me if wrong, I didn't check the full logic!), it looks like principal might be trusting that agent is providing the correct annotation, whereas, as discussed on Slack, principal can't necessarily trust information coming from agent (because managed agent can be compromised by attackers).

principal/event.go
```diff
@@ -219,6 +219,19 @@ func (s *Server) processApplicationEvent(ctx context.Context, agentName string,
...
+	// its own installation namespace. The OriginalNamespaceAnnotation
+	// (stamped by the agent when it remapped the app) records the correct
+	// principal-side namespace; restore it and strip the annotation.
+	// If the annotation is absent the app was never remapped, so leave
+	// incoming.Namespace as-is.
+	if s.destinationBasedMapping && agentMode.IsManaged() {
+		if originalNs, ok := incoming.Annotations[manager.OriginalNamespaceAnnotation]; ok {
+			incoming.SetNamespace(originalNs)
+			delete(incoming.Annotations, manager.OriginalNamespaceAnnotation)
+		}
```
jgwest, line 232: It appears the principal is trusting the agent here, e.g. the OriginalNamespaceAnnotation value of the Application corresponds to the actual original namespace that this came from.

As discussed on Slack, there is the potential for an attacker that has acquired access to a managed mode agent workload cluster to send false Application messages to principal, where those false messages contain an `manager.OriginalNamespaceAnnotation` annotation that is set to a different namespace than it should be. (Thus allowing attacker to potentially update/delete applications outside that agent's namespace)

agent/agent.go
```diff
@@ -139,6 +139,13 @@ type Agent struct {
 	// - The agent watches for applications in all namespaces
 	destinationBasedMapping bool
 
+	// principalNamespace is the namespace where the principal is running.
```
jgwest, line 142: (_attaching this comment to a random spot in the diff_)

We could use an E2E test for the use case of this PR, since it's a fairly fundamental agent behaviour that is also directly affecting agent security boundaries ...  (BUT, if it's a lot of work to create that, I'm fine with us handling it as part of a separate PR)

jgwest (changes requested):

gnunn1:
@chetan-rns Not ignoring this, I will be able to test on Tuesday as I'm motivated to see this work. Unfortunately my environment is being used for other things at the moment but will be free again on Tuesday.

gnunn1:
Creating an image based on the PR repo it looks like it creates the gpg keys just fine but there are a couple of errors in the logs that may confuse users?

```
time="2026-04-21T16:38:37Z" level=error msg="Failed to send request update" error="failed to construct a request update from resource argocd-gpg-keys-cm: source UID annotation not found" kind=GPGKey module=StreamEvent name=argocd-gpg-keys-cm namespace=argocd-agent serverAddr="192.168.1.5:443" uid=11597427-08e7-4b61-92f8-2632dfb51b0a
time="2026-04-21T16:38:37Z" level=error msg="Failed to send request update" agent="{\"clientID\":\"hub-prod-local\",\"mode\":\"managed\"}" error="failed to construct a request update from resource argocd-gpg-keys-cm: source UID annotation not found" event=io.argoproj.argocd-agent.event.request-resource-resync kind=GPGKey method=processIncomingEvents mode=managed module=GrpcEvent name=argocd-gpg-keys-cm namespace=argocd-agent resource_id=596bd96c-ed8d-40d4-9dcf-27c07930a278 uid=11597427-08e7-4b61-92f8-2632dfb51b0a
time="2026-04-21T16:38:37Z" level=info msg="Deleting GPG key ConfigMap" gpgKey=argocd-gpg-keys-cm method=DeleteGPGKey module=GrpcEvent
time="2026-04-21T16:38:37Z" level=info msg="Creating a new GPG key on behalf of an incoming event" gpgKey=argocd-gpg-keys-cm method=CreateGPGKey module=GrpcEvent
time="2026-04-21T16:38:37Z" level=warning msg="Could not manage GPG keys configmap argocd-gpg-keys-cm: resource argocd-gpg-keys-cm is already managed" component=GPGKeyManager
```

gnunn1:
So this still doesn't seem to be working but I'm wondering if there is something I need to configure specifically to make this work? My application doesn't get copied over and synced by the Agent and in the Agent logs I see this message:

```
time="2026-04-21T17:04:39Z" level=warning msg="namespace not allowed: argocd/hpl-banner" module=Agent
```

agent/agent.go
chetan-rns: Users should manually remove/migrate the apps from the old principal namespace. When the principal namespace changes, the apps in the old principal namespace will not be remapped to the agent namespace. The old namespace would be instead treated as a regular tenant namespace and the apps will be created on the same namespace as the principal (expected behaviour in destination-based mapping).

agent/agent.go
chetan-rns: Yeah. We must test the namespace remapping behaviour. Will address this in a different PR once we can run components in different namespaces https://github.com/argoproj-labs/argocd-agent/pull/939

gnunn1:
Re-tested enabling allowed namespaces for the Principal's namespace, on the Principal, and it is working fine now.

chetan-rns:
> So this still doesn't seem to be working but I'm wondering if there is something I need to configure specifically to make this work? My application doesn't get copied over and synced by the Agent and in the Agent logs I see this message:
> 
> ```
> time="2026-04-21T17:04:39Z" level=warning msg="namespace not allowed: argocd/hpl-banner" module=Agent
> ```

PR https://github.com/argoproj-labs/argocd-agent/pull/945 will address this issue. The workaround is to include the namespace under the allowed list

principal/event.go
chetan-rns: I've updated the annotation to be boolean-based. It doesn't expose the source namespace of the principal. If an attacker does get access to an agent cluster, they cannot escalate privileges by updating this annotation. However, they can still send unnecessary status updates.

jgwest:
Rebased to latest from `main` branch

jgwest (approved):
LGTM, thanks @chetan-rns!

PR #912, opened 2026-04-20
chore: Make logging less noisy

jannfis:

During event processing, the principal can be extremely noisy with a lot of information that really should be debug information and not being produced in normal operations.

Fixes #?

**Checklist**

## Summary by CodeRabbit

* **Chores**
  * Adjusted logging verbosity levels across various application state transitions and error handling paths for improved log clarity.

### DIFF
--- a/internal/manager/application/application.go
+++ b/internal/manager/application/application.go
@@ -167,10 +167,10 @@ func (m *ApplicationManager) Create(ctx context.Context, app *v1alpha1.Applicati
 	created, err := m.applicationBackend.Create(ctx, app)
 	if err == nil {
 		if err := m.Manage(created.QualifiedName()); err != nil {
-			log().Warnf("Could not manage app %s: %v", created.QualifiedName(), err)
+			log().Debugf("Could not manage app %s: %v", created.QualifiedName(), err)
 		}
 		if err := m.IgnoreChange(created.QualifiedName(), created.ResourceVersion); err != nil {
-			log().Warnf("Could not ignore change %s for app %s: %v", created.ResourceVersion, created.QualifiedName(), err)
+			log().Debugf("Could not ignore change %s for app %s: %v", created.ResourceVersion, created.QualifiedName(), err)
 		}
 	}
 
@@ -227,7 +227,7 @@ func (m *ApplicationManager) Upsert(ctx context.Context, app *v1alpha1.Applicati
 		return nil, err
 	}
 	if err := m.IgnoreChange(updated.QualifiedName(), updated.ResourceVersion); err != nil {
-		log().Warnf("Could not ignore change %s for app %s: %v", updated.ResourceVersion, updated.QualifiedName(), err)
+		log().Debugf("Could not ignore change %s for app %s: %v", updated.ResourceVersion, updated.QualifiedName(), err)
 	}
 	return updated, nil
 }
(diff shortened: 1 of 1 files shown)


internal/manager/application/application.go
```diff
@@ -352,17 +352,17 @@ func (m *ApplicationManager) UpdateManagedApp(ctx context.Context, incoming *v1a
 	})
 	if err == nil {
 		if updated.Generation == 1 {
-			logCtx.Infof("Created application")
+			logCtx.Debugf("Created application")
```
jgwest, line 355: Not sure I agree with switching this (and others) to debug: If by default we're not even logging the CRUD operations on the _actual resources being managed by argo cd agent_, what are we even logging?
- Now, if  you're removing this *because we are already logging it at Info elsewhere*, then that's fine (I don't have the context for the removal) 

In the case where Argo CD agent is managing a large number of clusters (and thus generating a large amount of log statements), users should switch the log level to WARN or ERROR.

internal/manager/application/application.go
```diff
@@ -167,10 +167,10 @@ func (m *ApplicationManager) Create(ctx context.Context, app *v1alpha1.Applicati
 	created, err := m.applicationBackend.Create(ctx, app)
 	if err == nil {
 		if err := m.Manage(created.QualifiedName()); err != nil {
-			log().Warnf("Could not manage app %s: %v", created.QualifiedName(), err)
+			log().Debugf("Could not manage app %s: %v", created.QualifiedName(), err)
```
jgwest, line 170: Burying an error at the debug level seems odd; do we perhaps need better differentiation of errors coming out of Manage? (for example, some at debug, others at error)?

You likely have more context here than me.

internal/manager/application/application.go
```diff
@@ -227,7 +227,7 @@ func (m *ApplicationManager) Upsert(ctx context.Context, app *v1alpha1.Applicati
 		return nil, err
 	}
 	if err := m.IgnoreChange(updated.QualifiedName(), updated.ResourceVersion); err != nil {
-		log().Warnf("Could not ignore change %s for app %s: %v", updated.ResourceVersion, updated.QualifiedName(), err)
+		log().Debugf("Could not ignore change %s for app %s: %v", updated.ResourceVersion, updated.QualifiedName(), err)
```
jgwest, line 230: As above, Warn -> Debug seems like a big step

internal/manager/application/application.go
```diff
@@ -352,17 +352,17 @@ func (m *ApplicationManager) UpdateManagedApp(ctx context.Context, incoming *v1a
...
 		if updated.Generation == 1 {
-			logCtx.Infof("Created application")
+			logCtx.Debugf("Created application")
 		} else {
-			logCtx.Infof("Updated application")
+			logCtx.Debugf("Updated application")
 		}
 		if err := m.IgnoreChange(updated.QualifiedName(), updated.ResourceVersion); err != nil {
-			logCtx.Warnf("Couldn't unignore change %s for app %s: %v", updated.ResourceVersion, updated.QualifiedName(), err)
+			logCtx.Debugf("Couldn't unignore change %s for app %s: %v", updated.ResourceVersion, updated.QualifiedName(), err)
```
jgwest, line 360: As above, burying error

internal/manager/application/application.go
```diff
@@ -352,17 +352,17 @@ func (m *ApplicationManager) UpdateManagedApp(ctx context.Context, incoming *v1a
...
 		}
 	}
 
 	if deletionTimestampChanged {
-		logCtx.Infof("deletionTimestamp of managed agent changed from nil to non-nil, so deleting Application")
+		logCtx.Debugf("deletionTimestamp of managed agent changed from nil to non-nil, so deleting Application")
```
jgwest, line 365: As above, this is important information to log as it relates to the state of the system (resources managed by Argo CD agent)

internal/manager/application/application.go
```diff
@@ -771,7 +771,7 @@ func (m *ApplicationManager) TerminateOperation(ctx context.Context, incoming *v
 		return nil, fmt.Errorf("failed to terminate operation for %s: %w", incoming.QualifiedName(), err)
 	}
 
-	logCtx.WithField(logfields.NewResourceVersion, updated.ResourceVersion).Infof("Set operation phase to Terminating")
+	logCtx.WithField(logfields.NewResourceVersion, updated.ResourceVersion).Debug("Set operation phase to Terminating")
```
jgwest, line 774: As above, this is important info to log regarding a resource managed by argocd-agent.

jgwest (changes requested):

jannfis:
Agree with your review comments. Closing in favour of https://github.com/argoproj-labs/argocd-agent/pull/914

Also, for the `IgnoreChange` semantics, I think we need to find some other way to decide whether to ignore something. I feel like this mechanism that we have is a bit odd, and as observed, not really working well. I have some ideas.

PR #913, opened 2026-04-20
feat(debug): Add a debugging routine to the principal

jannfis:

After chasing some alleged memory issues with pprof and the like, I thought it would make sense to have an optional debug printout every once in a while to have an insight into the principal.

This is more or less thought for development environments, where you might not have metrics collection and don't want to pull out pprof for every piece of information you are seeking.

Fixes #?

**Checklist**

## Summary by CodeRabbit

* **New Features**
  * Added an optional debug routine (toggleable via environment) that periodically logs runtime memory metrics, goroutine count, and per-queue depth totals (~30s).

* **Improvements**
  * Enhanced startup logging to give clearer visibility into informer synchronization progress.

### DIFF
--- a/principal/debug.go
+++ b/principal/debug.go
@@ -51,7 +51,7 @@ func (s *Server) logDebugOutput() {
 	logCtx.Infof("TotalQueueDepth=%d", totalQueueDepth)
 }
 
-// scheduleDebugRoutine launches a goroutine that prints debug information about the server to the logs every 10 seconds
+// scheduleDebugRoutine launches a goroutine that prints debug information about the server to the logs at debugInterval
 func (s *Server) scheduleDebugRoutine() {
 	ticker := time.NewTicker(debugInterval)
 	go func() {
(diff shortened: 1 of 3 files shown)


principal/debug.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		if recvLen > 0 {
+			logCtx.Infof("RecvQueueDepth(%s)=%d", queueName, recvLen)
+		}
+		if sendLen > 0 {
+			logCtx.Infof("SendQueueDepth(%s)=%d", queueName, sendLen)
+		}
+		totalQueueDepth += sendLen + recvLen
+	}
+	logCtx.Infof("TotalQueueDepth=%d", totalQueueDepth)
```
drewbailey, line 51: It may be nice to build up all the relevant debugging stats / details and log as a single line so you don't need to filter through other log statements to see the entire debug output

jannfis:
Actually, I think I'm gonna scrap this and replace with proper metrics (PR soon to come). I guess in a dev environment, one just can curl the metrics endpoint and grep out the appropriate metrics/values without a need for this extra log information.

PR #914, opened 2026-04-20
feat: Add additional logging/detail granularity to logging

### LINKED ISSUE #864: Add additional logging/detail granularity to Argo CD Agent logging

The goal of this issue is to take a holistic look at how we are logging important events/actions within Argo CD Agent.

There are three important things we should log within argocd-agent:
- Actions the agents are taking (when they create/update/delete an applications/appprojects/secrets/etc)
- Messages (events) that are being sent between agents and principal
- K8s resource events from informers (e.g. when K8s lets us know that a K8s resource/.spec/.status/.operation has been created/modified, or resource deleted).

These 3 things give us the high level picture to understand what Argo CD Agent is doing at any given moment, and why.

When logging these 3 things, there are specific details that are important to us that should be logged:
- 'Action' details
	- Log name/namespace of resource that is created/modified/deleted (plus any other useful metadata)
	- Log the actual contents of the action
		- for create: the full contents of the new resource
		- for update: the diff between old and new resource, where applicable (can use go-cmp's cmp.Diff)
		- HOWEVER, logging of create/update are likely to be very verbose (due to .spec/.status field being large). So we should add a new environment variable (something like `LOG_AGENT_FULL_DETAIL`,) that is disabled by default. When enabled, it will output the full .spec/.status content. When disabled, the .spec/.status content does not need to be logged. 
		- for delete: just the name/namespace/type of the resource

- 'Message (event)' details
	- Full contents of all the fields of the event (except for events that are large, like specupdate)
	- (For events that are large like spec/status update): Log only the non-verbose event fields, but add an env var to enable full logging of the entire event for these (See `LOG_AGENT_FULL_DETAIL` above)
	- Which agent sent it, what mode that agent was in, etc.

- 'K8s resource events from informers' details
	- Log name/namespace of resource that is created/modified/deleted (plus any other useful metadata)
	- Log the actual contents of the event
		- Same as above, etc,  for create/update we don't log the .status/.spec/.operation of the resource UNLESS `LOG_AGENT_FULL_DETAIL` is set. 

We are certainly doing much of this already, but the goal of this issue to ensure that all of the above is fully covered and consistent.

jparsai:
This PR updates logging mechanism and provides verbose logging. Introduces a --full-detail flag to enable it.

Here is how payload is attached: 
- **Actions:** Logs full JSON on create, diffs on update, name/namespace on delete. Logs about secrets don't have payload.
- **Events:** Logs event payloads for event exchanged between agent and principal. Meta-events (ACKs, heartbeats, cache updates) are ignored for verbose logging.
- **Informers:** Logs full JSON on add, diffs on update, name/namespace on delete and secrets don't have payload.

All logs include these fields for log filtering:
- log_category
- action
- resource_type
- name
- namespace
- direction
- event_target
- event_type .

Fixes https://github.com/argoproj-labs/argocd-agent/issues/864

Assisted by: Cursor

## Summary by CodeRabbit

* **New Features**
  * Introduced category-based full-detail logging with a global config, structured log fields, helpers for actions/events/informers, CloudEvent subject enrichment, informer lifecycle logs, and Secret redaction/suppressed diffs.
  * Added CLI flag to enable full-detail categories at startup.

* **Documentation**
  * Added docs describing full-detail logging, structured fields, secret-redaction, and usage examples.

* **Tests**
  * Expanded tests for full-detail parsing, logging helpers, Secret handling, and event/informer behaviors.

* **Chores**
  * Renamed agent log-format env var to ARGOCD_AGENT_LOG_FORMAT in deployment manifests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/argoproj-labs/argocd-agent/pull/914?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/argocd-agent/agent.go
+++ b/cmd/argocd-agent/agent.go
@@ -43,35 +43,36 @@ import (
 // NewAgentRunCommand returns a new agent run command.
 func NewAgentRunCommand() *cobra.Command {
 	var (
-		serverAddress       string
-		serverPort          int
-		logLevels           []string
-		logFormat           string
-		insecure            bool
-		insecurePlaintext   bool
-		rootCASecretName    string
-		rootCAPath          string
-		kubeConfig          string
-		kubeContext         string
-		namespace           string
-		agentMode           string
-		creds               string
-		tlsSecretName       string
-		tlsClientCrt        string
-		tlsClientKey        string
-		tlsMinVersion       string
-		tlsMaxVersion       string
-		tlsCipherSuites     []string
-		enableWebSocket     bool
-		metricsPort         int
-		healthzPort         int
-		enableCompression   bool
-		pprofPort           int
-		redisAddr           string
-		redisUsername       string
-		redisPassword       string
-		redisCredsDirPath   string
-		enableResourceProxy bool
+		serverAddress        string
+		serverPort           int
+		logLevels            []string
+		logFormat            string
+		fullDetailCategories []string
+		insecure             bool
+		insecurePlaintext    bool
+		rootCASecretName     string
+		rootCAPath           string
+		kubeConfig           string
+		kubeContext          string
+		namespace            string
+		agentMode            string
+		creds                string
+		tlsSecretName        string
+		tlsClientCrt         string
+		tlsClientKey         string
+		tlsMinVersion        string
+		tlsMaxVersion        string
+		tlsCipherSuites      []string
+		enableWebSocket      bool
+		metricsPort          int
+		healthzPort          int
+		enableCompression    bool
+		pprofPort            int
+		redisAddr            string
+		redisUsername        string
+		redisPassword        string
+		redisCredsDirPath    string
+		enableResourceProxy  bool
 
 		// Time interval for agent to principal ping
 		// Ex: "30m", "1h" or "1h20m10s". Valid time units are "s", "m", "h".
@@ -161,6 +162,9 @@ func NewAgentRunCommand() *cobra.Command {
 			}
 
 			agentOpts = append(agentOpts, agent.WithSubsystemLoggers(subLoggers.ResourceProxyLogger, subLoggers.RedisProxyLogger, subLoggers.GrpcEventLogger))
+
+			cmdutil.ParseFullDetail(fullDetailCategories)
+
 			if namespace == "" {
 				cmdutil.Fatal("namespace value is empty and must be specified")
 			}
@@ -324,6 +328,9 @@ func NewAgentRunCommand() *cobra.Command {
 	command.Flags().StringSliceVar(&logLevels, "log-level",
 		env.StringSliceWithDefault("ARGOCD_AGENT_LOG_LEVEL", nil, []string{"info"}),
 		"The log level to use. Comma-separated list of components in the format [<component>=]level")
+	command.Flags().StringSliceVar(&fullDetailCategories, "full-detail",
+		env.StringSliceWithDefault("ARGOCD_AGENT_FULL_DETAIL", nil, nil),
+		"Enable full detail logging for specified categories. Comma-separated list of: "+cmdutil.AvailableFullDetailCategories)
 
 	command.Flags().StringVar(&redisAddr, "redis-addr",
 		env.StringWithDefault("REDIS_ADDR", nil, "argocd-redis:6379"),
--- a/cmd/argocd-agent/agent.go
+++ b/cmd/argocd-agent/agent.go
@@ -43,35 +43,36 @@ import (
 // NewAgentRunCommand returns a new agent run command.
 func NewAgentRunCommand() *cobra.Command {
 	var (
-		serverAddress       string
-		serverPort          int
-		logLevels           []string
-		logFormat           string
-		insecure            bool
-		insecurePlaintext   bool
-		rootCASecretName    string
-		rootCAPath          string
-		kubeConfig          string
-		kubeContext         string
-		namespace           string
-		agentMode           string
-		creds               string
-		tlsSecretName       string
-		tlsClientCrt        string
-		tlsClientKey        string
-		tlsMinVersion       string
-		tlsMaxVersion       string
-		tlsCipherSuites     []string
-		enableWebSocket     bool
-		metricsPort         int
-		healthzPort         int
-		enableCompression   bool
-		pprofPort           int
-		redisAddr           string
-		redisUsername       string
-		redisPassword       string
-		redisCredsDirPath   string
-		enableResourceProxy bool
+		serverAddress        string
+		serverPort           int
+		logLevels            []string
+		logFormat            string
+		fullDetailCategories []string
+		insecure             bool
+		insecurePlaintext    bool
+		rootCASecretName     string
+		rootCAPath           string
+		kubeConfig           string
+		kubeContext          string
+		namespace            string
+		agentMode            string
+		creds                string
+		tlsSecretName        string
+		tlsClientCrt         string
+		tlsClientKey         string
+		tlsMinVersion        string
+		tlsMaxVersion        string
+		tlsCipherSuites      []string
+		enableWebSocket      bool
+		metricsPort          int
+		healthzPort          int
+		enableCompression    bool
+		pprofPort            int
+		redisAddr            string
+		redisUsername        string
+		redisPassword        string
+		redisCredsDirPath    string
+		enableResourceProxy  bool
 
 		// Time interval for agent to principal ping
 		// Ex: "30m", "1h" or "1h20m10s". Valid time units are "s", "m", "h".
@@ -165,6 +166,9 @@ func NewAgentRunCommand() *cobra.Command {
 			}
 
 			agentOpts = append(agentOpts, agent.WithSubsystemLoggers(subLoggers.ResourceProxyLogger, subLoggers.RedisProxyLogger, subLoggers.GrpcEventLogger))
+
+			cmdutil.ParseFullDetail(fullDetailCategories)
+
 			if namespace == "" {
 				cmdutil.Fatal("namespace value is empty and must be specified")
 			}
@@ -331,6 +335,9 @@ func NewAgentRunCommand() *cobra.Command {
 	command.Flags().StringSliceVar(&logLevels, "log-level",
 		env.StringSliceWithDefault("ARGOCD_AGENT_LOG_LEVEL", nil, []string{"info"}),
 		"The log level to use. Comma-separated list of components in the format [<component>=]level")
+	command.Flags().StringSliceVar(&fullDetailCategories, "full-detail",
+		env.StringSliceWithDefault("ARGOCD_AGENT_FULL_DETAIL", nil, nil),
+		"Enable full detail logging for specified categories. Comma-separated list of: "+cmdutil.AvailableFullDetailCategories)
 
 	command.Flags().StringVar(&redisAddr, "redis-addr",
 		env.StringWithDefault("REDIS_ADDR", nil, "argocd-redis:6379"),
--- a/cmd/argocd-agent/agent.go
+++ b/cmd/argocd-agent/agent.go
@@ -368,7 +368,7 @@ func NewAgentRunCommand() *cobra.Command {
 		"INSECURE: Do not verify Redis TLS certificate")
 
 	command.Flags().StringVar(&logFormat, "log-format",
-		env.StringWithDefault("ARGOCD_PRINCIPAL_LOG_FORMAT", nil, "text"),
+		env.StringWithDefault("ARGOCD_AGENT_LOG_FORMAT", nil, "text"),
 		"The log format to use (one of: text, json)")
 	command.Flags().BoolVar(&insecure, "insecure-tls",
 		env.BoolWithDefault("ARGOCD_AGENT_TLS_INSECURE", false),
--- a/cmd/cmdutil/log.go
+++ b/cmd/cmdutil/log.go
@@ -21,6 +21,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/argoproj-labs/argocd-agent/internal/logging"
 	"github.com/sirupsen/logrus"
 	"github.com/sirupsen/logrus/hooks/writer"
 )
@@ -191,4 +192,34 @@ func ParseLogLevels(input []string, ss *SubSystemLoggers) {
 			logrus.Warnf("an invalid subsystem %s was specified. subsystems are %s, skipping", split[0], AvailableSubSystems)
 		}
 	}
+}
+
+const AvailableFullDetailCategories = "all, actions, events, informers"
+
+// ParseFullDetail parses a comma-separated list of category names and sets
+// the global full detail logging configuration. The special value "all"
+// enables every category.
+func ParseFullDetail(input []string) {
+	cfg := logging.FullDetailConfig{}
+	for _, e := range input {
+		category := strings.TrimSpace(strings.ToLower(e))
+		if category == "" {
+			continue
+		}
+		switch category {
+		case "all":
+			cfg.Actions = true
+			cfg.Events = true
+			cfg.Informers = true
+		case "actions":
+			cfg.Actions = true
+		case "events":
+			cfg.Events = true
+		case "informers":
+			cfg.Informers = true
+		default:
+			logrus.Warnf("an invalid full-detail category %q was specified. Available categories are: %s, skipping", category, AvailableFullDetailCategories)
+		}
+	}
+	logging.SetFullDetailConfig(cfg)
 }
--- a/internal/informer/informer.go
+++ b/internal/informer/informer.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/argoproj-labs/argocd-agent/internal/filter"
+	"github.com/argoproj-labs/argocd-agent/internal/logging"
 	"github.com/argoproj-labs/argocd-agent/internal/metrics"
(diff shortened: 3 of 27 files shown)


cmd/cmdutil/log.go
```diff
@@ -192,3 +193,33 @@ func ParseLogLevels(input []string, ss *SubSystemLoggers) {
...
+			cfg.Actions = true
+		case "events":
+			cfg.Events = true
+		case "informers":
+			cfg.Informers = true
+		default:
+			logrus.Warnf("an invalid full-detail category %q was specified. Available categories are: %s, skipping", category, AvailableFullDetailCategories)
+		}
+	}
+	logging.SetFullDetailConfig(cfg)
```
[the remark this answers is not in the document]
jparsai: I dont think it is valid as `ParseFullDetail()` receives the `[]string` where each element is already a single category so `[]string{"actions,events"}` will never treated as one invalid category.

internal/informer/informer.go
```diff
@@ -202,6 +203,7 @@ func (i *Informer[T]) installEventHandlers() error {
 				if !ok {
 					return
 				}
+				logging.LogInformerAdd(i.logger, obj)
```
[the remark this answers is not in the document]
jparsai: It is not valid, we have check for secrets `marshalResource()` and secret payload is not attached with logs.

internal/logging/logging.go
```diff
@@ -344,3 +370,276 @@ func SelectLogger(l *CentralizedLogger) *CentralizedLogger {
...
+		logfields.Name:         name,
+		logfields.Namespace:    namespace,
+	})
+
+	// Include the full resource in the log if full detail is enabled
+	if fullDetailConfig.Actions {
+		logCtx = logCtx.WithField(logfields.Detail, marshalResource(obj))
+	}
+	logCtx.WithError(err).Errorf("Error performing action %s on %s %s/%s", action, resourceType, namespace, name)
+}
```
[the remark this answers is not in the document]
jparsai: I dont think this is valid. Secrets contain confidential data which is already removed from payload. Your concern is about rest of the resources but they dont have any confidential info and they are anyway visible to users in cluster.

internal/logging/logging.go
```diff
@@ -344,3 +370,276 @@ func SelectLogger(l *CentralizedLogger) *CentralizedLogger {
...
+
+// resourceType returns the lowercase type name of a K8s runtime object (e.g. "configmap", "application").
+func resourceType(obj any) string {
+	return strings.ToLower(reflect.TypeOf(obj).Elem().Name())
+}
+
+func isSecret(obj any) bool {
+	_, ok := obj.(*corev1.Secret)
+	return ok
+}
```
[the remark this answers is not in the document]
jparsai: This can be ignored, it is nice-to-have safety check, but in current implementation obj is always a valid non-nil pointer.

jannfis:
Just a quick note, after glimpsing over the code: We need to be careful here not to add too much additional processing unconditionally (i.e. JSON marshalling etc) on the hot paths.

jparsai:
> Just a quick note, after glimpsing over the code: We need to be careful here not to add too much additional processing unconditionally (i.e. JSON marshalling etc) on the hot paths.

Yes Jann, I agree. In this implementation, expensive operations like JSON marshalling and `cmp.Diff` are only performed when `--full-detail` is enabled. These are limited to action/informer create and update operations. If verbose logging seems superfluous at some place, I would remove/change it.

cmd/argocd-agent/agent.go
```diff
@@ -356,7 +363,7 @@ func NewAgentRunCommand() *cobra.Command {
 		"INSECURE: Do not verify Redis TLS certificate")
 
 	command.Flags().StringVar(&logFormat, "log-format",
-		env.StringWithDefault("ARGOCD_PRINCIPAL_LOG_FORMAT", nil, "text"),
+		env.StringWithDefault("ARGOCD_AGENT_LOG_FORMAT", nil, "text"),
```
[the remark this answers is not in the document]
jparsai: @jgwest should we provide migration path? CLI command is same but env variable name is changed. There was a name missmatch in original logging implementation, which is raised here https://github.com/argoproj-labs/argocd-agent/pull/914#discussion_r3256454036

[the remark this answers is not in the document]
jgwest: @jparsai since it's just log format, and not something more serious, so let's not bother with migration and/or preserving the old name. If users require a specific log format, it is easy for them to detect the change, and it is easy for them to fix it themselves.

It seems like coderabbit is also making a comment that the docs still reference the old variable (but it could be referring to the principal), so it would  be good to double check we've updated the appropriate doc references (if not already done).

(Also need to make sure we update the any references to the new value in argocd-operator)

internal/logging/logging.go
```diff
@@ -344,3 +370,276 @@ func SelectLogger(l *CentralizedLogger) *CentralizedLogger {
...
+	if strings.HasPrefix(evType, prefix) {
+		return evType[len(prefix):]
+	}
+	return evType
+}
+
+// isMetaEvent checks if the event is a meta event
+func isMetaEvent(ev *cloudevents.Event) bool {
+	target := ev.DataSchema()
+	return target == "eventProcessed" ||
```
jgwest, line 607: I think it would actually be useful to log `eventProcessed` events, but maybe only if the logging level is at debug or trace. Is that possible?

internal/manager/application/application.go
```diff
@@ -352,10 +355,8 @@ func (m *ApplicationManager) UpdateManagedApp(ctx context.Context, incoming *v1a
 		return patch, err
 	})
 	if err == nil {
-		if updated.Generation == 1 {
-			logCtx.Infof("Created application")
-		} else {
-			logCtx.Infof("Updated application")
+		if updated.Generation > 1 {
+			logging.LogActionUpdate(logCtx, "application", incoming, updated)
```
jgwest, line 359: The `if err := m.applicationBackend.Delete(ctx, incoming.Name, incoming.Namespace, ptr.To(backend.DeletePropagationForeground)); err != nil {` line in this function should be logged as an action (correct me if wrong)

internal/manager/application/application.go
```diff
@@ -28,6 +28,7 @@ import (
 
 	"github.com/argoproj-labs/argocd-agent/internal/backend"
```
jgwest, line 29: `SetOperation`/`UpdateOperation`/`UpdateStatus` in this file doesn't call `LogActionUpdate` AFAICT
`Delete` doesn't log (via LogActionUpdate) the update of Application to remove finalizers

internal/manager/appproject/appproject.go
```diff
@@ -154,6 +154,7 @@ func (m *AppProjectManager) Create(ctx context.Context, project *v1alpha1.AppPro
 func createAppProject(ctx context.Context, m *AppProjectManager, project *v1alpha1.AppProject) (*v1alpha1.AppProject, error) {
```
jgwest, line 154: In `Delete` function in this file, we are calling `RemoveFinalizers` (which will update the resource), but we are not logging that update via LogActionUpdate

cmd/argocd-agent/agent.go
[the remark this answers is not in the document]
jparsai: Yes @jgwest, I agree, this is low risk.

All docs are already updated, coderabbit is asking to document this name change as well which is not required.

And about argocd-operator, it is already using correct environment name (https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocdagent/agent/deployment.go#L387), so no actin needed there.

internal/logging/logging.go
```diff
@@ -344,3 +370,276 @@ func SelectLogger(l *CentralizedLogger) *CentralizedLogger {
...
+		logfields.LogCategory: "events",
+		logfields.Direction:   "send",
+		logfields.EventTarget: target,
+		logfields.EventType:   ev.Type(),
+	})
+
+	logCtx = parseEventSubject(logCtx, ev)
+
+	// Include the event data in the log if full detail is enabled and event data is not empty
+	if fullDetailConfig.Events && hasEventData(ev.Data()) {
```
jgwest: The original issue says that we should log "full contents of all the fields of the event (except for events that are large, like specupdate)". AFAICT the PR does not currently do this: we are only logging event fields if ARGOCD_AGENT_FULL_DETAILS is set.

internal/logging/logging.go
```diff
@@ -344,3 +370,276 @@ func SelectLogger(l *CentralizedLogger) *CentralizedLogger {
...
+		logCtx = logCtx.WithField(logfields.Detail, string(ev.Data()))
+	}
```
jgwest, line 474: FYI:
- It appears we are unintentionally logging secret data when output `event received: repository` for repository secrets.
    - For example: 06:28:04    `agent-managed | time="2026-05-20T06:28:04-04:00" level=info msg="Event received: repository delete" clientAddr="127.0.0.1:8443" detail="{\"kind\":\"Secret\",\"apiVersion\":\"v1\",\"metadata\":{\"name\":\"sample\",\"namespace\":\"argocd\",\"uid\":\"6fa794be-3eac-4347-bc7b-a9a718f26d01\",\"resourceVersion\":\"4195\",\"creationTimestamp\":\"2026-05-20T10:28:01Z\",\"labels\":{\"argocd.argoproj.io/secret-type\":\"repository\"},\"managedFields\":[{\"manager\":\"e2e.test\",\"operation\":\"Update\",\"apiVersion\":\"v1\",\"time\":\"2026-05-20T10:28:01Z\",\"fieldsType\":\"FieldsV1\",\"fieldsV1\":{\"f:data\":{\".\":{},\"f:project\":{},\"f:url\":{}},\"f:metadata\":{\"f:labels\":{\".\":{},\"f:argocd.argoproj.io/secret-type\":{}}},\"f:type\":{}}}]},\"data\":{\"project\":\"JGW_REMOVED_SECRET_DATA_HERE\",\"url\":\"JGW_ALSO_REMOVED_SECRET_DATA_HERE\"},\"type\":\"Opaque\"}" direction=recv event_id=sample_6fa794be-3eac-4347-bc7b-a9a718f26d01_4195 event_target=repository event_type=io.argoproj.argocd-agent.event.delete log_category=events module=StreamEvent name=sample namespace=argocd resource_id=sample_6fa794be-3eac-4347-bc7b-a9a718f26d01 type=io.argoproj.argocd-agent.event.delete`
    - likewise for create/update events for repo secrets
 
- We MAY be logging secrets if secret data is contained with redis-response or get-response (I say MAY because I haven't verified if this is true, can you verify?)
 
- _Under no circumstances should secret/confidential data be logged to console output_
    - (we might want to add an additional env var in the future like UNSAFE_LOG_SECRETS_TO_CONSOLE so users can explicitly opt into logging that data, but that is a separate PR)
    - The fact that secret data is only logged when `ARGOCD_AGENT_FULL_DETAILS` is set is not a sufficient defense

jgwest (changes requested):
Additional comments:
- In `principal/callbacks.go`, in `updateAppProjectCallback`, we are calling `s.projectManager.RemoveFinalizers(s.ctx, new)`, which will modify a resource, which we should log as an update resource action.

internal/logging/logging.go
```diff
@@ -344,3 +370,288 @@ func SelectLogger(l *CentralizedLogger) *CentralizedLogger {
...
+	// Meta events are logged at debug level to avoid noise.
+	if isMetaEvent(ev) {
+		logCtx.Debugf("Event sent: %s %s", target, action)
+		return
+	}
+
+	// Events for sensitive resources are never verbose logged to avoid leaking credentials.
+	if hasEventData(ev.Data()) && !isSensitiveEvent(ev) {
+		logCtx = logCtx.WithField(logfields.Detail, string(ev.Data()))
+	}
```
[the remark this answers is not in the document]
jparsai: This is expected. Event logs should always have payload as asked in https://github.com/argoproj-labs/argocd-agent/pull/914#discussion_r3273438972

internal/logging/logging.go
jparsai: I updated it, I was masking secret data, but missed repository events.

internal/logging/logging.go
```diff
@@ -344,3 +370,288 @@ func SelectLogger(l *CentralizedLogger) *CentralizedLogger {
...
+	logCtx = parseEventSubject(logCtx, ev)
+
+	// Meta events are logged at debug level to avoid noise.
+	if isMetaEvent(ev) {
+		logCtx.Debugf("Event sent: %s %s", target, action)
+		return
+	}
+
+	// Events for sensitive resources are never verbose logged to avoid leaking credentials.
+	if hasEventData(ev.Data()) && !isSensitiveEvent(ev) {
```
jgwest: Previous commits appear to have regressed functionality: we are now logging full detail regardless of whether `ARGOCD_AGENT_FULL_DETAIL`/`ARGOCD_PRINCIPAL_FULL_DETAIL` are set.

My suggestion is to go back to the previous verison of this logic where we would log details only if the above env var/parameters are set, and we can figure out additional details on a separate PR.

internal/logging/logging.go
```diff
@@ -344,3 +370,288 @@ func SelectLogger(l *CentralizedLogger) *CentralizedLogger {
...
+	return len(s) > 0 && s != "{}" && s != "null"
+}
+
+// isSensitiveEvent returns true if the event may carry sensitive data.
+// Repository events carry Secret objects directly. Redis and resource events
+// carry opaque payloads that could contain secrets (cached manifests, raw K8s
+// API responses), so they are treated as potentially sensitive.
+func isSensitiveEvent(ev *cloudevents.Event) bool {
+	switch ev.DataSchema() {
+	case "repository", "redis", "resource":
```
jgwest, line 624: Can replace these non-constant strings with reference to existing constants

internal/logging/logging.go
```diff
@@ -344,3 +370,288 @@ func SelectLogger(l *CentralizedLogger) *CentralizedLogger {
...
+	if strings.HasPrefix(evType, prefix) {
+		return evType[len(prefix):]
+	}
+	return evType
+}
+
+// isMetaEvent checks if the event is a meta event
+func isMetaEvent(ev *cloudevents.Event) bool {
+	target := ev.DataSchema()
+	return target == "eventProcessed" ||
```
jgwest, line 607: Can replace these constants with references to existing constants

jgwest (changes requested):

internal/logging/logging.go
jparsai: These constants are defined in `internal/event/event.go` but `internal/event` also imports `internal/logging`. This is creating circular dependency.

internal/logging/logging.go
jparsai: Same as above.

internal/logging/logging.go
jparsai: Fixed it, Now events are always logged in verbose except for spec/status update events. These events are verbose only of full details is enabled.

internal/logging/logging.go
jgwest: This is a solvable problem, for example, moving constants into their own package (as one possible option). We'll look at this on a separate PR.

internal/logging/logging.go
```diff
@@ -344,3 +370,301 @@ func SelectLogger(l *CentralizedLogger) *CentralizedLogger {
...
+	logCtx = logCtx.WithFields(logrus.Fields{
+		logfields.LogCategory:  "actions",
+		logfields.Action:       "create",
+		logfields.ResourceType: resourceType,
+		logfields.Name:         name,
+		logfields.Namespace:    namespace,
+	})
+
+	// Include the full resource in the log if full detail is enabled
+	if fullDetailConfig.Actions {
```
jgwest, line 387: Do we need the `isSecret(newObj)` for LogActionCreate/Update/Error? For example, for when we are creating/updating Secrets?

internal/logging/logging.go
```diff
@@ -344,3 +370,301 @@ func SelectLogger(l *CentralizedLogger) *CentralizedLogger {
...
+
+	logCtx = logCtx.WithFields(logrus.Fields{
+		logfields.LogCategory:  "informers",
+		logfields.Action:       "create",
+		logfields.ResourceType: resType,
+		logfields.Name:         name,
+		logfields.Namespace:    namespace,
+	})
+
+	if fullDetailConfig.Informers {
```
jgwest, line 539: For LogInformerAdd/Update, do we likewise need isSecret(newObj)?

jgwest (changes requested):

internal/logging/logging.go
jparsai: Actually it is already checked in `marshalResource()`

```
if fullDetailConfig.Actions {
	logCtx = logCtx.WithField(logfields.Detail, marshalResource(obj))
}
```
`marshalResource()` internally calls `isSecret(obj)` and `returns "<redacted: Secret>"` if object is of Secret type

internal/logging/logging.go
jparsai: Same as https://github.com/argoproj-labs/argocd-agent/pull/914#discussion_r3297884880

internal/logging/logging.go
jgwest: Ah, forgot about that, thanks!

internal/logging/logging.go
jparsai: Yes, that would require some changes in structure. I will keep a note of it and address in another PR soon.

jgwest (approved):
LGTM, thanks @jparsai!

PR #926, opened 2026-04-27
fix: Respect managed app finalizer updates

drewbailey:
Managed agent updates use JSON patch when the backend supports it. The patch target included incoming finalizers, but the source omitted existing finalizers, so removing a finalizer on the principal produced no diff and left the agent-side finalizer intact.

Include existing finalizers in the patch source comparison so additions and removals are emitted, and add a regression test for removing finalizers with a patch-backed managed update.

Fixes #?

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how the system handles application finalizers during updates to ensure proper state consideration.

* **Tests**
  * Added validation test for finalizer removal during application update operations.

### DIFF
--- a/internal/manager/application/application.go
+++ b/internal/manager/application/application.go
@@ -340,6 +340,7 @@ func (m *ApplicationManager) UpdateManagedApp(ctx context.Context, incoming *v1a
 			ObjectMeta: v1.ObjectMeta{
 				Annotations: existing.Annotations,
 				Labels:      existing.Labels,
+				Finalizers:  existing.Finalizers,
 			},
 			Spec:      existing.Spec,
 			Operation: existing.Operation,
--- a/internal/manager/application/application_test.go
+++ b/internal/manager/application/application_test.go
@@ -321,6 +321,36 @@ func Test_ManagerUpdateManaged(t *testing.T) {
 		require.Equal(t, "principal-B", updated.Annotations[manager.PrincipalUIDAnnotation])
 	})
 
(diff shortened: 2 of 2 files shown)


drewbailey:
From investigating the history around this behavior,

Timeline I found:

- Initial implementation did not sync finalizers in managed updates at all.
- PR #89 / b29451c added finalizer handling for Application deletion support. It added incoming.Finalizers to the managed update path and to the patch target, but did not add existing.Finalizers to the patch source. That made additions visible to jsondiff, but made removals invisible.
- That same PR also added principal-side code that stripped Application finalizers on update. So there was an older “remove finalizers from principal” behavior, but it was deletion-flow related, not a patch-diff design.
- PR #462 / 128e239 explicitly says cascade delete was broken because “the Argo CD Application deletion .metadata.finalizer was being removed by principal.” That PR removed the Application callback stripping, added managed deletion handling, and e2e coverage that finalizers plus deletion timestamp sync to the managed agent.
- It did not update the patch source object, so the add/cascade path worked, but plain finalizer removal still didn’t.

I think this was not intentional?

- UpdateManagedApp’s non-patch path already assigns existing.Finalizers = incoming.Finalizers.
- Existing unit test said “Finalizers must be in sync.”
T- he dedicated RemoveFinalizers patch path already compares against existing.Finalizers.
- The cascade-delete e2e tests explicitly expect finalizers to sync to the managed agent.

jannfis (approved):
Good catch. Verified behaviour pre and post this change manually.

LGTM.

PR #927, opened 2026-04-27
feat: source uid mismatch policy

### LINKED ISSUE #905: configurable source-uid mismatch policy

Today if there is a source-uid mismatch in managed agents, the resource is deleted and then re-created. This behavior and the edge cases it handles are covered in https://github.com/argoproj-labs/argocd-agent/blob/main/docs/proposals/Argo-CD-Agent-Sync-Proposal.md.

Managed-resource reconciliation paths treat `source-uid` mismatch as a destructive replace operation by deleting the existing local resource and recreating it from the incoming event.
Source delete handling is also immediate in places where the source is treated as authoritative, including resync paths that translate `NotFound` on the source into a delete event on the peer.

This creates a potentially large blast radius for operators who may want safer semantics for identity mismatch or source-driven deletes, especially for high-risk resources such as `argocd-gpg-keys-cm`.

Add an explicit deletion and identity-mismatch policy model so operators can choose what level of safety/correctness behavior per resource type instead of always using delete-and-recreate semantics.

Examples of desired behavior:
- configurable handling for `source-uid` mismatch such as `recreate`, `upsert`, or `block`

- Current behavior

drewbailey:

Add a --source-uid-mismatch-policy flag (default: recreate) that controls
agent behavior when a source-UID mismatch is detected on an incoming managed
resource. The previous hardcoded behavior (delete + recreate) is preserved as
the default.

Policies:
- recreate: delete the existing resource, then create the incoming one (default)
- upsert: update the existing resource in-place without deleting it

The policy can also be overridden per-resource via annotation:
  argocd.argoproj.io/source-uid-mismatch-policy: upsert

Annotation takes precedence over the global flag. Unknown annotation values
fall back to the global policy with a warning log. Applies to all four managed
resource types: Application, AppProject, Repository, and GPGKey.

Motivation: destructive replacement on source-UID mismatch is unsafe for
sensitive resources like argocd-gpg-keys-cm. The upsert policy allows
operators to opt into non-destructive reconciliation where appropriate.

Fixes #905

**Checklist**

## Summary by CodeRabbit

* **New Features**
  * Added configurable source-UID mismatch resolution for managed resources: choose between recreating (`recreate`, default) or updating in-place (`upsert`)
  * Support per-resource policy override via annotation

* **Documentation**
  * Added sync protocol and configuration reference documentation for source-UID mismatch handling


jannfis (approved):
LGTM, thanks!

PR #928, opened 2026-04-27
ci: harden Helm chart validation pipeline

C4tWithShell:

The Helm chart validation in CI had three gaps that let real bugs slip
through and one supply-chain weakness:

1. `required`, type, and enum constraints in `values.schema.json` were never enforced.
 `hack/validate-values-schema.sh` only walks paths in `values.yaml` and checks each one exists in the schema. That direction catches *unknown* keys, but completely misses:
   - a required top-level block being deleted from `values.yaml`,
   - a value with the wrong type (e.g. `replicaCount: "1"` instead of `1`),
   - a value violating an `enum`.
   These would surface only at chart install time, not in CI.

2. Empty containers in `values.yaml` were skipped. The path extractor used `jq … paths(scalars)`, so `tolerations: []`  produced no path at all and could not be compared against the schema.

3. `yq` was downloaded with `wget` and no integrity check. A compromised release asset would be installed silently. The script also accepted *any* `yq` on PATH, including the Python `yq` (kislyuk/yq), which uses different syntax and produces wrong output without erroring — so a developer running the script locally with the wrong `yq` got a green pass on broken values.

4. `set -e` doesn't propagate failures from `$(…)` substitutions in bash. A broken `yq` pipeline could exit non-zero while the caller silently continued with empty output.

### `hack/validate-values-schema.sh`
- Add a `helm lint` invocation against the chart, which evaluates `values.yaml` against `values.schema.json` and enforces `required`, `type`, and `enum`. This is complementary to the existing path walker (which still catches unknown-key drift in the other direction).
- Add `helm` to `required_binaries`.
- Detect and reject non-mikefarah `yq` implementations up front, with a clear error message pointing at the right install.
- Enable `shopt -s inherit_errexit` so command-substitution failures propagate.
- Switch the YAML path extractor from `paths(scalars)` to `paths` (filtering numeric array indices), so empty arrays/objects in `values.yaml` still produce a path that gets compared to the schema.

### `.github/workflows/ci.yaml` (`validate-helm-charts` job)
- Replace `wget` + chmod with `curl --fail --retry` + `sha256sum -c` to verify the yq binary against a pinned SHA-256.
- Pin `YQ_VERSION` and `YQ_SHA256` as step env vars.
- Add an `Install Helm` step using `azure/setup-helm` pinned to a SHA (`v5.0.0`), with `helm` pinned to `v3.20.2`.

N/A

Ran `hack/validate-values-schema.sh` locally against the chart in this repo and exercised three cases:

| Scenario                                          | Expected | Result   |
| ------------------------------------------------- | -------- | -------- |
| Unmodified `values.yaml`                          | pass     | ✓ pass   |
| Required `image` block deleted from `values.yaml` | fail (helm lint) | ✓ fail with `missing property 'image'` |
| Unknown key `unknownField` added to `values.yaml` | fail (path walker) | ✓ fail with `unknownField missing in schema` |
| Python `yq` on PATH instead of mikefarah `yq`     | early exit with guidance | ✓ exits 1 with install link |

No changes to runtime code or chart contents.

**Checklist**

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -277,12 +277,27 @@ jobs:
         with:
           persist-credentials: false
       
-      - name: Install required packages
+      - name: Install yq
+        env:
+          YQ_VERSION: v4.53.2
+          # SHA-256 of yq_linux_amd64 from
+          # https://github.com/mikefarah/yq/releases/download/v4.53.2/checksums
+          YQ_SHA256: d56bf5c6819e8e696340c312bd70f849dc1678a7cda9c2ad63eebd906371d56b
         run: |
-          # Install yq
-          YQ_VERSION="v4.44.3"
-          sudo wget -qO /usr/local/bin/yq "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_amd64"
-          sudo chmod +x /usr/local/bin/yq
+          set -euo pipefail
+          curl --fail --silent --show-error --location --retry 3 \
+            --output /tmp/yq \
+            "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_amd64"
+          echo "${YQ_SHA256}  /tmp/yq" | sha256sum --check --status
+          sudo install -m 0755 /tmp/yq /usr/local/bin/yq
+          rm -f /tmp/yq
+          yq --version
+
+      - name: Install Helm
+        uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
+        with:
+          version: v3.20.2
+
       - name: Validate schema
         working-directory: ${{ github.workspace }}
         run: |
--- a/hack/validate-values-schema.sh
+++ b/hack/validate-values-schema.sh
@@ -14,6 +14,10 @@
 # limitations under the License.
 
 set -euo pipefail
+# Propagate failures from command substitutions (e.g., $(yq ...)) into the
+# parent shell. Without this, a broken yq pipeline could exit non-zero while
+# the caller silently continues with an empty value.
+shopt -s inherit_errexit 2>/dev/null || true
 
 # Script directory
 SCRIPT_DIR="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -277,12 +277,27 @@ jobs:
...
+          curl --fail --silent --show-error --location --retry 3 \
+            --output /tmp/yq \
+            "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_amd64"
+          echo "${YQ_SHA256}  /tmp/yq" | sha256sum --check --status
+          sudo install -m 0755 /tmp/yq /usr/local/bin/yq
+          rm -f /tmp/yq
+          yq --version
+
+      - name: Install Helm
+        uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
```
jannfis, line 297: Validated checksum matches tag

.github/workflows/ci.yaml
```diff
@@ -277,12 +277,27 @@ jobs:
         with:
           persist-credentials: false
       
-      - name: Install required packages
+      - name: Install yq
+        env:
+          YQ_VERSION: v4.53.2
+          # SHA-256 of yq_linux_amd64 from
+          # https://github.com/mikefarah/yq/releases/download/v4.53.2/checksums
+          YQ_SHA256: d56bf5c6819e8e696340c312bd70f849dc1678a7cda9c2ad63eebd906371d56b
```
jannfis, line 285: Validated checksum against binary

jannfis (commented):
Thanks @C4tWithShell !

I have one question. I could not make sense of the `checksums` file published by `yq`. Using `sha256sum -c checksums` doesn't work. So I was wondering how to proceed with when we update `yq` - I feel like downloading the binary and running `sha256sum` on it really doesn't make much sense.

C4tWithShell:
@jannfis, yeah,  it is a tricky one

It's a custom space-separated table with one row per asset and many hash algorithms per row:
```
<filename> <CRC32> <MD4> <MD5> <SHA1> <TIGER> <TTH> <BTIH> <ED2K> <AICH> \
           <WHIRLPOOL> <RIPEMD-160> <GOST94> <GOST94-CRYPTOPRO> <HAS-160> \
           <GOST12-256> <GOST12-512> <SHA-224> <SHA-256> <SHA-384> <SHA-512>
```

The column layout is documented in the companion file `checksums_hashes_order `published in the same release. SHA-256 is the 18th hash, so column 19 of the line (column 1 is the filename).

That means the SHA-256 we pin is reachable directly from upstream with one awk:

`curl -fsSL "https://github.com/mikefarah/yq/releases/download/v4.53.2/checksums" \
  | awk '$1 == "yq_linux_amd64" {print $19}'
→ d56bf5c6819e8e696340c312bd70f849dc1678a7cda9c2ad63eebd906371d56b

Or you can simply check hash in yq release 

image



I could write documentation on the upgrade process for it, but I'm not sure where to put it

jannfis:
@C4tWithShell Ah, thanks. Yeah, that makes sense. For whatever reason they chose this weird format. I'm fine with it, now that I know.

jannfis (approved):
LGTM

Thanks!

PR #937, opened 2026-05-05
feat: Allow buffering of Application events at informer to reduce event volume

jgwest:
This PR introduces a new, optional, configurable 'informer event buffer' layer that sits between K8s informer and the argo cd agent `Application` resource informers event handlers. 
- Only buffers and de-duplicates Application update events (e.g. not add/delete)
- Does not touch other resource types (AppProject, etc)

This new layer ensures that in an X second period (for example, 10 second), the argo cd agent 'update' event handlers will be called at most once every X seconds.

This is an example of how the algorithm works:
- @0 seconds) Application informer informs us of an Application Event (for example, .status change)
    - Agent's Application update informer handler is immediately called. (Since we have not called the informer handler within the last 10 seconds)
    - We set the 'next available opportunity to send event' window to @10 seconds.

- @1 second) Application informer informs us of another Application Event
    - We DON'T call the Application informer handler.
    - We instead put this event at the top of the buffer.

- @6 seconds) Application informer informs us of another Appliation Event
	- Same as above, we don't call the Appliation infomer handler.
	- This event replaces the event from @1 second in the buffer (since the one at @1 second is now stale).

- @10 seconds) We detect that the 10 seconds has elaped, and send the message from @6 seconds.

- @20 seconds) Application informer informs us of an Application Event (for example, .status change)
	- Since we have not seen any events for the last 10 seconds, we call Application informer handler immediately. 

This ensures that principal (downstream for Application .status update) is not overloaded by large numbers of Application .status changes at once. For example, as can occur on initial deploy of an Application, or less commonly, when Argo CD has been misconfigured such that it is battling another controller for ownership of a resource. 

To test the efficacy, I configured the layer with X=10 seconds (send Application update events at most once every 10 seconds), and then ran the standard E2E tests.
- I observed an 70-80% reduction in Application informer update calls, with the tests continuing to pass.

Other details:
- Only Application resources are handled by this PR. (These are the most impacted by large number of .status updates)
- The semantics of this new layer preserve the same semantics as `SharedInformer`:
	- As the SharedInformer docs note:  "For a given SharedInformer and client, the notifications are delivered sequentially.  For a given SharedInformer, client, and object ID, the notifications are delivered in order."
	- When informer calls us, we return quickly (the PR implementation is always equal to or faster than current implementation)
	- Each Application instance will be processed by at most 1 goroutine (e.g. no concurrent processing of different events from same Application entity)    
    - Resources ordering same as SharedInformer: creates will be processed first, then updates (in chronological order, never concurrently), then delete.
    		- BUT: If we see a 'delete', it will short circuit any remaining updates that are still to process.

Fixes N/A

**Checklist**

## Summary by CodeRabbit

* **New Features**
  * Added configurable throttling for Kubernetes Application informer *update* events via `--buffer-informer-event-interval` / `ARGOCD_AGENT_BUFFER_INFORMER_EVENT_INTERVAL` (default `0` = no buffering).

* **Logging**
  * Extended `--log-level` / `ARGOCD_AGENT_LOG_LEVEL` component support with `informer-event-buffer` to control verbosity for the informer buffering subsystem.

* **Bug Fixes**
  * Improved handling for already-managed app creation by routing it through the update path instead of ignoring it.


jgwest:
@coderabbitai review

jparsai (approved):
LGTM also tested locally, it works as expected.
Thanks!!

jgwest:
Thanks @jparsai!

PR #938, opened 2026-05-05
feat: deduplicate unsent event writer events for spec/statusupdate

jgwest:
- The event writer is responsible for ensuring that outbound messages are written (and acked) from principal to agent.
- This PR observes that in at least a couple cases (spec update and status update), it is straightforward to remove stale events from the queue, in order to avoid spending bandwidth/cpu/memory on transmitting events that are already known to be stale at the source.
    - This PR only de-duplicates unsent events; we never throw away a previously sent event (even though it will potentially be more stale then some events in the unsent queue)
- This PR also increases the exponential backoff to more aggressively ramp up, tuning it to avoid a large number of messages in the worst case scenario (initial cluster+application initialization)
    - In a previous PR I made the backoff much more gradual, but in retrospect I think that is now incorrect give the possible failure case we will see.
- This PR also increases the max number of retries to infinite, such that we never give up on sending a non-stale message (or at least until agent OS process restart)
    - This is done for correctness: we really shouldn't be dropping message, as there is no mechanism to get that data after the fact (e.g. without an agent/principal process restart)
    - In the future/a separate PR, we should instead implement a 'max ttl' for messages which is configurable by users, based on the max length of time they expect agents to be reasonably disconnected from principal.
    - We can also more quickly drop messages that are known to be ephemeral, like redis/request proxy responses.
- Adds a randomize to avoid disproportionately favouring somes keys in the map over others

Fixes N/A

**Checklist**

## Release Notes

* **Bug Fixes**
  * Improved event retry/backoff behavior for more reliable delivery.
  * Fixed unsent-event queue handling to safely inspect and remove the correct next item.
  * Enhanced deduplication of repeated status/spec updates to reduce redundant processing.
  * Randomized the order of waiting resources processed each cycle to help prevent starvation.

* **Tests**
  * Added focused test coverage for deduplicating event message items across multiple scenarios.

### DIFF
--- a/internal/event/event_writer.go
+++ b/internal/event/event_writer.go
@@ -2,6 +2,8 @@ package event
 
 import (
 	"context"
+	"math"
+	"math/rand/v2"
 	"strings"
 	"sync"
 	"time"
@@ -19,7 +21,7 @@ import (
 
 const (
 	// maxEventRetries is the maximum number of times an event will be retried before giving up.
-	maxEventRetries = 12
+	maxEventRetries = math.MaxInt
 )
 
 type streamWriter interface {
@@ -124,9 +126,10 @@ func (ew *EventWriter) Add(ev *cloudevents.Event) {
 
 	defaultBackoff := wait.Backoff{
 		Steps:    maxEventRetries,
-		Duration: 1 * time.Second,
-		Factor:   1.5,
-		Jitter:   0.1,
+		Duration: 5 * time.Second,
+		Factor:   4,
+		Jitter:   1,
+		Cap:      2 * time.Minute, // never wait longer than 2 minutes
 	}
 
 	if resID == "" {
@@ -182,7 +185,7 @@ func (ew *EventWriter) Get(resID string) *eventMessage {
 
 	// Then check unsent queue
 	if eq, exists := ew.unsentEvents[resID]; exists {
-		return eq.get()
+		return eq.peek()
 	}
 	return nil
 }
@@ -207,13 +210,18 @@ func (ew *EventWriter) Remove(ev *cloudevents.Event) {
 		}
 	}
 
+	// JGW-TODO:
+	// - A) Should Remove remove from both sentEvents/unsentEvents? (I mean, it seems likely it will be in one or the other, but we could check both?)
+	// and
+	// - B) In the unsent case, why does remove only remove from the front of the queue?
+
 	// If not in sent events, check unsent queue
 	eq, exists := ew.unsentEvents[resourceID]
 	if !exists {
 		return
 	}
 
-	front := eq.get()
+	front := eq.peek()
 	if front == nil {
 		return
 	}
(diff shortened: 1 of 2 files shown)


internal/event/event_writer.go
```diff
@@ -484,26 +497,63 @@ func (eq *eventQueue) add(ev *eventMessage) {
...
+
+	// key: Type() of event
+	// value: (not used)
+	haveWeSeenMsgWithType := make(map[string]bool, 0)
+
+	// Work backwards through the list:
+	// - Items at the end of the list are 'fresher', items are the beginning of the list are more stale
+	// - We thus remove items early in the list in favour of those that are later in the list
+	for idx := len(*items) - 1; idx >= 0; idx-- {
+		item := (*items)[idx]
```
chetan-rns, line 536: I like this new approach. I wonder if there is a way to avoid scanning all the queue items for each incoming event. The old approach avoided this by only de-duplicating events that were next to each other. The downside was that older events could still remain in the queue if events of a different type appeared in between. For example:

A,B,C: different incoming event types ordered from left(oldest) to right(latest)
Incoming events: [A1 A2 B1 B2 A3 A4 B3 B4 C1 C2 A5 A6]
Old logic: [A2 B2 A4 B4 C2 A6] (de-duplicate items of the same type next to each other)
New logic: [B4 C2 A6] (de-duplicate items of the same type within the array)

The new logic definitely helps reduce the queue memory by removing all duplicate events. But it trades additional CPU time by scanning the queue for each incoming item. If there are M incoming events and N items in the queue, in the worst case, we take O(N) to scan the queue and O(N) to delete (Go will shift the remaining items), so I think the overall complexity is O(M * N * N). 

This may still be completely fine in our case, since our main goal is to **reduce queue memory**, and the queue size remains relatively smaller. I'm just thinking out loud about a possible future improvement.

jgwest:
@coderabbitai review

internal/event/event_writer.go
chetan-rns: I came across this while working on https://github.com/argoproj-labs/argocd-agent/pull/994. We could optimize and avoid the linear scan by using a doubly linked list and tracking the node pointers in a map. So, for an incoming event, we use the map to locate the old event in the list and remove the node O(1). But it adds further complexity. I'm using the linear scan to keep it simple for now, and added a TODO to revisit if it becomes an issue.

jgwest: +1 to everything you mentioned. Additional thoughts...
- There is likely a better algorithm we could use that uses slightly less CPU, without giving up the memory gains. 
- But, as you said, it's likely fine as it is.
- My expectation is what we have is fine, as:
    - The list is likely to be very short, in most cases (if there are no redis/resource proxy responses, then the list should generally only be 0-4 items long)
    - CPUs/memory are extremely fast, and the data is likely small enough to be fully within cpu cache
    - CPU/memory are extremely fast, especially relative to any other parts of the pipeline (e.g. agent still needs to send this data over I/O network, which is [going to be  thousands of times slower then anything the CPU is doing](https://gist.github.com/jboner/2841832) )

internal/event/event_writer.go
```diff
@@ -407,6 +417,8 @@ func (ew *EventWriter) sendUnsentEvent(resID string) {
 		ew.scheduleRetry(eventMsg)
 		ew.sentEvents[resID] = eventMsg
 	}
+	// Create thread local copy of target under lock to avoid a data race with UpdateTarget.
```
jgwest, line 420: This isn't strictly related to this PR, but Claude correctly flagged this as a data race, so I'm fixing it here.

chetan-rns (approved):
Thanks! LGTM.

jannfis (approved):
LGTM

PR #941, opened 2026-05-06
fix: principal send deadlock fix

drewbailey:

We observed argocd-agent principal with repeated error logs with "transport is closing" Send errors against a single agent. 

```
time="2026-05-05T20:49:32Z" level=error msg="Error while sending: rpc error: code = Unavailable desc = transport is 
closing\n" clientAddr="100.101.10.228:52936" event_id=m-p-yun-headlamp-production_ad4d09d4-146b-4b38-8b72-00c3ba5c0d1b_93364803 event_target=application event_type=io.argoproj.argocd-agent.event.spec-update
method=retrySentEvent module=EventWriter resource_id=m-p-yun-headlamp-production_ad4d09d4-146b-4b38-8b72-00c3ba5c0d1b retry_count=5
```

Root cause: when an agent reconnected before the old TCP connection fully closed, two concurrent Subscribe calls raced on activeClients and EventWriter.UpdateTarget. The loser left EventWriter pointed at the dead send half-stream while the winner held the live recv — so every retry hit the broken transport with no path to recovery.

Commit 1: Subscribe preemption + Send-error teardown
- Hold activeClientsMu across UpdateTarget so (activeClient, target, observer) swap atomically; cancel the prior client's context first.
- Install a SendErrorObserver that calls cancelFn on fatal errors (transport-closing, context-canceled) so the Subscribe goroutines exit and force a clean reconnect.
- Add EventWriterSendErrors counter (by agent + reason) and EventWriterQueueDepth gauge; guard all disconnect cleanup on activeClients[name] == c so a preempted connection's late close can't corrupt metrics for its live successor.

Commit 2: Bind SendWaitingEvents to its target generation
- NewEventWriter and UpdateTarget return a targetGen. SendWaitingEvents carries that gen through every Send; snapshotTargetForGen returns ok=false once the gen moves, preventing two loops from writing to the same gRPC stream concurrently.
- notifySendError drops notifications when targetGen has advanced, so a stale loop's Send error can't cancel a healthy new Subscribe.
- sendUnsentEvent skips the queue pop on gen mismatch so events aren't consumed from a stream that won't deliver them.

Commit 3: Fix lock-order inversion in retrySentEvent
- The prior shape acquired sentMsg.mu while holding ew.mu.RLock, while a concurrent path acquired them in the opposite order via notifySendError → ew.mu.RLock. Restructured to snapshot (stream, targetGen) under ew.mu.RLock, release it, check gen before touching retry state, release sentMsg.mu before Send(), and call notifySendError only after all locks are dropped.

Fixes #?

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes & Improvements**
  * Lease-aware event writing, improved concurrent subscription preemption, guarded cleanup to avoid races, and send-error handling that cancels faulty subscribers.

* **Monitoring & Observability**
  * Added per-agent queue-depth gauge and send-error counters; periodic publishing of queue depth.

* **Tests**
  * Expanded coverage for preemption, send-error cancellation, depth tracking, transport-error simulation, and concurrent scenarios.

### DIFF
--- a/agent/connection.go
+++ b/agent/connection.go
@@ -177,12 +177,13 @@ func (a *Agent) handleStreamEvents() error {
 	streamCtx, streamCancel := context.WithCancel(a.context)
 	defer streamCancel()
 
+	var lease event.TargetLease
 	if a.eventWriter == nil {
-		a.eventWriter = event.NewEventWriter("", stream)
+		a.eventWriter, lease = event.NewEventWriter("", stream)
 	} else {
-		a.eventWriter.UpdateTarget(stream)
+		lease = a.eventWriter.UpdateTarget(stream)
 	}
-	go a.eventWriter.SendWaitingEvents(streamCtx)
+	go a.eventWriter.SendWaitingEvents(streamCtx, lease)
 
 	logCtx := log().WithFields(logrus.Fields{
 		logfields.Module:     "StreamEvent",
--- a/internal/event/event_writer.go
+++ b/internal/event/event_writer.go
@@ -2,6 +2,7 @@ package event
 
 import (
 	"context"
+	"errors"
 	"sync"
 	"time"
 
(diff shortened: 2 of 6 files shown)


drewbailey:
this further complicates event writer, for now proposing minimal fix in #943

PR #943, opened 2026-05-06
fix(principal): reconnect Subscribe on EventWriter send failures

drewbailey:
Supersedes #941 

Add principal-side detection for EventWriter stream Send failures by wrapping the Subscribe stream before passing it to EventWriter. When Send returns a reconnectable transport error, cancel the owning Subscribe context so the connection tears down and the agent can reconnect cleanly.

This keeps EventWriter itself unchanged and avoids adding ownership tracking, stream leases, target generations, queue-depth metrics, or retry semantic changes. EventWriter continues to manage retry state as before; the principal only observes send failures at the stream boundary where they occur.

Add a new principal metric:

  principal_event_writer_send_errors_total{agent_name,reason}

The reason label classifies failures as:

  - context-canceled
  - transport-closing
  - other

Only context-canceled and reconnectable transport-closing errors cancel the Subscribe context. Other send errors are counted for observability but do not force reconnect behavior.

Add a focused Subscribe test that forces the mocked stream Send method to return codes.Unavailable and verifies the active Subscribe context is canceled.

Fixes #?

**Checklist**

## Summary by CodeRabbit

* **Improvements**
  * Enhanced monitoring and observability for event streaming operations with new error-tracking metrics for writer send failures.
  * Improved error classification and handling for event stream transport issues, with better categorization of error types.
  * Added error observability layer to track and respond to send failures on the event streaming path.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -63,7 +63,8 @@ type PrincipalMetrics struct {
 	EventReceived prometheus.Counter
 	EventSent     prometheus.Counter
 
-	EventProcessingTime *prometheus.HistogramVec
+	EventProcessingTime   *prometheus.HistogramVec
+	EventWriterSendErrors *prometheus.CounterVec
 
 	PrincipalErrors *prometheus.CounterVec
 }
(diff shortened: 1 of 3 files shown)


jannfis (approved):
LGTM, thanks

PR #945, opened 2026-05-07
fix: principal should allow apps from its namespace in the app filter chain

chetan-rns:

Currently, the principal doesn't reconcile apps from its own namespace. Users need to include the namespace under the allowed namespace list explicitly. This PR includes the principal's namespace when destination-based mapping is enabled.

Fixes #?

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced namespace filtering to include a principal's own namespace when using destination-based mapping mode, while maintaining existing behavior for standard filtering configurations.


jparsai (approved):
I verified changes locally, app is reconciled and synced properly when deployed in principal's namespace and allowed namespace doesn't include it.

PR #946, opened 2026-05-07
docs: support for hybrid architecture

### LINKED ISSUE #708: Support a hybrid architecture in the same Argo CD instance

Most people that take a look at the argocd-agent project will already have an Argo CD instance, and many of them use Argo CD in production. Trying out or evaluating argocd-agent in this context comes with quite some complexity, as the user would need to stand up a dedicated Argo CD instance (i.e. without running an application controller) next to their existing instance(s).

The argocd-agent architecture should support a "drop-in" installation, where the user can install and configure the principal component into an existing Argo CD installation, without any (or only minimal) changes to this installation would be required.

This would make it much easier for people to evaluate the argocd-agent, and it would aid in migrating from the traditional "push" based approach to the "pull" based approach introduced by the argocd-agent. Users would be able to perform migration one cluster at a time, with a potential rollback path.

N/A

**Additional context**
N/A

chetan-rns:

Add documentation for hybrid architecture setup: running argocd agent alongside an existing argocd instance setup.

Assisted-by: Cursor

Fixes https://github.com/argoproj-labs/argocd-agent/issues/708

**Checklist**

* [✅ ] Documentation update is required by this PR (and has been updated) OR no documentation update is required.

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive Hybrid Architecture guide describing how to run principal + agent alongside an existing Argo CD installation, with architecture overview, isolation controls, prerequisites, step‑by‑step setup, verification, and an application migration workflow.
  * Documented UI behavior for mixed app types, alternative hybrid topologies, a validation checklist, and a note that one isolation option requires Argo CD v3.4+.
  * Updated site navigation to include the new Hybrid Architecture guide.

[Review Change Stack](https://app.coderabbit.ai/change-stack/argoproj-labs/argocd-agent/pull/946?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/user-guide/hybrid-architecture.md
+++ b/docs/user-guide/hybrid-architecture.md
@@ -1,0 +1,361 @@
+# Hybrid Architecture: Running Agent alongside an existing Argo CD instance
+
+This guide explains how to run argocd-agent (principal + agent) alongside a pre-existing Argo CD installation on the same hub cluster. This enables you to try out argocd-agent without replacing your existing setup, gradually migrate applications, or run both architectures long-term.
+
+!!! tip "Who is this for?"
+    This guide is for teams that already have a working Argo CD installation managing applications and want to adopt argocd-agent incrementally. If you are setting up argocd-agent from scratch, see the [Getting Started](../getting-started/index.md) guide instead.
+
+## Overview
+
+In a hybrid architecture, the traditional Argo CD components (server, app controller, repo-server, Redis) continue to operate on the hub cluster. The argocd-agent principal is installed alongside them. Both systems share the same Argo CD UI and API, but each manages a distinct set of applications:
+
+- **Traditional apps**: managed by the hub app controller, deployed to spokes via direct cluster access
+- **Agent apps**: managed by the principal, deployed to spokes via the agent pull model
+
+The two systems coexist through isolation mechanisms on both sides: label selectors on the principal and agent, the `skip-reconcile` annotation on agent cluster secrets, and the `ignore-unmanaged-apps` flag as an additional safety net.
+
+## Architecture
+
+```mermaid
+flowchart TB
+    subgraph hub [Hub Cluster]
+        ArgoServer["Argo CD Server"]
+        AppController["App Controller"]
+        RepoServer["Repo Server"]
+        Redis["Redis"]
+        Principal["Principal"]
+        RedisProxy["Redis Proxy"]
+
+        ArgoServer -->|all Redis traffic| RedisProxy
+        RedisProxy -->|"non-agent keys (cluster|*, mfst|*, etc.)"| Redis
+        RedisProxy -->|"agent keys (app|resources-tree|*, app|managed-resources|*)"| Principal
+        AppController --> Redis
+    end
+
+    subgraph spoke [Spoke Cluster]
+        Agent["Agent"]
+        SpokeAppCtrl["App Controller"]
+        SpokeRepo["Repo Server"]
+        SpokeRedis["Redis"]
+        SpokeAppCtrl --> SpokeRedis
+    end
+
+    Principal <-->|"gRPC (agent-initiated)"| Agent
+    AppController -.->|"traditional apps (direct access)"| spoke
+```
+
+**Hub cluster** runs all existing Argo CD components unchanged, plus the principal and its Redis proxy. The Argo CD server is reconfigured to talk to the Redis proxy instead of Redis directly. The proxy transparently routes traffic: traditional app data goes to the Redis on the hub cluster, agent app data is forwarded to the appropriate agent via the principal.
+
+**Spoke cluster** runs the agent, a local app controller, repo-server, and Redis. The agent initiates a gRPC connection to the principal. The local app controller reconciles applications that the principal pushes to the spoke.
+
+## How Isolation Works
+
+The following mechanisms prevent conflicts between the traditional setup and the agent setup.
+
+### Label Selector
+
+The principal and agent are configured with a label selector. They only watch and process Kubernetes resources (Applications, AppProjects, Repositories) that carry the matching label. The existing app controller is unaware of this label and continues to process all applications as before.
+
+```yaml
+# Principal configuration (argocd-agent-params ConfigMap)
+principal.label-selector: "migrate=true"
+
+# Agent configuration (argocd-agent-params ConfigMap on spoke)
+agent.label-selector: "migrate=true"
+```
+
+The label selector is combined with a built-in exclusion of resources labeled `argocd-agent.argoproj-labs.io/ignore-sync=true`, using Kubernetes label selector AND semantics.
+
+### skip-reconcile Annotation
+
+When you create an agent using `argocd-agentctl agent create`, the resulting cluster secret is stamped with the annotation `argocd.argoproj.io/skip-reconcile: "true"`. This tells the hub app controller to skip reconciliation for all applications targeting that cluster.
+
+This is the key mechanism that prevents the hub app controller from interfering with applications that have been migrated to the agent. The agent cluster secret can coexist with a traditional cluster secret for the same physical spoke cluster because they have different names and point to different endpoints (the agent secret points to the resource proxy).
+
+!!! warning "Requires Argo CD v3.4+"
+    The `skip-reconcile` annotation is only supported in Argo CD v3.4 and later. On older versions, the hub app controller will still attempt to reconcile applications targeting agent clusters, which will cause conflicts.
+
+### Agent Label Selector
+
+The agent also supports a label selector (`agent.label-selector`), which filters resources at the Kubernetes API level on the spoke cluster. When configured, the agent's informers only list and watch Applications, AppProjects, and Repository secrets that match the selector. Pre-existing apps on the spoke that lack the label are invisible to the agent.
+
+```yaml
+# Agent configuration (argocd-agent-params ConfigMap on spoke)
+agent.label-selector: "migrate=true"
+```
+
+When the principal pushes an application to the spoke, it preserves the application's existing labels. Since only labeled apps are picked up by the principal in the first place, the spoke-side label selector naturally ensures the agent only processes apps that came through the principal.
+
+### ignore-unmanaged-apps (Additional Safety Net)
+
+As an additional safeguard, the `ignore-unmanaged-apps` flag tells the agent to skip any application that lacks the `argocd.argoproj.io/source-uid` annotation, which is automatically set by the agent when it creates an application locally on the spoke. This is an in-memory filter that runs after the label selector.
+
+This flag is passed via the `--ignore-unmanaged-apps` CLI flag or the `ARGOCD_AGENT_IGNORE_UNMANAGED_APPS` environment variable on the agent Deployment:
+
+```yaml
+# Add to the agent Deployment's container env
+- name: ARGOCD_AGENT_IGNORE_UNMANAGED_APPS
+  value: "true"
+```
+
+This is useful as a defense-in-depth measure: if a pre-existing spoke app happens to carry the same label used for migration, the agent will still skip it because it lacks the `source-uid` annotation. Without this flag, the agent would treat such apps as managed, send status events to the principal for apps that don't exist there, and cause errors.
+
+## Prerequisites
+
+Before setting up the hybrid architecture, ensure:
+
+- **Argo CD v3.4+** is installed on the hub cluster (required for `skip-reconcile`)
+- The hub cluster has a working Argo CD installation with server, app controller, repo-server, and Redis
+- Spoke clusters are network-accessible from the hub (for the traditional setup) and can initiate outbound connections to the hub (for the agent setup)
+- You have `argocd-agentctl` installed for PKI management and agent creation
+
+## Setup Guide
+
+### Step 1: Install the Principal
+
+Deploy the principal on the hub cluster alongside the existing Argo CD installation. Configure the label selector to ensure the principal only watches resources you explicitly label.
+
+```yaml
+# argocd-agent-params ConfigMap
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: argocd-agent-params
+  namespace: argocd
+data:
+  principal.label-selector: "migrate=true"
+  principal.destination-based-mapping: "true"
+  ..............
+```
+
+!!! note "Destination-based mapping"
+    Enabling destination-based mapping is recommended for hybrid setups. It allows applications to reside in any namespace on the hub and routes to agents based on `spec.destination.name` rather than the application's namespace. This means your existing applications can stay in their current namespaces during migration.
+
+### Step 2: Point the Argo CD Server to the Redis Proxy
+
+The Redis proxy is a component of the principal that intercepts Redis traffic from the Argo CD server. It routes agent-related keys (resource trees, managed resources) to the correct agent and forwards everything else to the Redis on the hub cluster.
+
+Update the Argo CD configuration to use the Redis proxy:
+
+```bash
+kubectl patch configmap argocd-cmd-params-cm -n argocd --type merge \
+  -p '{"data":{"redis.server":"argocd-agent-redis-proxy:6379"}}'
+```
+
+After patching, restart the Argo CD server to pick up the change:
+
+```bash
+kubectl rollout restart deployment argocd-server -n argocd
+```
+
+!!! important
+    This step is critical. Without it, the Argo CD UI cannot display resource trees or managed resources for agent-managed applications. The Redis proxy is transparent for traditional apps: their data continues to flow through to the Redis on the hub cluster.
+
+### Step 3: Create the Agent
+
+Use `argocd-agentctl` to create the agent on the hub. This creates a cluster secret with the `skip-reconcile` annotation and TLS credentials for the agent.
+
+```bash
+argocd-agentctl agent create agent-spoke-1 \
+  --namespace argocd
+```
+
+This creates:
+
+- A Kubernetes secret named `cluster-agent-spoke-1` in the `argocd` namespace
+- The secret has the annotation `argocd.argoproj.io/skip-reconcile: "true"`
+- The cluster's `server` URL points to the principal's resource proxy
+- TLS client credentials for the agent to authenticate with the principal
+
+You can verify the secret:
+
+```bash
+kubectl get secret cluster-agent-spoke-1 -n argocd \
+  -o jsonpath='{.metadata.annotations.argocd\.argoproj\.io/skip-reconcile}'
+# Output: true
+```
+
+### Step 4: Install the Agent on the Spoke
+
+Deploy the agent, app controller, repo-server, and Redis on the spoke cluster. Configure the agent to connect to the principal and enable the same label selector and mapping mode.
+
+```yaml
+# argocd-agent-params ConfigMap on the spoke cluster
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: argocd-agent-params
+  namespace: argocd
+data:
+  agent.destination-based-mapping: "true"
+  agent.create-namespace: "true"
+  agent.allowed-namespaces: "*"
+  agent.label-selector: "migrate=true"
+  ......................
+```
+
+### Step 5: Verify Connectivity
+
+Check that the agent has connected to the principal:
+
+```bash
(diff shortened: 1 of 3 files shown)


jannfis (commented):
I think this has to be linked from the TOC

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+
+The following mechanisms prevent conflicts between the traditional setup and the agent setup.
+
+### Label Selector
+
+The principal and agent are configured with a label selector. They only watch and process Kubernetes resources (Applications, AppProjects, Repositories) that carry the matching label. The existing app controller is unaware of this label and continues to process all applications as before.
+
+```yaml
+# Principal configuration (argocd-agent-params ConfigMap)
+principal.label-selector: "migrate=true"
```
jgwest: My assumption is that migrate=true is just the name of the label selector that was picked here, and doesn't have any intrinsic meaning to the code itself.

A) Is the migration scenario the most common scenario that users will use, and thus it makes sense to use `migrate=true` as the label selector? Or, would it make sense to use a more generic selector value?

B) If we don't want to change the selector from `migrate=true`, we could thus say something like:

In this example we are using 'migrate=true', which demonstrates how Applications can be migrated from traditional Argo CD push-based model, to Argo CD agent assisted pull-based model. We will expand on this more later in the documentation.
Any valid label selector can be used, but it must be consistent between agent and principal.

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+
+# Agent configuration (argocd-agent-params ConfigMap on spoke)
+agent.label-selector: "migrate=true"
+```
+
+The label selector is combined with a built-in exclusion of resources labeled `argocd-agent.argoproj-labs.io/ignore-sync=true`, using Kubernetes label selector AND semantics.
+
+### skip-reconcile Annotation
+
+When you create an agent using `argocd-agentctl agent create`, the resulting cluster secret is stamped with the annotation `argocd.argoproj.io/skip-reconcile: "true"`. This tells the hub app controller to skip reconciliation for all applications targeting that cluster.
```
jgwest: JGW: `app controller` -> `Argo CD application controller` throughout the document. To me 'app controller' feel like the colloquial term we use as devs to describe it, whereas `(Argo CD) application controller` would be its official title.

Presuming you agree, this can be updated here and elsewhere in the doc.

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+    AppController -.->|"traditional apps (direct access)"| spoke
+```
+
+**Hub cluster** runs all existing Argo CD components unchanged, plus the principal and its Redis proxy. The Argo CD server is reconfigured to talk to the Redis proxy instead of Redis directly. The proxy transparently routes traffic: traditional app data goes to the Redis on the hub cluster, agent app data is forwarded to the appropriate agent via the principal.
+
+**Spoke cluster** runs the agent, a local app controller, repo-server, and Redis. The agent initiates a gRPC connection to the principal. The local app controller reconciles applications that the principal pushes to the spoke.
+
+## How Isolation Works
+
+The following mechanisms prevent conflicts between the traditional setup and the agent setup.
```
jgwest: Is it expected that, during configuration, users would need to decide which of these features to use? If so, we should provide a general guideline on when these individual features should be enabled.

My comment here is that we are describing the configuration settings, but not suggesting when/why users should use them. (Or perhaps they should always be used, in which case we should say that).

For example, for label selector:
- Label selector should be set on argo cd agent, and that label selector should be applied to Applications which you wish to be synced to workload clusters via argo cd agent.

For example, for skip-reconcile:
- (skip reconcile description is already good, but...)  skip-reconcile should be added to a argo cd agent cluster secret to ensure it is not reconciled by non-argo-cd-agent, push-based argo cd mechanisms.

etc

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+# argocd-agent-params ConfigMap
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: argocd-agent-params
+  namespace: argocd
+data:
+  principal.label-selector: "migrate=true"
+  principal.destination-based-mapping: "true"
+  ..............
```
jgwest: IMHO, replace `..............` with ` # (...)`, here and elsewhere in the doc.

I wasn't able to find a place elsewhere in the documentation where we are using the `......` convention.

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+```
+
+!!! note "Destination-based mapping"
+    Enabling destination-based mapping is recommended for hybrid setups. It allows applications to reside in any namespace on the hub and routes to agents based on `spec.destination.name` rather than the application's namespace. This means your existing applications can stay in their current namespaces during migration.
+
```
jgwest, line 140: Can add a sentence like 'See the destination-based mapping page for details.', with a link to the page.

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+!!! important
+    This step is critical. Without it, the Argo CD UI cannot display resource trees or managed resources for agent-managed applications. The Redis proxy is transparent for traditional apps: their data continues to flow through to the Redis on the hub cluster.
+
+### Step 3: Create the Agent
+
+Use `argocd-agentctl` to create the agent on the hub. This creates a cluster secret with the `skip-reconcile` annotation and TLS credentials for the agent.
+
+```bash
+argocd-agentctl agent create agent-spoke-1 \
```
jgwest, line 166: Correct me if I'm wrong, but additional parameter are required here? If yes, we can either A) add those parameters (e.g. resource proxy url) or B) include a `(...)` with a note indicating additional parameters are required.

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+  --namespace argocd
+```
+
+This creates:
+
+- A Kubernetes secret named `cluster-agent-spoke-1` in the `argocd` namespace
+- The secret has the annotation `argocd.argoproj.io/skip-reconcile: "true"`
```
jgwest: Can add to the end of the sentence: (...) which tells the hub app controller to skip reconciliation for all applications targeting that cluster.

(Yes this is stated elsewhere, but it's useful to keep hammering that point)

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+- The cluster's `server` URL points to the principal's resource proxy
```
jgwest: Can add to the end of the sentence: ensuring Argo CD has visibility (via agent) into spoke cluster resources

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+kind: ConfigMap
+metadata:
+  name: argocd-agent-params
+  namespace: argocd
+data:
+  agent.destination-based-mapping: "true"
+  agent.create-namespace: "true"
+  agent.allowed-namespaces: "*"
+  agent.label-selector: "migrate=true"
+  ......................
```
jgwest: Same comment re: `......................` as above

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+### Step 5: Verify Connectivity
+
+Check that the agent has connected to the principal:
+
+```bash
+# On the hub, check principal logs for agent connection
+kubectl logs deployment/argocd-agent-principal -n argocd | grep "agent-spoke-1"
+
+# Verify the agent is connected
+kubectl logs deployment/argocd-agent-agent -n argocd | grep "connected"
```
jgwest: Is grepping for just 'connected' too generic? Will this actually allow the user to know the agent is successfully connected?

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+```bash
+# Traditional apps should still be Synced/Healthy
+argocd app list
+```
+
+## Migrating Applications
+
+Once the hybrid setup is running, you can migrate applications from the traditional setup to the agent-based setup. Migration is done per-application.
+
+### Step 1: Prepare the AppProject
```
jgwest: Can append `on workload cluster`

It's obvious from the context, but lets make it even more obvious

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+
+The AppProject must be available on the spoke cluster before migrating applications that reference it. Ensure the project's `destinations` and `sourceNamespaces` include the agent name, then add the migration label that matches the principal's label selector.
```
jgwest: Throughout the document, we use the term 'spoke cluster'. In the  upstream documentation we appear to prefer the term 'workload cluster'.

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+Verify that existing applications are unaffected:
+
+```bash
+# Traditional apps should still be Synced/Healthy
+argocd app list
+```
+
+## Migrating Applications
+
+Once the hybrid setup is running, you can migrate applications from the traditional setup to the agent-based setup. Migration is done per-application.
```
jgwest: I had to think through what specifically migration meant in this context. Here's what I came up with. I think we can include something like this in this paragraph:
- Migration in this context means moving from workload deployed by the hub's Argo CD instance, to that same workload deployed via spoke's Argo CD instance.
- That is, Applications that were previously deployed to the spoke cluster (by hub Argo CD)  via a push-based model, will now be deployed to the spoke cluster via _agent-managed pull-based_ model.

docs/user-guide/hybrid-architecture.md
jgwest: I thought in the managed mode, appproject is synced automatically?

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+    - namespace: '*'
+      name: agent-spoke-1   # Must match agent name
+  sourceRepos:
+    - '*'
+```
+
+### Step 2: Prepare Repository Secrets
+
+If your application uses private repositories, ensure the repository secret is project-scoped and labeled. Only project-scoped repository secrets are distributed to agents.
+
```
jgwest, line 269: Can add a short, quick example of what a project-scoped repository secret looks like (or link to upstream if an example is available there)

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+  sourceNamespaces:
+    - "agent-spoke-1"  # Must match agent name
+  destinations:
+    - namespace: '*'
+      name: agent-spoke-1   # Must match agent name
+  sourceRepos:
+    - '*'
+```
+
+### Step 2: Prepare Repository Secrets
```
jgwest: Can append 'on control plane cluster'

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+```yaml
+# Agent configuration (argocd-agent-params ConfigMap on spoke)
+agent.label-selector: "migrate=true"
+```
+
+When the principal pushes an application to the spoke, it preserves the application's existing labels. Since only labeled apps are picked up by the principal in the first place, the spoke-side label selector naturally ensures the agent only processes apps that came through the principal.
+
+### ignore-unmanaged-apps (Additional Safety Net)
+
+As an additional safeguard, the `ignore-unmanaged-apps` flag tells the agent to skip any application that lacks the `argocd.argoproj.io/source-uid` annotation, which is automatically set by the agent when it creates an application locally on the spoke. This is an in-memory filter that runs after the label selector.
```
jgwest: `This is an in-memory filter that runs after the label selector.` IMHO this an implementation detail that isn't worth mentioning to user. (Unless there is some way that this informs user's high level mental model of behaviour, here)

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+
+If your application uses private repositories, ensure the repository secret is project-scoped and labeled. Only project-scoped repository secrets are distributed to agents.
+
+```bash
+kubectl label secret repo-my-private-repo -n argocd migrate=true
+```
+
+### Step 3: Migrate the Application
+
+Migration involves two changes to the Application: updating the destination to point to the agent cluster, and adding the migration label.
```
jgwest: 'updating the destination' -> 'updating the destination (`.spec.destination`'s `name` and `server` fields)'

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+
+**Update the destination** to point to the agent cluster name. Since the agent cluster secret has `skip-reconcile`, the hub app controller will stop reconciling this application:
```
jgwest: `to point to the agent cluster name` -> `to point to the agent cluster name, while removing server field`

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+
+```bash
+kubectl patch application my-app -n argocd --type merge \
+  -p '{"spec":{"destination":{"name":"agent-spoke-1","server":""}}}'
+```
+
+**Add the migration label** so the principal picks up the application:
```
jgwest: `Add the migration label` -> `Add the shared selector label (in our case migrate: true)`

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
+# Hybrid Architecture: Running Agent alongside an existing Argo CD instance
+
+This guide explains how to run argocd-agent (principal + agent) alongside a pre-existing Argo CD installation on the same hub cluster. This enables you to try out argocd-agent without replacing your existing setup, gradually migrate applications, or run both architectures long-term.
```
jgwest: We can mention that this feature is currently managed mode specific

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+# Verify the app shows Synced/Healthy
+```
+
+## How the UI Works in Hybrid Mode
+
+The Argo CD UI works seamlessly for both traditional and agent-managed applications. All applications appear in the same UI, and users interact with them the same way.
+
+Behind the scenes, the Redis proxy handles the routing:
+
+| Operation | Traditional Apps | Agent Apps |
```
jgwest: This section feels to me like it is moving too far into describing internal technical details which users will not find useful.

It is generally beneficial for documentation to allow users to build a high-level mental model that allows them to understand argo cd agent behaviour, but IMHO this section goes too far into internal technical details that are perhaps too low level to be useful to them.

WDYT?

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+  name: argocd-agent-params
+  namespace: argocd
+data:
+  principal.label-selector: "migrate=true"
+  principal.destination-based-mapping: "true"
+  ..............
+```
+
+!!! note "Destination-based mapping"
+    Enabling destination-based mapping is recommended for hybrid setups. It allows applications to reside in any namespace on the hub and routes to agents based on `spec.destination.name` rather than the application's namespace. This means your existing applications can stay in their current namespaces during migration.
```
jgwest: Is it recommended or required?  If we support both ways of enabling this feature (dbm=true, dbm=false), we should document both.

OR, for example, if both are technically possible, but we think users really shouldn't do dbm=false, we could just prevent that in the code itself.

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,361 @@
...
+This flag is passed via the `--ignore-unmanaged-apps` CLI flag or the `ARGOCD_AGENT_IGNORE_UNMANAGED_APPS` environment variable on the agent Deployment:
+
+```yaml
+# Add to the agent Deployment's container env
+- name: ARGOCD_AGENT_IGNORE_UNMANAGED_APPS
+  value: "true"
+```
+
+This is useful as a defense-in-depth measure: if a pre-existing spoke app happens to carry the same label used for migration, the agent will still skip it because it lacks the `source-uid` annotation. Without this flag, the agent would treat such apps as managed, send status events to the principal for apps that don't exist there, and cause errors.
+
```
jgwest, line 109: Can we provide a guideline for when users should enable/disable the 'ignore-unmanaged-apps' feature?

For example, as described, it seems like it would always be useful to enable this feature. What are the circumstances where is would be useful to disable it?

jgwest (changes requested):
Because this topic is new to all readers, we need maximum hand holding. That means, for example, repeating/reminding the user of important points, avoiding implicit assumptions, consistent terminology, examples, etc.

The PR in its current form does a great job of that; as I've gone through I've thought of additional places we can add that information. LMK what you think, and definitely fact check what I am suggesting to ensure it actually makes sense 😅

docs/user-guide/hybrid-architecture.md
chetan-rns: The name has no special meaning. I have changed it to a more generic `agent-agent=true`. Also added a note saying that it can be any valid key/value label that suits the user's conventions.

docs/user-guide/hybrid-architecture.md
chetan-rns: I have updated the paragraphs to include which of them are required and optional.

docs/user-guide/hybrid-architecture.md
chetan-rns: Yes. The command is incomplete. Added `#(...)` mentioning additional parameters required here

docs/user-guide/hybrid-architecture.md
chetan-rns: It is synced automatically as long as the agent name is included in the destination and sourceNamespace. I've rephrased the sentence to make it less confusing.

docs/user-guide/hybrid-architecture.md
chetan-rns: Makes sense. I've linked the live resource page if the user wants to know how the Argo CD UI relies on the resource and redis proxy.

docs/user-guide/hybrid-architecture.md
chetan-rns: Hybrid architecture without dbm is technically possible, but the migration would be disruptive since the app namespaces would change. I have rephrased the paragraph to mention why it is highly recommended and the issue with namespaced mapping.

docs/user-guide/hybrid-architecture.md
chetan-rns: I have used hub/spoke to keep it consistent within this document. I think we are using both interchangeably. I saw them on other pages as well. But I agree that we should stick to one throughout the docs.

https://argocd-agent.readthedocs.io/latest/concepts/components-terminology/#hub

docs/user-guide/hybrid-architecture.md
```diff
@@ -0,0 +1,375 @@
...
+- [ ] Resource tree loads in the UI for agent apps
+- [ ] Pod logs are accessible through the UI
+- [ ] The hub's application controller does NOT reconcile agent apps
+
+### Coexistence
+
+- [ ] Both traditional and agent apps can be synced simultaneously without interference
+- [ ] Modifying a traditional app does not affect agent apps, and vice versa
+- [ ] Adding the label (e.g. `argocd-agent=true`) moves an app to agent management
+- [ ] Removing the label (e.g. `argocd-agent=true`) returns an app to traditional management
```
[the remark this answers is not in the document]
chetan-rns: I've removed the generic items since it requires more than adding/removing a label to migrate/rollback.

jannfis:
As part of this PR, can you please remove existing content in `docs/user-guide/migration.md` and put a notice in it that migration is now possible using hybrid mode, and link to your new doc?

jannfis (approved):
LGTM

jgwest (approved):
LGTM, great work @chetan-rns, thanks for the detailed documentation and for putting up with my nitpicky reviews 😅 !

PR #956, opened 2026-05-20
fix(redis-proxy): fallback to principal redis for local apps in namespace-based mapping mode and hybrid architecture

lvkins:
This PR fixes a bug in the principal's redis proxy that drops TCP connections (EOF error in the UI) for local/traditional ArgoCD applications when the principal is running in namespace-based-mapping mode (`destination-based-mapping: "false"`). The change will enable a hybrid architecture support when using autonomous agents alongside traditional, local ArgoCD applications on the principal (hub) cluster.

Architecture & Context:
We run a hub-and-spoke architecture:
1. Spoke clusters (tenants) run argocd-agent in autonomous mode. They deploy applications locally and only push observability data/status to the hub
2. The hub cluster (principal) serves as a single pane of glass for observability, but also manages its own traditional, local ArgoCD applications (e.g. in the `argocd` namespace).

Problem:
Because destination-based mapping is explicitly not supported for `autonomous` agents (it throws a `[FATAL] ...` error on agent startup), the principal must run in the default namespace-based mapping mode (`principal.destination-based-mapping: "false"`).

However, in this mode, `redisproxy.go` assumes that any application resource key (e.g., `app|resources-tree|...`) routed through it must contain an underscore `_` separator. If a user clicks on a traditional, local application resource (i.e. Pod/Namespace) in the ArgoCD UI, the proxy encounters a key without an underscore (e.g. `app|resources-tree|my-local-app|1.8.3`), throws an `unexpected lack of '_' namespace/name separator` error and immediately drops the TCP connection (`EOF` in the ArgoCD UI). This completely breaks the UI for local applications on the hub.

Fixes #?

To reproduce and test:

1. Deploy principal with `principal.destination-based-mapping: "false"`
2. Deploy a local/traditional ArgoCD application (no `_` in name/namespace) directly on the hub cluster.
3. Connect an agent in `mode: autonomous`.
4. Open the local/traditional application in the ArgoCD UI. The tree should render.
5. Click on any resource node (e.g. a pod or namespace) to open its details pane (Live Manifest / Diff).
   - Before this PR: The dialog fails to load, displaying an EOF error. The `argocd-agent-principal` logs show `unexpected lack of '_' namespace/name separator`. Dialog only renders properly for agent-managed resources.
   - After this PR: The proxy forwards the request to the local Redis. The manifest/diff dialog loads perfectly for both local and agent applications.

Unit tests in `redisproxy_test.go` have been updated to reflect the new expected fallback behavior for keys without an underscore.

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Redis key handling: keys missing the usual namespace separator are no longer treated as errors. They're now recognized as local application keys, forwarded appropriately, and a debug-level message is emitted. This prevents spurious failures when encountering non-standard key formats and improves overall stability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/argoproj-labs/argocd-agent/pull/956?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


chetan-rns (approved):
Thanks. LGTM!

PR #957, opened 2026-05-20
fix: check if resource being patched is managed

cjcocokrisp:
While reading through the code for the agent resource operations when working on the CRD related `GET` bug. I noticed that the `PATCH` operation for non sub-resources did not contain a call to `getManagedResource` which returns an error denying the request with forbidden if the resource is not managed by Argo CD. This PR adds that check to the operation to deny an attacker from being able to update all resources on the agent cluster if the principal was compromised. 

Fixes #?

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Patch requests now verify a resource is managed by the application before applying changes, preventing patches to unmanaged resources.
* **Tests**
  * Tests updated to cover managed vs. unmanaged resource patch behavior, ensuring correct success and error outcomes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/argoproj-labs/argocd-agent/pull/957?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/agent/resource.go
+++ b/agent/resource.go
@@ -162,6 +162,12 @@ func (a *Agent) processIncomingPostResourceRequest(ctx context.Context, req *eve
 }
 
 func (a *Agent) processIncomingPatchResourceRequest(ctx context.Context, req *event.ResourceRequest, gvr schema.GroupVersionResource) (*unstructured.Unstructured, error) {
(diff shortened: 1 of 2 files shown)


chetan-rns (approved):
Thanks @cjcocokrisp

LGTM!

Issue #950, opened 2026-05-15
Improve Principal Application adoption in Managed Mode.


drewbailey:
**Is your feature request related to a problem? Please describe.**

Today, when an Application exists on an Agent, and then an operator creates the same Application on a Principal cluster, the Agent silently ignores updates from the Principal. 

This is scenario is currently hard to detect, and the agent behavior to get the application managed by the Principal is unclear.

On the principal side, the Application appears in an odd state 
`Unable to load data: error getting destination cluster`

The agent log
```
time="2026-05-14T22:03:23Z" level=debug msg="skipping resource without source UID annotation" agent="{\"clientID\":\"m-ax-ea1-us\",\"mode\":\"managed\"}" event =io.argoproj.argocd-agent.event.response-synced-resource kind=Application method=processIncomingEvents mode=managed module=GrpcEvent name=myapp-m-ax-ea1-us namespace=project-myapp resource_id=38e5cb4c-4854-4a60-9cf0-58d2fc1f6daf uid=581d6693-0196-49c2-81c4-38724567180e
```

We intervened by deleting the application (non-cascading) on the agent cluster. 

On the principal side, the application had already been requested to be refreshed, so there were no new events flowing to the agent. We ended up just restarting the agent pod, and it picked up the Application from the Principal.

**Describe the solution you'd like**

At a minimum, maybe a annotation added to the application on the Principal that it is currently being excluded/ignored, and a way for the Agent to sync/check the principal on Application delete, so that it gets picked up from the principal quicker.

I think it could also make sense to have a policy flag to allow a principal to adopt an existing Application in managed mode.

**Describe alternatives you've considered**

**Additional context**

cjcocokrisp:
For implementing the adoption of applications, I think going for a similar implementation to the upsert policy for a mismatch in source UUIDs. Recreating might be a little hard because we then need to worry about not deleting resources.  One thing with stamping the existing application is that it might be out of sync with the principal but in that case it will be updated anyway.

WDYT?

PR #964, opened 2026-05-29
fix: detect if unmanaged app exists on agent before principal creates one

cjcocokrisp:

#950 discusses that in managed mode, if an unmanaged application exists on an agent that has the same name and namespace as one a principal app would create on the agent then there will be errors and it is hard to track down. This PR adds in a check to return a proper error if this happens. This check happens when an application is set to be created and the identity comparison shows that it exists but is missing a uid. 

N/A

1. Create an app on a managed agent
2. Create an app on the principal that would target the same app you created on the managed agent
3. Check the error logs to see if the new error messages are there

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Stricter validation in managed mode when an existing agent-side application lacks a source identifier: non-create events now fail to prevent invalid transitions.
  * For create events, if an unmanaged app already exists on the agent, the system logs and returns a clearer error with guidance to clean up or switch to autonomous mode.

* **Tests**
  * Internal test declarations refactored for clarity; no behavioral changes.

### DIFF
--- a/internal/manager/application/application.go
+++ b/internal/manager/application/application.go
@@ -38,8 +38,10 @@ import (
 	ty "k8s.io/apimachinery/pkg/types"
 )
 
-type updateTransformer func(existing, incoming *v1alpha1.Application)
-type patchTransformer func(existing, incoming *v1alpha1.Application) (jsondiff.Patch, error)
+type (
+	updateTransformer func(existing, incoming *v1alpha1.Application)
+	patchTransformer  func(existing, incoming *v1alpha1.Application) (jsondiff.Patch, error)
+)
 
 // LastUpdatedAnnotation is a label put on applications which contains the time
 // when an update was last received for this Application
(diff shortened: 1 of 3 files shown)


internal/manager/application/application.go
```diff
@@ -455,6 +456,9 @@ func (m *ApplicationManager) CompareSourceUID(ctx context.Context, incoming *v1a
 	if err != nil {
 		return result != nil && result.Exists, false, err
 	}
+	if result.ExistingMissingSourceUID {
+		return result != nil && result.Exists, false, fmt.Errorf("source UID Annotation is not found for app: test")
```
jparsai: I guess `app: test` is hardcoded by mistake.

jparsai (commented):
What I understand, this PR is only improving error logging to give clear info to user, but it is not fixing the issue reported in https://github.com/argoproj-labs/argocd-agent/issues/950.
For error visibility, it LGTM (execpt one minor fix), but original issue is yet to be fixed

cjcocokrisp:
> What I understand, this PR is only improving error logging to give clear info to user, but it is not fixing the issue reported in #950. For error visibility, it LGTM (execpt one minor fix), but original issue is yet to be fixed

Correct, actually functionality for the adoption will be addressed in a separate PR once we discuss how we want to implement it.

jparsai (approved):
Thanks, for error handling, it LGTM

PR #965, opened 2026-05-31
refactor: use %w instead of %v for error wrapping in fmt.Errorf calls

immanuwell:
Just replace `%v` with `%w` in 51 `fmt.Errorf` calls (20 files) where the last argument is an error. 
Using `%v` breaks the error chain - `errors.Is()` and `errors.As()` can't unwrap through it. 
The codebase already uses `%w` in 258 places; these were the remaining inconsistencies

## Test plan
- [x] `go build ./...` passes
- [x] `go test -race -p 1 ./...` passes (excluding E2E tests)

## Summary by CodeRabbit

* **Refactor**
  * Enhanced error handling throughout the application to improve diagnostic capabilities and troubleshooting when failures occur.


chetan-rns (approved):
Thanks. LGTM!

jannfis:
Thanks @immanuwell!

PR #969, opened 2026-06-01
feat: track dropped EventWriter events in metrics

### LINKED ISSUE #902: Metrics for events that are discarded by the event writer

We are using an event writer with retry functionality to ensure outbound events are being delivered. The number of retries is capped, and if it is reached, the outbound event will be dropped. This happens very much silently, but we need insight into when and how often that happens.

The event writer should export a metric that is updated whenever events are discarded before having been delivered.

**Additional context**

jparsai:
This PR is to add metrics for event that are discarded by the event writer for both agent and principal.

Fixes: https://github.com/argoproj-labs/argocd-agent/issues/902

Assisted by: Cursor

## Summary by CodeRabbit

* **New Features**
  * Metrics added to track events discarded after exhausting retry attempts, labeled by event and resource type for better observability.

* **Behavior**
  * Event writers now invoke a discard handler when events are dropped, causing the new discarded-event metrics to be incremented.

* **Tests**
  * Tests updated to assert discarded-event reporting is triggered when retries are exhausted.

* **Documentation**
  * Metrics docs updated to document the new discarded-events metrics and labels.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -177,7 +177,7 @@ func NewPrincipalMetrics() *PrincipalMetrics {
 		}, []string{"agent_name", "reason"}),
 
 		EventWriterEventsDiscarded: promauto.NewCounterVec(prometheus.CounterOpts{
-			Name: "principal_event_writer_events_discarded_total",
+			Name: "argocd_principal_event_writer_events_discarded_total",
 			Help: "The total number of events discarded by the EventWriter after exhausting retries",
 		}, []string{"agent_name", "event_type", "resource_type"}),
 
@@ -211,7 +211,7 @@ func NewAgentMetrics() *AgentMetrics {
 		}, []string{"resource_type"}),
 
 		EventWriterEventsDiscarded: promauto.NewCounterVec(prometheus.CounterOpts{
-			Name: "agent_event_writer_events_discarded_total",
+			Name: "argocd_agent_event_writer_events_discarded_total",
 			Help: "The total number of events discarded by the EventWriter after exhausting retries",
 		}, []string{"event_type", "resource_type"}),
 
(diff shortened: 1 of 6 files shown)


internal/metrics/metrics.go
```diff
@@ -174,6 +176,11 @@ func NewPrincipalMetrics() *PrincipalMetrics {
 			Help: "The total number of EventWriter send errors observed by principal",
 		}, []string{"agent_name", "reason"}),
 
+		EventWriterEventsDiscarded: promauto.NewCounterVec(prometheus.CounterOpts{
+			Name: "principal_event_writer_events_discarded_total",
```
jannfis:
```suggestion
			Name: "argocd_principal_event_writer_events_discarded_total",
```

internal/metrics/metrics.go
```diff
@@ -203,6 +210,11 @@ func NewAgentMetrics() *AgentMetrics {
 			Buckets: prometheus.DefBuckets,
 		}, []string{"resource_type"}),
 
+		EventWriterEventsDiscarded: promauto.NewCounterVec(prometheus.CounterOpts{
+			Name: "agent_event_writer_events_discarded_total",
```
jannfis:
```suggestion
			Name: "argocd_agent_event_writer_events_discarded_total",
```

jannfis (commented):
Generally LGTM, thanks!

Just two suggestions for the preparation of renaming of metric names, we might want to introduce new variables in the new naming scheme already.

internal/metrics/metrics.go
jparsai: It is renamed, PTAL

jannfis (approved):
LGTM!

PR #980, opened 2026-06-06
test: resolve go vet warnings in test files

immanuwell:

Fixes two `go vet` warnings in test files:

1. `cmd/cmdutil/text_test.go:17` - unexported field `invisible` had json/yaml/text tags; all marshalers skip unexported fields so the tags do nothing. The json tag triggers a vet warning. Removed all tags from that field.

2. `internal/manager/application/application_test.go:70` - `t.Fatalf` called from a goroutine. Go docs say Fatal/FailNow must only be called from the test goroutine; from a spawned goroutine it exits that goroutine silently, masking real failures. Changed to `t.Errorf`.

Fixes #?

Reproduce with `go vet ./...` before/after:

```
cmd/cmdutil/text_test.go:17:3: struct field invisible has json tag but is not exported
internal/manager/application/application_test.go:70:4: call to (*testing.T).Fatalf from a non-test goroutine
```

**Checklist**

## Summary by CodeRabbit

* **Tests**
  * Adjusted marshaling tests to verify visible text rendering while excluding internal metadata and ensuring unexported fields remain unchanged after round-trip.
  * Changed test error reporting to surface informer startup failures without aborting the whole test, improving diagnostics and test stability.

### DIFF
--- a/cmd/cmdutil/text_test.go
+++ b/cmd/cmdutil/text_test.go
@@ -14,7 +14,7 @@ import (
 func Test_MarshalStruct(t *testing.T) {
 	type good struct {
 		Visible   string `json:"visible" yaml:"visible" text:"Visible"`
-		invisible string `json:"invisible" yaml:"invisible" text:"Invisible"`
+		invisible string
 	}
 	t.Run("Marshal to JSON", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


jannfis (approved):
LGTM, thanks

PR #985, opened 2026-06-09
fix: noisy warning logs for gitfiles redis key

### LINKED ISSUE #984: Reduce "Unexpected redis key seen" to info or debug?
**Describe the bug**

Log messages in the principal complaining about an unexpected redis key `gitfiles`. I'd filed a previous bug about this in #751. I'm wondering if this message should be reduced to an `info` or maybe even `debug` level message instead of `warning` since new redis keys don't have an impact on the functionality of the Agent AFAIK.

Otherwise I feel like this is a never handing battle and as the Argo CD project adds new Redis keys the Agent code will need to be constantly updated to squash this message.

```
time="2026-06-08T21:25:34Z" level=warning msg="Unexpected redis key seen: 'gitfiles|https://github.com/gnunn-gitops/cluster-config-v2|d9dec7aed84f20b6b25905fe24ab0844bfa38a8e|apps/**/**/overlays/hub-prod-local/app-config.yaml|1.8.3.gz'. Redirecting to principal by default." connUUID=b7c97e19-2794-4a29-aa84-43a94b72b6f0 function=redisFxn module=redisProxy
```

1. Deploy an ApplicationSet using the git file generator
2. Check the principal logs for the Unexpected Redis Key message.

**Expected behavior**

This message should not be a Warning but rather Info or Debug. If there are specific keys that are problematic for the Agent for some reason those keys should be whitelisted to output a Warning rather then doing it for everything.

jparsai:
This PR is to fix https://github.com/argoproj-labs/argocd-agent/issues/984, by changing warning message to debug and add `gitfiles` to known redis key prefixes.

Assisted by: Cursor

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Redis cache routing for git files to be handled consistently with other git repository cache types.

### DIFF
--- a/agent/inbound_redis.go
+++ b/agent/inbound_redis.go
@@ -314,8 +314,8 @@ func (a *Agent) handleRedisGetMessage(logCtx *logrus.Entry, rreq *event.RedisReq
 // "app|resources-tree|my-app|1.8.3.gz
 func stripNamespaceFromRedisKey(key string, logCtx *logrus.Entry) (string, error) {
 
-	// git-refs and gitdirs keys don't contain namespace information; return as-is
-	if strings.HasPrefix(key, "git-refs|") || strings.HasPrefix(key, "gitdirs|") {
+	// git-refs, gitdirs, and gitfiles keys don't contain namespace information; return as-is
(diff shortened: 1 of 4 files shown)


chetan-rns (approved):
LGTM. Thanks!

Issue #979, opened 2026-06-04
Further improvements to informer/event/action logging


jgwest:
**Describe the bug**
As part of https://github.com/argoproj-labs/argocd-agent/pull/914 and https://github.com/argoproj-labs/argocd-agent/issues/864, we improved the logging of Argo CD agent such that we should now have full visiblity into input/outputs/actions, via logging of sent/received events, logging actions the agent takes, and logging of informer calls.

Now that we've had a chance to examine the log output from that PR, there are some additional tweaks we can make:
- Log message: `Event (sent/)received: resource GET`
	- At present, we are logging this, but not logging the details of the request. 
	- It would be beneficial to log the details of the request at info level (e.g. the full event body)
		- This should be logged when both ARGOCD_AGENT_FULL_DETAIL/ARGOCD_PRINCIPAL_FULL_DETAIL=false and ARGOCD_AGENT_FULL_DETAIL/ARGOCD_PRINCIPAL_FULL_DETAIL=true
	- We should continue NOT to log the response body, for security reasons.
	- We should only log GET, not PUT/POST (as they may contain sensitive data, depending on how they are used)
	
- `Event (sent/)received: redis redis-request`
	- At present, we are logging this, but not logging the details of the request. 
	- It would be beneficial to log the details of the request at info level (e.g. the full event body)
		- This should be logged when both ARGOCD_AGENT_FULL_DETAIL/ARGOCD_PRINCIPAL_FULL_DETAIL=false and ARGOCD_AGENT_FULL_DETAIL/ARGOCD_PRINCIPAL_FULL_DETAIL=true
	- We should continue NOT to log the response body, for security reasons.

- ~Event (sent/)received: clusterCacheInfoUpdate cluster-cache-info-update~
	- At present, we are logging this, but not logging the details of the request.
		- We should continue to log this at info level (no change)
	- However, it would be beneficial to ALSO log the full event body:
		- BUT, only when ARGOCD_AGENT_FULL_DETAIL/ARGOCD_PRINCIPAL_FULL_DETAIL=true
		- And only when trace level is either debug/trace
			- If controller log level is info, never log the body

Tweaks to the code:
- As [mentioned in one of the PR comments](https://github.com/argoproj-labs/argocd-agent/pull/914#discussion_r3297328265), there are some hardcoded constants which we can fix in the original PR:
	- Go through 'internal/logging/logging.go' and identify missing text constants which are defined elsewhere
		- For example in these functions:
[...]

PR #986, opened 2026-06-09
refactor: move event targets to new package

jparsai:
This PR is to move event targets to a dedicated package as requested in https://github.com/argoproj-labs/argocd-agent/issues/979, in another PR for logging (https://github.com/argoproj-labs/argocd-agent/pull/914#discussion_r3297328265) it was discussed to use these existing constants,  but because of circular dependency we couldn't do it in that PR.

Assisted by: Cursor

## Summary by CodeRabbit

* **Refactor**
  * Centralized event target identifiers into a shared module and updated event handling, ACK detection, and logging to use these shared identifiers for consistent behavior.
* **Tests**
  * Updated unit and e2e tests and import ordering to align with the new identifiers; no functional test logic changed.


chetan-rns (approved):
LGTM!

PR #987, opened 2026-06-09
feat: adopt existing applications on agent in managed mode

### LINKED ISSUE #950: Improve Principal Application adoption in Managed Mode.

cjcocokrisp:

At the current moment in managed mode, if an application is created on the agent before the principal and then the principal app targets that app it will appear in an unknown state. This PR fixes that issue by having the principal stamp the application if it detects it exists. This PR also adds a new annotation that can be put on agent apps and a global flag to not adopt applications if the user does not want this to be the default behavior. 

Fixes #950 

1. Set up an agent in managed mode
2. Create an application on the managed agent
3. Create an application on the principal that would create that same application on the agent you just created
4. See if the principal adopted the application on the agent by checking source UID annotations or if principal app shows up as healthy when before it would show up as unknown

**Checklist**

## Summary by CodeRabbit

* **New Features**
  * Agents can adopt and stamp pre-existing applications that lack a source UID.
  * New CLI/env flag to globally disable application adoption.
  * Per-application annotation to opt out or opt back in to adoption.

* **Tests**
  * Added unit and end-to-end tests covering adoption, the global disable flag, and the per-application annotation.

* **Documentation**
  * Added docs describing adoption behavior, the flag, and the per-application opt-out.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -164,6 +164,8 @@ type Agent struct {
 	// recreateAction defines the agent's behavior after recreating an app from
 	// an unauthorized deletion. It is only applicable in managed mode.
 	recreateAction manager.RecreateAction
+	// disableAdoption disables existing applications being stamped with source-uid on creation in managed mode
+	disableAdoption bool
 }
 
 const defaultQueueName = "default"
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -165,7 +165,7 @@ type Agent struct {
 	// an unauthorized deletion. It is only applicable in managed mode.
 	recreateAction manager.RecreateAction
 	// disableAdoption disables existing applications being stamped with source-uid on creation in managed mode
-	disableAdoption bool
+	adoptionPolicy string
 }
 
 const defaultQueueName = "default"
@@ -209,6 +209,7 @@ func NewAgent(ctx context.Context, client *kube.KubernetesClient, namespace stri
 	a.namespace = namespace
 	a.mode = types.AgentModeAutonomous
 	a.mismatchPolicy = manager.MismatchPolicyRecreate
+	a.adoptionPolicy = manager.AdoptionPolicyAlways
 	a.recreateAction = manager.RecreateActionIgnore
 	a.redisProxyMsgHandler = &redisProxyMsgHandler{}
 	// Resource proxy is enabled by default.
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -165,7 +165,8 @@ type Agent struct {
 	// an unauthorized deletion. It is only applicable in managed mode.
 	recreateAction manager.RecreateAction
 
-	// disableAdoption disables existing applications being stamped with source-uid on creation in managed mode
+	// adoptionPolicy sets the adoption policy for when an application that's going to be create
+	// already exists in managed mode.
 	adoptionPolicy manager.AdoptionPolicy
 }
 
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -164,8 +164,9 @@ type Agent struct {
 	// recreateAction defines the agent's behavior after recreating an app from
 	// an unauthorized deletion. It is only applicable in managed mode.
 	recreateAction manager.RecreateAction
+
 	// disableAdoption disables existing applications being stamped with source-uid on creation in managed mode
-	adoptionPolicy string
+	adoptionPolicy manager.AdoptionPolicy
 }
 
 const defaultQueueName = "default"
--- a/agent/inbound.go
+++ b/agent/inbound.go
@@ -1378,9 +1378,9 @@ func (a *Agent) ensureNamespaceExists(namespace string) error {
 	return nil
 }
 
-// shouldAdopt checks to see if an application should not be adopted
-// The annotation on the existing application takes priority over the the global setting
-func (a *Agent) effectiveAdoptionPolicy(incoming metav1.Object) string {
+// effectiveAdoptionPolicy gets the adoption policy for the existing appilcation
+// The annotation takes precedence over the global default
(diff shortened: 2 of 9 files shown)


test/e2e/adoption_test.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+
+		annotations := managedApp.GetAnnotations()
+		if annotations == nil {
+			return false
+		}
+
+		_, exists := annotations[manager.SourceUIDAnnotation]
+		return !exists
+	}, 30*time.Second, 1*time.Second)
```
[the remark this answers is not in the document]
cjcocokrisp: Not sure about this feedback it seems like the suggested change pretty much does the same thing because eventually checks every 1 second. Removing the time.Sleep for 3 seconds also has the chance of it not having the time to sync.

agent/inbound.go
```diff
@@ -332,6 +329,22 @@ func (a *Agent) syncManagedApplication(logCtx *logrus.Entry, incomingApp *v1alph
 			}
 			return nil
 		}
+	case identityActionAdoptExistingApp:
+		logCtx.Info("Application already exists on agent, adopting existing app")
+		dontAdopt, err := a.shouldNotAdopt(incomingApp)
```
jannfis: I think the naming is a bit confusing. Instead of using `dontAdopt` and `shouldNotAdopt` as names, using `adopt` and `shouldAdopt` (and reversing the logic) is better to understand (think of testing for `!dontAdopt`, for example).

agent/inbound.go
```diff
@@ -1358,3 +1380,23 @@ func (a *Agent) ensureNamespaceExists(namespace string) error {
 	log().Infof("Created namespace %s", namespace)
 	return nil
 }
+
+// shouldNotAdopt checks to see if an application should not be adopted
+// The annotation on the existing application takes priority over the the global setting
+func (a *Agent) shouldNotAdopt(incoming metav1.Object) (bool, error) {
```
jannfis: I think this function should not return an error. The way this function is written is that it may return a non-nil error and meaningful data in the other return value. This is considered bad practice, and since the caller does not treat the error returned by this function as an error (but rather a warning), it might make more sense to:

* log the failure (and consequences) in this function, and 
* just return `bool`, instead of `(bool, error)`.

jannfis (commented):
Thanks for the PR @cjcocokrisp. I have a couple of comments, PTAL.

agent/inbound.go
cjcocokrisp: Makes sense, how do you feel about the annotation and flag name then. Do you think that should stay don't adopt?

jannfis: Yeah, good call. 

Thinking more about it, do you feel there could be more values than `true` and `false` in the future? I'm thinking more along the lines of `argocd.argoproj.io/adoption-policy` (would probably affect the global policy, too). Instead of `true` and `false`, we could then use `always` and `never` instead to leave room for future cases that are not true or false, but depend on something else. Just floating as an idea.

cjcocokrisp: Yeah I think that's the best move. Was trying to make this so we could support more ways to adopt apps in the future if need came up.

agent/agent.go
```diff
@@ -164,6 +164,8 @@ type Agent struct {
 	// recreateAction defines the agent's behavior after recreating an app from
 	// an unauthorized deletion. It is only applicable in managed mode.
 	recreateAction manager.RecreateAction
+	// disableAdoption disables existing applications being stamped with source-uid on creation in managed mode
+	adoptionPolicy string
```
jannfis: For one, the comment mismatches reality and second, I think this should be its own type instead of `string`, e.g. `AdoptionPolicy` in the `manager` package (following the `RecreateAction`)

cjcocokrisp: Oops, I didn't realize those were types. Fixed.

agent/agent.go
```diff
@@ -164,6 +164,9 @@ type Agent struct {
 	// recreateAction defines the agent's behavior after recreating an app from
 	// an unauthorized deletion. It is only applicable in managed mode.
 	recreateAction manager.RecreateAction
+
+	// disableAdoption disables existing applications being stamped with source-uid on creation in managed mode
```
jannfis: The docstring comment is still wrong :)

jannfis (approved):
LGTM, thanks!

PR #895, opened 2026-04-13
fix(helm): use appVersion as default image tag instead of latest

enbiyagoral:
## Summary

The agent Helm chart hardcodes `image.tag` to `latest` and `appVersion` is stuck at `0.4.1` (project is at `v0.8.1`). This makes deployments non-deterministic and `helm list` shows incorrect version info.

This PR aligns the agent chart's image tag handling with the [official argo-cd Helm chart pattern](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/_helpers/_common.tpl#L54-L56) (`argo-cd.defaultTag`):

- Add `defaultTag` helper that falls back to `Chart.AppVersion` when `image.tag` is empty
- Change `image.tag` default from `"latest"` to `""` so `appVersion` is used
- Change `pullPolicy` from `Always` to `IfNotPresent`
- Bump `appVersion` to `v0.8.1` and chart version to `0.2.0`
- Sync `values.schema.json` defaults

## Behavior

| Scenario | Before | After |
|---|---|---|
| Default install | `argocd-agent:latest` | `argocd-agent:v0.8.1` |
| `--set image.tag=custom` | `argocd-agent:custom` | `argocd-agent:custom` |
| `helm list` APP VERSION | `0.4.1` | `v0.8.1` |

## Test plan

- [x] `helm template` renders `v0.8.1` as default tag
- [x] `helm template --set image.tag=custom` override works
- [x] `imagePullPolicy` renders as `IfNotPresent`
- [x] Schema validation passes

## Summary by CodeRabbit

* **Chores**
  * Helm chart bumped to v0.2.0 and appVersion to v0.8.1
  * Default image pull policy changed to IfNotPresent
  * Image tag now defaults to the chart appVersion when no explicit tag is set
  * Chart templates updated to consistently resolve the image tag from values or appVersion

* **Documentation**
  * README and chart metadata badges updated to reflect new versions and defaults

PR #888, opened 2026-04-09
feat(helm): Add missing params to agent helm chart

C4tWithShell:
**What does this PR do / why we need it**:
1. Missing `ARGOCD_AGENT_INSECURE_PLAINTEXT` env var
2. `labelSelector` and `redisTLS` missing from values.schema.json required fields
3. Fix the Redis TLS enabled default comment

**Which issue(s) this PR fixes**:
N/A

**How to test changes / Special notes to the reviewer**:

**Checklist**

* [X] Documentation update is required by this PR (and has been updated) OR no documentation update is required.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable plaintext TLS mode for the agent (new required chart value).
  * Introduced a label selector property as a required chart value for finer deployment targeting.
  * Redis TLS options updated and exposed for clearer control; default for agent Redis TLS changed to disabled.

* **Documentation**
  * Chart values and README updated to document the new TLS, label selector, and Redis TLS options.

PR #989, opened 2026-06-09
fix(helm): restore ARGOCD_AGENT_INSECURE_PLAINTEXT env on agent deployment

aplr:
## Summary

The agent chart writes `agent.tls.insecure-plaintext` to its params ConfigMap, but the deployment template doesn't map that key to an env var on the agent container. Effect: setting `tlsInsecurePlaintext: "true"` in values is silently a no-op. The binary's `ARGOCD_AGENT_INSECURE_PLAINTEXT` env stays unset, `insecurePlaintext` defaults to `false`, and the `cmd/argocd-agent/agent.go` else-branch fires — which loads the CA + client cert from secrets that mesh-only deployments don't have (and shouldn't need), and fatals on startup with `could not read secret: secrets "argocd-agent-ca" not found`.

This restores the env mapping the agent chart used to have. The principal chart's equivalent (`principal.tls.insecure-plaintext` → `ARGOCD_PRINCIPAL_INSECURE_PLAINTEXT` at `templates/principal-deployment.yaml:198-202`) was never affected.

## History on this exact env block

- [83f3243](https://github.com/argoproj-labs/argocd-agent/commit/83f32439ee9b820ec24674abb5d4c65081b3361f) (`feat(helm): Add missing params to agent helm chart`, #888) — **added** `ARGOCD_AGENT_INSECURE_PLAINTEXT` on the agent deployment.
- [e589ae1](https://github.com/argoproj-labs/argocd-agent/commit/e589ae1c6be4889bf055d29ca306ec95308a779d) (`fix(helm): use appVersion as default image tag instead of latest`, #895) — **removed** it. Looks like an unintentional drop during the surrounding `args` / `image:` restructure (the diff also dropped the container `name:` / `image:` lines from that env block area and re-inserted them after the env list; this env entry got caught in the same hunk).

## Test plan

- [x] `helm template` shows the new env entry on the agent container, sourced from the params CM at `agent.tls.insecure-plaintext` with `optional: true`.
- [x] Symmetric with the principal chart's existing mapping.

## Summary by CodeRabbit

* **New Features**
  * Added an optional Helm configuration to allow insecure plaintext communication for the agent; can be enabled via chart values or an agent ConfigMap.

* **Chores**
  * Bumped chart package version to 0.2.3.


jparsai (approved):
LGTM, Thanks!!

PR #990, opened 2026-06-09
fix: refresh self-registered cluster TLS data

drewbailey:
Existing self-registered cluster secrets could keep stale embedded TLS material after the source client cert secret rotated, because valid bearer tokens skipped registration entirely. Refresh the copied TLS config while preserving bearer tokens so resource-proxy discovery keeps working after cert rotation.

Fixes #?

**Checklist**

## Summary by CodeRabbit

* **New Features**
  * Agent registration now refreshes cluster TLS from rotated client certs/keys/CA and only updates cluster secrets when TLS data actually changes, reducing unnecessary secret writes.
  * Both valid- and invalid-token registration flows now ensure TLS is refreshed after rotation.

* **Tests**
  * Added tests validating TLS refresh behavior, preservation of existing bearer tokens, and retry/update behavior during secret rotation.


principal/registration/registration.go
```diff
@@ -82,14 +82,24 @@ func (mgr *AgentRegistrationManager) RegisterAgent(ctx context.Context, agentNam
...
+			}
+			logCtx.Debug("Cluster secret already exists with valid token")
 			return nil
 		}
 
 		// Token is invalid, update it
 		if err := cluster.UpdateClusterBearerTokenFromSecret(ctx, mgr.kubeclient, mgr.namespace, agentName, existingSecret, mgr.issuer); err != nil {
 			return fmt.Errorf("failed to refresh cluster bearer token: %w", err)
 		}
+		if _, err := cluster.UpdateClusterTLSFromSecret(ctx, mgr.kubeclient, mgr.namespace, agentName, existingSecret, mgr.clientCertSecretName); err != nil {
```
jparsai: I think after updating object via `UpdateClusterBearerTokenFromSecret` and before updating again via `UpdateClusterTLSFromSecret` it should fetch the latest version of object and pass that.

internal/argocd/cluster/cluster.go
```diff
@@ -305,6 +306,44 @@ func UpdateClusterBearerTokenFromSecret(ctx context.Context, kubeclient kubernet
 	return nil
 }
 
+func UpdateClusterTLSFromSecret(ctx context.Context, kubeclient kubernetes.Interface,
```
jannfis, line 311: Please give some godoc-style commentary for this function here.

principal/registration/registration.go
```diff
@@ -82,14 +82,31 @@ func (mgr *AgentRegistrationManager) RegisterAgent(ctx context.Context, agentNam
...
+		updatedSecret, err := cluster.GetClusterSecret(ctx, mgr.kubeclient, mgr.namespace, agentName)
+		if err != nil {
+			return fmt.Errorf("could not get updated cluster secret: %w", err)
+		}
+		if updatedSecret == nil {
+			return fmt.Errorf("cluster secret disappeared after bearer token refresh")
+		}
+		if _, err := cluster.UpdateClusterTLSFromSecret(ctx, mgr.kubeclient, mgr.namespace, agentName, updatedSecret, mgr.clientCertSecretName); err != nil {
+			return fmt.Errorf("failed to refresh cluster TLS data: %w", err)
+		}
```
jannfis: Is there a reason we do not log the TLS refresh status here (in the case of invalid token, and refreshed TLS data inside the cluster secret?

Just asking, because in the case where the token is valid, we explicitly state that the TLS data has been refreshed.

drewbailey: we should log, updated

jannfis (approved):
LGTM, thanks!

PR #991, opened 2026-06-10
feat: enhance informer/event/action logging

jparsai:
This PR is to improve informer/event/action logging as requested by https://github.com/argoproj-labs/argocd-agent/issues/979

Assisted by: Cursor

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced noisy logging by suppressing sensitive event payload details.
  * Refined event sensitivity rules: repository events always sensitive; resource events only non-sensitive for `GET`; redis request events non-sensitive while redis responses are sensitive.
  * Logged Kubernetes object diffs now omit `managedFields` without mutating originals.
* **Tests**
  * Expanded unit tests to verify managed-fields removal, non-mutating behavior, and updated sensitivity expectations.
  * Improved e2e stability by retrying application creation and extending a connection-state wait timeout.

### DIFF
--- a/internal/logging/logging.go
+++ b/internal/logging/logging.go
@@ -32,6 +32,7 @@ import (
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/argoproj-labs/argocd-agent/internal/event/targets"
 	"github.com/argoproj-labs/argocd-agent/internal/logging/logfields"
@@ -407,7 +408,7 @@ func LogActionUpdate(logCtx *logrus.Entry, resourceType string, oldObj, newObj a
 	// Include the diff in the log if full detail is enabled
 	if fullDetailConfig.Actions && oldObj != nil && newObj != nil && !isSecret(newObj) {
 		// Only include the diff if it is non-empty
-		if diff := cmp.Diff(oldObj, newObj); diff != "" {
+		if diff := cmp.Diff(dropMetadataManagedFields(oldObj), dropMetadataManagedFields(newObj)); diff != "" {
 			logCtx = logCtx.WithField(logfields.Detail, diff)
 		}
 	}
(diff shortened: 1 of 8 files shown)


jparsai:
https://github.com/argoproj-labs/argocd-agent/pull/986 needs to be merged first as it would cause conflicts.

internal/logging/logging.go
```diff
@@ -618,13 +619,17 @@ func hasEventData(data []byte) bool {
...
+// Exceptions: resource GET requests and redis requests are safe to log.
 func isSensitiveEvent(ev *cloudevents.Event) bool {
 	switch targets.EventTarget(ev.DataSchema()) {
-	case targets.Repository, targets.Redis, targets.Resource:
+	case targets.Repository:
 		return true
+	case targets.Resource:
+		return ev.Type() != "GET"
+	case targets.Redis:
+		return strings.HasSuffix(ev.Type(), "redis-response")
```
jgwest: Should use the `redis-response` constant, rather than `"redis-response"`. 

Unfortunately in this case, not using the constant has the potential to cause us a problem in the future. You can imagine a scenario like this:
- We change the name of the `redis-response` value in the event constant.
- BUT, we forget to update this reference.

If we were to accidentally do that, we would thus start leaking private confidential data via redis into the logs, which would be bad. 

I say 'Unfortunately' here, because I'm assuming the constant was intentionally not used here so as to avoid more refactor 😛

jgwest (changes requested):

internal/logging/logging.go
jparsai: Yes @jgwest that is a valid scenario. We are using 2 string here "GET" and "redis-response" in this file, but I dont think just moving these 2 event constants to new package and keeping others is existing constants would be a good idea, and if we move all event constants then we have to change 249 references in over 22 files. Should I do it?

jgwest:
Pushed commit to address review comment in another way, rebased.

test/e2e/fixture/argoclient.go
```diff
@@ -476,36 +476,45 @@ func GetArgoCDServerEndpoint(k8sClient KubeClient) (string, error) {
...
+		resp, err := c.Do(req)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create application: %w", err)
+		}
 
-	if resp.StatusCode != http.StatusOK {
 		body, _ := io.ReadAll(resp.Body)
-		return nil, fmt.Errorf("failed to create application: status %d, body: %s", resp.StatusCode, string(body))
-	}
+		resp.Body.Close()
```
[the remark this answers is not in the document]
jparsai: I think this is not required.

jgwest (approved):
LGTM, thanks @jparsai!

PR #994, opened 2026-06-11
feat: dedupe incoming events on the principal

### LINKED ISSUE #958: Implement event de-duplication for incoming events on the principal

Currently, the event writer implements partial de-duplication to send only the most recent events for a resource. This prevents processing unnecessary events and also helps reduce the queue size. However, when running at scale, the principal may be overwhelmed by the number of incoming events from large number of agents. The rate of processing events might be slower than the rate of inbound events. This results in accumulation of events on the principal waiting to be processed leading to increased memory usage. We need to implement de-duplication on the principal for incoming events so that the principal only processes the most recent event. This results in lower memory usage and also reduces the number of events processed.

The principal should only keep track and process the most recent incoming event for a resource. 

**Additional context**

chetan-rns:

In this PR, we de-dupe the incoming events on the principal. When running the principal at scale, the event processing rate might be slower than the inbound rate. This leads to more events accumulated in the principal's workqueue, increasing the memory usage. In this PR, we replace the standard workqueue with a custom queue that deduplicates any incoming event. The queue discards any older events before appending the latest event to the end of the queue. 

Alternate approach:
We don't have to replace the existing workqueue. We can add the dedupe queue as a separate component between the gRPC stream and the standard RecvQ work queue. There could be a goroutine that periodically drains the dedupe queue and fills the RecvQ workqueue. The downside is that we cannot remove the old events once it enters the workqueue. So, there is a possibility that the principal might still process a few old events. 

Assisted-by: Cursor

Fixes https://github.com/argoproj-labs/argocd-agent/issues/958

**Checklist**

## Release Notes

* **New Features**
  * Implemented a bounded deduplicating in-memory FIFO queue that automatically replaces eligible queued events for spec and status updates (while preserving non-deduped event types).
  * Added Prometheus metrics for the deduplication queue, including depth, adds, deduped events, and dequeue duration, and ensured they’re registered when metrics are enabled.

* **Tests**
  * Expanded unit tests to validate FIFO order, replacement behavior, max-size dropping, shutdown behavior, and mixed dedup/non-dedup scenarios.

### DIFF
--- a/internal/queue/dedupe_queue.go
+++ b/internal/queue/dedupe_queue.go
@@ -15,6 +15,7 @@
 package queue
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
@@ -47,6 +48,10 @@ func NewDedupeQueueForTest(maxSize int) *dedupeQueue {
 }
 
 func newDedupeQueue(maxSize int, name string) *dedupeQueue {
+	if maxSize <= 0 {
+		panic(fmt.Sprintf("maxSize must be positive, got %d", maxSize))
+	}
+
 	return &dedupeQueue{
 		items:   make([]*queueItem, 0),
 		notify:  make(chan struct{}, 10),
@@ -121,29 +126,29 @@ func (dq *dedupeQueue) Get() (*event.Event, bool) {
 	dq.mu.Unlock()
 
 	// Block until an item is available or shutdown
-	for {
-		select {
-		case <-dq.notify:
-			dq.mu.Lock()
-			if len(dq.items) > 0 {
-				item := dq.pop()
+	for range dq.notify {
+		dq.mu.Lock()
+		if len(dq.items) > 0 {
+			item := dq.pop()
 
-				if dq.metrics != nil {
-					dq.metrics.Depth.WithLabelValues(dq.name).Set(float64(len(dq.items)))
-					dq.metrics.Duration.WithLabelValues(dq.name).Observe(time.Since(item.enqueuedAt).Seconds())
-				}
-
-				dq.mu.Unlock()
-				return item.event, false
+			if dq.metrics != nil {
+				dq.metrics.Depth.WithLabelValues(dq.name).Set(float64(len(dq.items)))
+				dq.metrics.Duration.WithLabelValues(dq.name).Observe(time.Since(item.enqueuedAt).Seconds())
 			}
 
-			if dq.isShutdown {
-				dq.mu.Unlock()
-				return nil, true
-			}
 			dq.mu.Unlock()
+			return item.event, false
 		}
+
+		if dq.isShutdown {
+			dq.mu.Unlock()
+			return nil, true
+		}
+		dq.mu.Unlock()
 	}
+
+	// notify channel was closed (shutdown)
+	return nil, true
 }
 
 func (dq *dedupeQueue) Done(_ *event.Event) {
--- a/internal/queue/dedupe_queue.go
+++ b/internal/queue/dedupe_queue.go
@@ -179,6 +179,7 @@ func (dq *dedupeQueue) removeDuplicates(incoming *event.Event) {
 		return
 	}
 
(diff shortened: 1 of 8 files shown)


chetan-rns:
@coderabbitai full review

internal/queue/dedupe_queue.go
```diff
@@ -0,0 +1,199 @@
...
+
+func newDedupeQueue(maxSize int, name string) *dedupeQueue {
+	return &dedupeQueue{
+		items:   make([]*queueItem, 0),
+		notify:  make(chan struct{}, 10),
+		name:    name,
+		maxSize: maxSize,
+		metrics: metrics.GetDedupeQueueMetrics(),
+	}
+}
```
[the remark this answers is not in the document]
chetan-rns: Updated

chetan-rns:
I was going through the workqueue implementation. As an alternate approach, we can also replace the internal queue and customize it to support deduplication. PR https://github.com/argoproj-labs/argocd-agent/pull/1003 implements this alternate approach. Will add more details.

internal/queue/dedupe_queue.go
```diff
@@ -0,0 +1,205 @@
...
+	if item == nil || dq.isShutdown {
+		return
+	}
+
+	if canDedupe(item) {
+		dq.removeDuplicates(item)
+	}
+
+	if len(dq.items) >= dq.maxSize {
+		dq.pop()
```
jgwest, line 89: We're silently dropping messages here (of course the original queue code was also previously doing this); we can definitely log it, but at what level? Debug is the most obvious, but I'm worried that dropping messages is a legitimate error that might be missed (and users would not be running in debug by default, in production). 

---

Slightly related, Claude had the suggestion that we add a prom metric for dropped messages here (similar to elsewhere), but this is probably something we could add on a separate PR.

internal/queue/dedupe_queue.go
```diff
@@ -0,0 +1,205 @@
...
+
+// removeDuplicates removes all queued events matching the same resourceID and
+// event type as item. Must be called with dq.mu held.
+func (dq *dedupeQueue) removeDuplicates(incoming *event.Event) {
+	resID, evType := dedupeKey(incoming)
+	if resID == "" {
+		return
+	}
+
+	// TODO: avoid this linear scan/shift for every incoming event.
```
jgwest, line 182: Rather then leaving a `TODO:` comment, could instead leave a non `TODO:` comment like, `// This function uses linear scan/shift, which may be inefficient with large numbers of message. Consider improving if this is found to be a performance bottleneck` (or similar)

internal/queue/dedupe_queue.go
```diff
@@ -0,0 +1,205 @@
...
+type dedupeQueue struct {
+	mu         sync.Mutex
+	items      []*queueItem
+	notify     chan struct{}
+	name       string
+	maxSize    int
+	isShutdown bool
+	metrics    *metrics.DedupeQueueMetrics
+}
+
```
jgwest, line 39: Can add:
```
var _ RecvQueue = (*dedupeQueue)(nil)
```
Just to make it clear that `dedupeQueue` implements `RecvQueue`, and to flag with a compiler error when it doesn't.

jgwest (changes requested):

marcosaranda-appspace:
Following up from https://github.com/argoproj-labs/argocd-agent/issues/1051 (we are on **v0.10.0** now; the old retransmission collapse is gone in our fleet).

Quick status ask:

1. Is this PR still the intended approach vs #1003?
2. GitHub currently reports `mergeable_state=dirty` for this PR. Is a rebase planned for a post-0.10.0 release?
3. Confirming understanding from @chetan-rns on #1051: even after this lands, the **per-agent namedlock ceiling remains**. We are not treating this PR as the fix for #1051.

We can help validate a rebased build against our hub if useful.

PR #995, opened 2026-06-12
fix(agent): return error on nil item in sender instead of calling Done(nil)

immanuwell:

The agent `sender()` had a nil-item guard that called `q.Done(nil)` before returning. Thats wrong: the item was never inserted into the processing set by `Get()`, so calling `Done(nil)` operates on something that was never tracked. The `TODO` comment in the code explicitly flagged this as uncertain behavior.

The principal's equivalent `sendFunc()` already handles the nil case correctly - it returns an error. This PR makes the agent consistent with that.

Also fixes a typo in a comment: `princiapl` -> `principal`.

Fixes #?

The nil path itself is unreachable today (all `sendQ.Add()` callers pass non-nil events, and `Get()` only returns nil with `shutdown=true` which is caught earlier). But the code in that branch was still wrong.

To confirm the old behavior was broken, you can trace through:
1. `q.Get()` returns `(nil, false)` - only possible if `nil` was explicitly enqueued
2. shutdown check passes
3. `q.Done(nil)` - looks up nil in `q.processing`, which never had nil inserted, so its a no-op with wrong semantics
4. `return nil` - silently swallows the unexpected state

The principal does `return fmt.Errorf("panic: nil item in queue")` in the same situation. This PR brings the agent in line with that.

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Message-queue sender now surfaces an error when an unexpected nil queue item is encountered, improving visibility into delivery failures and preventing silent no-ops.

* **Documentation**
  * Fixed a spelling mistake in event-related documentation to improve clarity of message direction between agent and principal.

### DIFF
--- a/agent/connection.go
+++ b/agent/connection.go
@@ -86,9 +86,7 @@ func (a *Agent) sender(stream eventstreamapi.EventStream_SubscribeClient) error 
 	}
 	logCtx.Trace("Grabbed an item")
 	if ev == nil {
-		// TODO: Is this really the right thing to do?
-		q.Done(ev)
-		return nil
+		return fmt.Errorf("nil item in send queue")
 	}
 	defer q.Done(ev)
 	logCtx = logCtx.WithFields(logrus.Fields{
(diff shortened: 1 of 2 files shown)


chetan-rns (approved):
LGTM. Thanks!

Issue #961, opened 2026-05-27
Managed mode resource proxy fails Argo CD application-controller OpenAPI discovery, leaving remote Applications in Unknown


dlmather:
**Describe the bug**
Not 100% sure if this is just some mistake in my setup, or if there's something larger I'm missing but hitting the following issues:
  In managed mode, an Argo CD Application targeting a workload cluster through the argocd-agent resource proxy fails during application-controller comparison because Argo CD cannot load the destination cluster OpenAPI schema.

  The Application destination cluster server is the agent resource proxy, roughly:

  https://argocd-agent-resource-proxy.<namespace>.svc:9090?agentName=<agent-name>

  During reconciliation, Argo CD tries to sync its live-state cluster cache through that endpoint. API discovery partially works, but OpenAPI schema loading
  fails:

  failed to load open api schema while syncing cluster cache:
  error getting openapi resources:
  the server rejected our request for an unknown reason

  The Application remains in Unknown, and auto-sync is skipped.

**Steps to reproduce the behaviour**
  1. Install argocd-agent in managed mode with a principal in a control-plane/admin cluster and an agent in a workload cluster.
  2. Register the workload cluster in the control-plane Argo CD using the resource proxy as the cluster server, for example:

     https://argocd-agent-resource-proxy.<namespace>.svc:9090?agentName=<agent-name>

  3. Create an Argo CD Application in the control-plane Argo CD that targets the workload cluster by destination name/server.
  4. Let the control-plane Argo CD application-controller reconcile the Application.
  5. Observe the Application status and application-controller logs.

**Expected behavior**
  A control-plane Argo CD instance should be able to reconcile and sync an Application targeting a managed-mode workload cluster through the argocd-agent
  resource proxy.

  Specifically, Argo CD’s application-controller should be able to complete cluster cache sync, including Kubernetes OpenAPI discovery, so the Application can
  resolve to Synced or OutOfSync instead of Unknown.

**Additional context**

Thanks!

jannfis:
Do I understand correctly, that you are trying to have an application controller on the control plane reconcile an Application on a spoke by using one of the pseudo cluster secrets from the agent, instead of having a direct connection between the application controller and the remote cluster?

If yes, I'm not sure if that's an intended use case for the resource proxy.

dlmather:
[...]

PR #1001, opened 2026-06-17
fix: support resource proxy API group discovery

immanuwell:
What does this PR do / why we need it:
`/apis/<group>` is a valid Kubernetes discovery endpoint, but the resource proxy only matched `/apis/<group>/<version>`.

So discovery was kinda half-working. `/apis/apps` got rejected by the proxy before it ever reached the agent. This patch adds a dedicated matcher for group-only discovery and returns the matching `APIGroup` metadata from the agent.

Which issue(s) this PR fixes:
Related to #961

How to test changes / Special notes to the reviewer:
Repro for the old behavior:
`/apis/apps => false`
`/apis/apps/v1 => true`

Run:
`go test ./principal -run 'TestResource(RequestRegexp|GroupRequestRegexp)' -count=1`
`go test ./agent -run Test_getAvailableAPIs -count=1`
`go test ./agent ./principal -count=1 -timeout=10m`

Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved API group metadata retrieval when only a group name is provided, including preferred-version details.
  * Added routing support for Kubernetes API group discovery requests like `/apis/<group>` (no version specified).
* **Bug Fixes**
  * Now returns an error when a requested API group name doesn’t exist while no version is specified.
* **Tests**
  * Expanded coverage for group-only metadata lookups, missing-group error behavior, and the new discovery-request URL matching.


chrodrigues:
We run into #961 on a fleet of about 45 clusters (hub and spokes on v0.8.1), so I gave this patch a closer look.

The matcher and the agent side look correct to me, and `/apis/<group>` really is rejected today before it ever reaches the agent. One thing worth fixing though: when the group does not exist, `getAvailableAPIs` returns a plain `fmt.Errorf`. That goes through `event.HTTPStatusFromError`, which only understands `apierrors.APIStatus`, so the proxy answers 500. A real API server answers 404 with a Status object for an unknown group, and some clients probe for optional groups and rely on the 404 to mean "not installed". Returning `apierrors.NewNotFound` (or any StatusError carrying 404) would keep that behavior.

Minor: the real endpoint also sets `kind: APIGroup` and `apiVersion: v1` on the response. The converted struct comes out without TypeMeta, so strict clients get an untyped object back.

For what it is worth, this fixes one of the rejected discovery paths but not the whole story from #961. In our measurements the hub still fails cluster cache sync because the agent answers discovery requests one at a time, roughly 1.8s each and around 40 per sync on our clusters, which is enough to time out and poison the controller cache until a restart. We are preparing a patch for that part and will open a separate PR.

Happy to test this change against our lab clusters if that helps move it forward.

PR #1007, opened 2026-06-22
feat: add error event type and use it for when application adoption policy is never

cjcocokrisp:

This PR adds in an error event type that a managed agent can send back to the principal. When the principal receives this event it will update the status field with the error that was sent over. This event can help present errors to users. Error events are ignored by autonomous agents.

Sending an error over to the principal when the adoption policy is set to never was also added in this PR. This stops the application from showing up as unknown and instead shows the app as degraded. An example of the error can be seen below.

Screenshot 2026-06-22 at 11 40 41 AM

None

1. Create a scenario where an application will be adopted by the principal
   - Make sure the application on the managed agent has an adoption policy of never
2. Check to see if the application on the principal is degraded with the error information

**Checklist**

## Summary by CodeRabbit

* **New Features**
  * Applications can now surface adoption or agent failures through degraded health plus a dedicated error condition (with message details).
  * “Error” application events are now recognized in managed mode and reflected on application health/status.
* **Bug Fixes**
  * Adoption-disallowed scenarios (policy set to never) now enqueue an error condition for processing instead of only logging.
* **Tests**
  * Added/extended unit tests for managed vs autonomous handling of error events.
  * Updated E2E coverage for the “policy is never” adoption case (including health/condition validation).

### DIFF
--- a/agent/inbound.go
+++ b/agent/inbound.go
@@ -341,15 +341,16 @@ func (a *Agent) syncManagedApplication(logCtx *logrus.Entry, incomingApp *v1alph
 		case manager.AdoptionPolicyNever:
 			logCtx.WithField(logfields.Application, incomingApp.GetName()).Info("Existing application is set to not be adopted")
 			targetNamespace := a.getTargetNamespaceForApp(incomingApp)
-			a.addAppErrorToQueue(
-				incomingApp,
-				fmt.Sprintf(
-					"Application %s already exists on cluster managed by agent %s and its adoption policy is set to \"never.\" Please delete the existing Application on the agent cluster in namespace %s or change the adoption policy then recreate this app.",
-					incomingApp.Name,
-					a.namespace,
-					targetNamespace,
-				),
+			errData := &event.ErrorData{}
+			errData.ResourceNamespace = incomingApp.Namespace
+			errData.ResoureName = incomingApp.Name
+			errData.Message = fmt.Sprintf(
+				"Application %s already exists on cluster managed by agent %s and its adoption policy is set to \"never.\" Please delete the existing Application on the agent cluster in namespace %s or change the adoption policy then recreate this app.",
+				incomingApp.Name,
+				event.ErrorMessageAgentNamePlaceholder,
+				targetNamespace,
 			)
+			return a.addErrorEventToQueue(event.ApplicationError, errData)
 		}
 		return nil
 	default:
--- a/agent/inbound.go
+++ b/agent/inbound.go
@@ -340,17 +340,10 @@ func (a *Agent) syncManagedApplication(logCtx *logrus.Entry, incomingApp *v1alph
 			}
 		case manager.AdoptionPolicyNever:
 			logCtx.WithField(logfields.Application, incomingApp.GetName()).Info("Existing application is set to not be adopted")
-			targetNamespace := a.getTargetNamespaceForApp(incomingApp)
 			errData := &event.ErrorData{}
 			errData.ResourceNamespace = incomingApp.Namespace
 			errData.ResourceName = incomingApp.Name
-			errData.Message = fmt.Sprintf(
-				"Application %s already exists on cluster managed by agent %s and its adoption policy is set to \"never.\" Please delete the existing Application on the agent cluster in namespace %s or change the adoption policy then recreate this app.",
-				incomingApp.Name,
-				event.ErrorMessageAgentNamePlaceholder,
-				targetNamespace,
-			)
-			return a.addErrorEventToQueue(event.ApplicationError, errData)
+			return a.addErrorEventToQueue(targets.Application, errData)
 		}
 		return nil
 	default:
--- a/agent/inbound.go
+++ b/agent/inbound.go
@@ -340,7 +340,16 @@ func (a *Agent) syncManagedApplication(logCtx *logrus.Entry, incomingApp *v1alph
 			}
 		case manager.AdoptionPolicyNever:
 			logCtx.WithField(logfields.Application, incomingApp.GetName()).Info("Existing application is set to not be adopted")
-			a.addAppErrorToQueue(incomingApp, application.AppConditionAdoptionError, "Application that was to be created matches an existing application on the principal and the adoption policy is set to \"never\"")
+			targetNamespace := a.getTargetNamespaceForApp(incomingApp)
+			a.addAppErrorToQueue(
+				incomingApp,
+				fmt.Sprintf(
+					"Application %s already exists on cluster managed by agent %s and its adoption policy is set to \"never.\" Please delete the existing Application on the agent cluster in namespace %s or change the adoption policy then recreate this app.",
+					incomingApp.Name,
+					a.namespace,
+					targetNamespace,
+				),
+			)
 		}
 		return nil
 	default:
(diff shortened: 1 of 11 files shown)


jannfis (commented):
I think the text in the condition is more confusing than anything, tbh.

How about:

1. Renaming the condition itself to `AgentError`
2. Rephrasing the condition's message to something like:

 "Application `<app_name>` already exists on cluster managed by agent `<agent_name>` in namespace `<namespace>` and adoption policy is set to `never`. Either delete the existing Application on cluster `<agent_name>` or change the adoption policy".

This way, it becomes a) clear that the error originates from argocd-agent, and b) it clearly states the problem and involved parties.

Not too hung up on the phrasing, but I think the condition should be clear enough for people to be able to act on it.

cjcocokrisp:
Sounds good and makes sense, I just updated the error message so let me know what you think.

agent/outbound.go
```diff
@@ -171,6 +172,26 @@ func (a *Agent) addAppDeletionToQueue(app *v1alpha1.Application) {
 	logCtx.WithField(logfields.SendQueueLen, q.Len()).Debugf("Added app delete event to send queue")
 }
 
+// addAppErrorToQueue copies an application and adds error information and then puts it in the send queue
+func (a *Agent) addAppErrorToQueue(app *v1alpha1.Application, message string) {
+	logCtx := a.logGrpcEvent().WithField(logfields.Event, "AppError").WithField(logfields.Application, app.QualifiedName())
+	errorApp := application.SetErrorCondition(app, application.AppConditionAgentError, message)
```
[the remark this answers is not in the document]
cjcocokrisp: PR description was just never updated.

agent/inbound.go
```diff
@@ -340,6 +340,16 @@ func (a *Agent) syncManagedApplication(logCtx *logrus.Entry, incomingApp *v1alph
 			}
 		case manager.AdoptionPolicyNever:
 			logCtx.WithField(logfields.Application, incomingApp.GetName()).Info("Existing application is set to not be adopted")
+			targetNamespace := a.getTargetNamespaceForApp(incomingApp)
+			a.addAppErrorToQueue(
+				incomingApp,
+				fmt.Sprintf(
+					"Application %s already exists on cluster managed by agent %s and its adoption policy is set to \"never.\" Please delete the existing Application on the agent cluster in namespace %s or change the adoption policy then recreate this app.",
+					incomingApp.Name,
+					a.namespace,
```
jannfis: Not sure if the name of the agent actually can be expressed through `a.namespace`. Now that I'm thinking about it, I'm not even sure if the agent is aware of it's own name, because the name only has meaning on the principal (and the identity is extracted from either TLS cert or HTTP header).

Probably the error message needs to get enriched by the agent's name on the principal? 🤔 

Looking further at your code, I think we should not send the Application itself along with the error event, but some kind of specialized data structure that has all the required information so that the receiving event can handle it appropriately (i.e. in this case, set the error condition on the Application). This will also leave room for errors that cannot be expressed in an Application resource.

cjcocokrisp: Yeah in some cases the namespace might not be the actual name it was the closest thing I could find on the agent that could be its name. 

I think enriching it on the principal side could be the way to go (as the process event functions have agent name as an arg). Just send the error message over with some string to replace maybe like $AN.

Will change it to send an event data structure that just has an error message and resource type + name as the body. Anything else you can think of to add to that?

cjcocokrisp:
PR has been updated to now use a new error event target instead. Let me know if you think about this approach. Another option I thought of after finishing this, would be to keep the original approach and have it be an error event instead of target and have the event processing functions have a check to see if its error above everything else. That feels a little bit less clean to me because then it would be a lot of conditional switching inside of the already existing processing functions but I could be overthinking that.

cjcocokrisp:
Seems to be something flakey with the e2e test will investigate that tomorrow.

cjcocokrisp:
E2E failure is fixed. The agent was sending status updates on an update event which was overwriting the status and having it show up as healthy in the principal with some other errors. I checked without the error change and this was happening on the main branch too. Added a check into the queue add for status updates so that if its not from the principal it will not send the event because even if the app isn't in an adoption never scenario it will be a useless event to the principal and will fail with a resource not found on the update.

internal/event/event.go
```diff
@@ -61,6 +61,7 @@ const (
 	EventRequestResourceResync EventType = targets.TypePrefix + ".request-resource-resync"
 	ClusterCacheInfoUpdate     EventType = targets.TypePrefix + ".cluster-cache-info-update"
 	TerminalRequest            EventType = targets.TypePrefix + ".terminal-request"
+	ApplicationError           EventType = targets.TypePrefix + ".application-error"
```
jannfis: Shouldn't this be an `EventType` of just `error`, with a specific `EventTarget` indicating the source of the error?

agent/inbound.go
```diff
@@ -340,6 +340,17 @@ func (a *Agent) syncManagedApplication(logCtx *logrus.Entry, incomingApp *v1alph
 			}
 		case manager.AdoptionPolicyNever:
 			logCtx.WithField(logfields.Application, incomingApp.GetName()).Info("Existing application is set to not be adopted")
+			targetNamespace := a.getTargetNamespaceForApp(incomingApp)
+			errData := &event.ErrorData{}
+			errData.ResourceNamespace = incomingApp.Namespace
+			errData.ResourceName = incomingApp.Name
+			errData.Message = fmt.Sprintf(
```
jannfis: I apologize, but on a second thought I believe we shouldn't do any text substitution (i.e. placeholders) on the principal. Maybe we should just leave the agent name for now and remove any substitutions/placeholders.

We can prepend the error message with something like "Error occured on cluster managed by agent <agent>: " on the principal when setting the condition.

agent/outbound.go
```diff
@@ -114,6 +114,13 @@ func (a *Agent) addAppUpdateToQueue(old *v1alpha1.Application, new *v1alpha1.App
 	case types.AgentModeAutonomous:
 		eventType = event.SpecUpdate
 	case types.AgentModeManaged:
+		// If resource does not have principal uuid annotation we do not need to send it
+		// because it would fail to update with not found and in an adoption never case
+		// would set the app to healthy and override the error
+		if !isResourceFromPrincipal(new) {
+			logCtx.Errorf("Application %s is not from the principal, not sending status update", new.Name)
```
jannfis: Is this really an error? In which circumstances does this happen?

jannfis (commented):
Another couple of comments here, PTAL

agent/outbound.go
cjcocokrisp: This would happen when the local application that was set to never be adopted was updated (either manually or when it was synced). The informer would pick up the update and then send an update back to the principal and that update event would overwrite the error condition. This would happen even before the error condition the status update would just error out when trying to update the application during event processing because it does not exist. So I added this check in to not send the update event if the application is in managed mode and the application is not owned by a principal.

One thing I am concerned about after thinking about this some more is do you think that it effects HA or app sets? For app sets specifically I have seen the `MissingSourceUID` in the compare identity results is for when the app set wipes it so I'm a bit concerned that might be a false positive sometimes. 

Another solution to this error getting overwritten problem could be setting the app as unmanaged before sending the error event when the adoption policy is set to never.

internal/event/event.go
cjcocokrisp: I think I confused myself when fixing it. 😅

jannfis (approved):
LGTM, thanks @cjcocokrisp !

PR #1009, opened 2026-06-24
fix: temporarily disable data race detection

jgwest:
- This PR temporarily disables data race detection for E2E tests.
- Our current hypothesis is that the recently merged https://github.com/argoproj-labs/argocd-agent/pull/793 PR that enabled data race detection, has triggered a latent issue with one of the E2E tests, causing E2E test failures.
- We can thus disable data race detection for now until the issue is addressed.

**Checklist**

## Summary by CodeRabbit

* **Chores**
  * Changed the default test-run configuration so the data race detector is now off unless explicitly enabled. This should make end-to-end runs faster and less resource-intensive by default.


jparsai (approved):
LGTM @jgwest Thanks

PR #1021, opened 2026-06-30
fix(resource-proxy): Correctly resolve OLM relationships

### LINKED ISSUE #1017: Resource proxy fails to return resource for OLM operator
**Describe the bug**

The Argo CD Agent fails to return the resource in some circumstances but it's not 100% clear why. In this case where it fails the Argo CD Application is deployed in the same namespace as Argo CD. The ownership chain is fairly long since it is an OLM operator that deploys a subscription. 

The resource tree returns everything correctly. However clicking on the details of the ReplicaSet in the UI returns an error `Resource not found in cluster: cert-manager-operator-controller-manager-5c47dfd5dc`

Looking at the Argo CD Agent logs I see the following error message: 

```
time="2026-06-29T14:56:57Z" level=error msg="could not request resource: replicasets.apps \"cert-manager-operator-controller-manager-5c47dfd5dc\" is forbidden: resource not managed by app: resource not managed by app" http_method=GET method=processIncomingResourceRequest module=ResourceProxy uuid=87b5486e-0738-4557-be04-0adbb905a5ef
```

Image

I have another Argo CD Application which is a conventional Deployment, deployed in a different namespace using Apps-In-Any-Namespace, and there this works fine without issue.

I am mentioning the namespace because Jann thought there might a bug with the resource proxy for Applications that reside in the namespace as Argo CD.

1. Deploy an OLM operator using the Agent that in turn creates replicaset and pod, in my case I'm using cert-manager
2. Confirm the OLM operator replicaset and pod are rendered in the tree 
3. Click on the replicaset and see the  error

**Expected behavior**

The Agent should return the resource as expected.

jannfis:

OLM does not use ownerReferences to associate ClusterServiceVersion with an OperatorGroup, but instead uses an annotation. This PR mimics Argo CD's resource tree behaviour and resolves the relationship using that annotation, so that the resource proxy is able to correctly fetch all involved resources.

Fixes #1017 

**Checklist**

## Summary by CodeRabbit

* **New Features**
  * Improved ownership detection for Kubernetes resources managed through Operator Lifecycle Manager (OLM).
  * Resources can now be traced through inferred OLM ownership relationships, even when explicit ownership data is missing.

* **Bug Fixes**
  * `ClusterServiceVersion` resources with an operator group annotation are now recognized as managed.
  * Ownership checks now return a clear error when the referenced operator group cannot be found.
  * Resources without the required operator group annotation are correctly treated as unmanaged.

### DIFF
--- a/agent/resource.go
+++ b/agent/resource.go
@@ -470,6 +470,23 @@ func (a *Agent) getAvailableAPIs(ctx context.Context, group, version string) (*u
 	return &unstructured.Unstructured{Object: obj}, nil
 }
 
+// inferOwnerReferences returns synthetic owner references for resources where
+// Kubernetes ownerReferences are missing but the parent relationship is known
+// from other metadata. Mirrors gitops-engine's resolveResourceReferences logic.
+func inferOwnerReferences(res *unstructured.Unstructured) []v1.OwnerReference {
+	gvk := res.GroupVersionKind()
+	if gvk.Group == "operators.coreos.com" && gvk.Kind == "ClusterServiceVersion" {
+		if ogName := res.GetAnnotations()["olm.operatorGroup"]; ogName != "" {
+			return []v1.OwnerReference{{
+				APIVersion: "operators.coreos.com/v1",
+				Kind:       "OperatorGroup",
+				Name:       ogName,
+			}}
+		}
+	}
+	return nil
+}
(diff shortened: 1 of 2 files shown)


chetan-rns (approved):
LGTM!

PR #1024, opened 2026-07-03
fix(helm): prevent resource name collisions on long release names

chickenbeef:

Installing the `argocd-agent-agent` chart with a sufficiently long release name fails with the error:

```shell
Error: 1 error occurred:
* services "<name>-agent-helm" already exists
```

`agentBaseName` (`<release>-agent-helm`) can itself reach the 63-char DNS label limit. `resourceName` appends a suffix (`-metrics`, `-healthz`, `-role`, etc.) and *then* truncates to 63 chars — by that point the suffix is the part sliced off, so every suffixed resource silently collapses to the same name as the base. In practice this means the metrics and healthz Services (and potentially Role/RoleBinding, ClusterRole/ClusterRoleBinding, etc. for even longer release names) render identically, and Kubernetes rejects the second one on apply.

**Fix**

Introduce `argocd-agent-agent.safeName`, which truncates to 54 chars and appends an 8-char hash of the *full* untruncated candidate name whenever it would exceed 63 chars. Because the hash is derived from the complete base+suffix string, two inputs that only differ in their suffix always produce different, deterministic names.

Short release names are unaffected and will have the full names that fit within the 63-char limit.

Another option I thought of is to truncate long names based on the longest suffix (e.g. `clusterrolebinding` = 18 chars + hyphen = 19 so max safe base is 63 - 19 = 44 for full suffix preservation) but that could be a bit hairy as other object Kinds are added and we need to reduce it even further/yet again. The advantage of this would be more human readable suffixes though so maybe it's worth the maintenance overhead? Open to other ideas.

## Testing

- `helm lint .` passes.
- `helm template <short-release>` output unchanged (no hash suffix added when not needed):

  ```shell
  helm template "test-release" install/helm-repo/argocd-agent-agent -n test 2>&1 \
  | grep "^kind:" -A3 | grep -B1 "name:" | grep -v "kind:" | grep "name:" | sort -u
  name: test-release-agent-helm
  name: test-release-agent-helm-clusterrole
  name: test-release-agent-helm-clusterrolebinding
  name: test-release-agent-helm-healthz
  name: test-release-agent-helm-metrics
  name: test-release-agent-helm-params
  name: test-release-agent-helm-role
  name: test-release-agent-helm-rolebinding
  name: test-release-agent-helm-sa
  ```

- `helm template <62-char-release-name>` previously produced duplicate names for the metrics/healthz Services (and would have for other suffixed resources with a longer name); now every resource has a unique name:

  ```shell
  helm template "argocd-agent-afs1-npintexample-dev-apex-sharedple-np" install/helm-repo/argocd-agent-agent -n test 2>&1 \
  | grep "^kind:" -A3 | grep -B1 "name:" | grep -v "kind:" | grep "name:" | sort -u
  name: argocd-agent-afs1-npintexample-dev-apex-sharedple-np-a-399f7b2d
  name: argocd-agent-afs1-npintexample-dev-apex-sharedple-np-a-4a8fac82
  name: argocd-agent-afs1-npintexample-dev-apex-sharedple-np-a-7d8f636d
  name: argocd-agent-afs1-npintexample-dev-apex-sharedple-np-a-8b34c90f
  name: argocd-agent-afs1-npintexample-dev-apex-sharedple-np-a-9d3ab2b1
  name: argocd-agent-afs1-npintexample-dev-apex-sharedple-np-a-cc1309a6
  name: argocd-agent-afs1-npintexample-dev-apex-sharedple-np-a-e7bf8d81
  name: argocd-agent-afs1-npintexample-dev-apex-sharedple-np-a-eb0819a4
  name: argocd-agent-afs1-npintexample-dev-apex-sharedple-np-agent-helm
  ```

Fixes #?

Install release with a very long name:

  ```shell
  helm template "argocd-agent-afs1-npintexample-dev-apex-sharedple-np" \
  install/helm-repo/argocd-agent-agent \
  -n argocd \
  ```

**Checklist**

- [x] Documentation update is required by this PR (and has been updated) OR no documentation update is required.

## Summary

- **New Features**
  - Added deterministic, Kubernetes-safe resource name generation helpers (`safeName`) for long inputs across relevant charts.
- **Bug Fixes**
  - Updated generated resource names (including test resources) to use the new length-safe, hash-suffixed naming approach to avoid truncation-related collisions.
- **Chores**
  - Bumped Helm chart versions for the affected charts.
- **Documentation**
  - Updated chart version badges in the README files.

### DIFF
--- a/install/helm-repo/argocd-agent-agent/Chart.yaml
+++ b/install/helm-repo/argocd-agent-agent/Chart.yaml
@@ -5,7 +5,7 @@ description: Argo CD Agent for connecting managed clusters to a Principal
 type: application
 
 # chart version
-version: 0.2.4
+version: 0.2.5
 
 # application version, ArgoCD Agent version
 appVersion: v0.8.1
--- a/install/helm-repo/argocd-agent-agent/README.md
+++ b/install/helm-repo/argocd-agent-agent/README.md
@@ -1,6 +1,6 @@
 # argocd-agent-agent
 
-![Version: 0.2.4](https://img.shields.io/badge/Version-0.2.4-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.8.1](https://img.shields.io/badge/AppVersion-v0.8.1-informational?style=flat-square)
+![Version: 0.2.5](https://img.shields.io/badge/Version-0.2.5-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.8.1](https://img.shields.io/badge/AppVersion-v0.8.1-informational?style=flat-square)
 
 Argo CD Agent for connecting managed clusters to a Principal
 
--- a/install/helm-repo/argocd-agent-agent/templates/_helpers.tpl
+++ b/install/helm-repo/argocd-agent-agent/templates/_helpers.tpl
@@ -35,6 +35,28 @@ Base name for Helm-created resources, derived from the release.
 {{- end }}
 
 {{/*
+Produce a Kubernetes-compliant name (<= 63 chars, the DNS label limit).
+
+If the input already fits, it is returned unchanged. Otherwise it is
+truncated to 54 chars and suffixed with an 8-char hash derived from the
+full, untruncated input. This guarantees that two inputs which only differ
+in their tail (e.g. "<base>-metrics" vs "<base>-healthz") still produce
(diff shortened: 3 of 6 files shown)


jparsai (commented):
I guess principal chart would have the same issue and should be fixed along with agent chart.

chickenbeef:
> I guess principal chart would have the same issue and should be fixed along with agent chart.

Correct, I've applied the same change to the principal chart and also bumped the chart version to fix the failing status check

jannfis (approved):

jannfis:
LGTM, thanks

PR #1025, opened 2026-07-03
feat(helm/agent): add deploymentAnnotations value for Deployment metadata annotations

runderwoodcr14:
The agent chart exposes `podAnnotations` (pod template) but no way to set annotations on the Deployment object itself. Tools like [Stakater Reloader](https://github.com/stakater/Reloader) read their trigger annotations (e.g. `secret.reloader.stakater.com/reload` for TLS secret rotation restarts) from the **workload's** `metadata.annotations` only — so cert-rotation-driven agent restarts can't currently be wired via the chart.

This adds an optional `deploymentAnnotations` value rendered into the agent Deployment's `metadata.annotations` (empty by default — no behavior change), mirroring the existing `podAnnotations` pattern. Use case: mTLS client-cert rotation with cert-manager + Reloader bouncing the agent onto the renewed cert.

## Summary by CodeRabbit

* **New Features**
  * Added support for configurable Deployment metadata annotations for the agent via a new `deploymentAnnotations` setting.
  * Updated the Helm chart schema and defaults to include `deploymentAnnotations` (also reflected as required in the schema).
* **Documentation**
  * Documented `deploymentAnnotations` in the chart README and clarified how the annotations are applied.
* **Chores**
  * Bumped the chart version to `0.2.6`.


install/helm-repo/argocd-agent-agent/values.schema.json
```diff
@@ -799,7 +815,7 @@
         },
         "honorLabels": {
           "default": false,
-          "description": "When true, honorLabels preserves the metric’s labels when they collide with the target’s labels.",
+          "description": "When true, honorLabels preserves the metric\u2019s labels when they collide with the target\u2019s labels.",
```
jannfis: I don't think any of these should be unicode, but the ` ASCII character instead.

install/helm-repo/argocd-agent-agent/values.schema.json
```diff
@@ -198,7 +198,7 @@
     },
     "nameOverride": {
       "default": "",
-      "description": "Override the chart name used in `app.kubernetes.io/name`\nAlso changes `spec.selector.matchLabels`, which is immutable — do not set after initial\ninstall unless you are prepared to delete+reinstall.",
+      "description": "Override the chart name used in `app.kubernetes.io/name`\nAlso changes `spec.selector.matchLabels`, which is immutable \u2014 do not set after initial\ninstall unless you are prepared to delete+reinstall.",
```
jannfis: I disagree with the use of unicode here.

install/helm-repo/argocd-agent-agent/templates/agent-deployment.yaml
```diff
@@ -5,6 +5,10 @@ metadata:
   namespace: {{ include "argocd-agent-agent.namespace" . }}
   labels:
     {{- include "argocd-agent-agent.labels" . | nindent 4 }}
+  {{- with .Values.deploymentAnnotations }}
```
jannfis: I would prefer if both `labels` and `annotations` had the same style here

jannfis (commented):
Thanks for your PR @runderwoodcr14!

I have a couple of minor comments, PTAL.

Also, could you please sign-off your commits to satisfy the DCO as well as bump the Chart version. Thanks.

runderwoodcr14:
Thanks for the review @jannfis! Addressed: dropped the unrelated schema description/format changes, made `labels` and `annotations` symmetric in the Deployment metadata as requested, aligned the `deploymentAnnotations` schema entry with the podAnnotations pattern, bumped the chart to 0.2.6, and signed off the commit for DCO. Also rebased onto current main.

runderwoodcr14:
The failing `Validate Helm charts` check was the chart README lagging behind the last push — regenerated it with helm-docs to pick up the 0.2.6 bump and the new `deploymentAnnotations` entry. The remaining workflow runs are awaiting maintainer approval.

@jannfis ready for another look whenever you get a chance — thanks!

PR #1027, opened 2026-07-07
feat: Add TLS certificate blocklist to revoke agent access by fingerprint

jparsai:
This PR is to add support for TLS certificate blocklist that allows administrators to block agents by their client certificate fingerprint. The agents using blocked certificate are immediately disconnected and not allowed to reconnect until fingerprint is removed from blocklist or a new certificate is issued. This PR also adds new CLI commands to manage blocklist.

**Steps to testing**
- Build the CLI and use new CLI binary.
- Start principal and agent.
- Get fingerprint from principal logs
```
level=info msg="client authentication successful" agent_version=99.9.9-unreleased authmethod=mtls client=agent-autonomous fingerprint="D6:E6:2E:5F:10:CF" method=Authenticate module=grpc.AuthenticationServer
```
or get it from workload cluster using `argocd-agentctl agent inspect` CLI command

- Add the fingerprint to blocklist
```
argocd-agentctl blocklist --principal-context vcluster-control-plane --principal-namespace argocd-principal add <fingerprint>
```
- Agent having certificate of provided fingerprint will be disconnected and wont be allowed to reconnect.
- List the fingerprint 
```
argocd-agentctl blocklist --principal-context vcluster-control-plane --principal-namespace argocd-principal list
```
- Delete fingerprint to unblock the certificate
```
argocd-agentctl blocklist --principal-context vcluster-control-plane --principal-namespace argocd-principal remove <fingerprint>
```

Assisted by: Cursor

## Summary by CodeRabbit

* **New Features**
  * Added TLS certificate blocklist management via new `blocklist` CLI subcommands (`add`, `remove`, `list`).
  * Agent inspection now displays the client-certificate SHA-256 fingerprint.
  * The principal enforces the TLS blocklist for mTLS connections, immediately disconnecting matching agents and preventing reconnection until unblocked, including dynamic updates.
* **Documentation**
  * Documented blocklist behavior, how updates are applied, how to find fingerprints, and the ConfigMap data-key format.
* **Tests**
  * Added unit and end-to-end coverage for blocklist enforcement, reconnection behavior, and principal restart scenarios.

### DIFF
--- a/cmd/argocd-agent/principal.go
+++ b/cmd/argocd-agent/principal.go
@@ -30,6 +30,7 @@ import (
 	"github.com/argoproj-labs/argocd-agent/internal/auth/header"
 	"github.com/argoproj-labs/argocd-agent/internal/auth/mtls"
 	"github.com/argoproj-labs/argocd-agent/internal/auth/userpass"
+	"github.com/argoproj-labs/argocd-agent/internal/blocklist"
 	"github.com/argoproj-labs/argocd-agent/internal/config"
 	"github.com/argoproj-labs/argocd-agent/internal/env"
 	"github.com/argoproj-labs/argocd-agent/internal/grpcutil"
@@ -314,8 +315,20 @@ func NewPrincipalRunCommand() *cobra.Command {
 					}
 				}
 				mtlsauth := mtls.NewMTLSAuthentication(regex, source)
+
+				blockList := blocklist.New()
+				fingerprints, err := blocklist.LoadFromConfigMap(ctx, kubeConfig.Clientset, namespace)
+				if err != nil {
+					logrus.Warnf("Could not load TLS blocklist: %v", err)
+				} else {
+					blockList.Replace(fingerprints)
+					logrus.Infof("Loaded TLS blocklist with %d entries", blockList.Len())
+				}
+				mtlsauth.Blocklist = blockList
+				opts = append(opts, principal.WithBlocklist(blockList))
+
 				logrus.Infof("Using mTLS authentication (source: %s, pattern: %s)", source, regexStr)
-				err := authMethods.RegisterMethod("mtls", mtlsauth)
+				err = authMethods.RegisterMethod("mtls", mtlsauth)
 				if err != nil {
 					cmdutil.Fatal("Could not register mtls auth method: %v", err)
 				}
--- a/cmd/argocd-agent/principal.go
+++ b/cmd/argocd-agent/principal.go
@@ -305,7 +305,7 @@ func NewPrincipalRunCommand() *cobra.Command {
 			}
 
 			switch authMethod {
-			case "mtls":
+			case auth.MethodMTLS:
 				source, regexStr := parseMTLSConfig(authConfig)
 				var regex *regexp.Regexp
 				if regexStr != "" {
@@ -319,16 +319,15 @@ func NewPrincipalRunCommand() *cobra.Command {
 				blockList := blocklist.New()
 				fingerprints, err := blocklist.LoadFromConfigMap(ctx, kubeConfig.Clientset, namespace)
 				if err != nil {
-					logrus.Warnf("Could not load TLS blocklist: %v", err)
-				} else {
-					blockList.Replace(fingerprints)
-					logrus.Infof("Loaded TLS blocklist with %d entries", blockList.Len())
+					cmdutil.Fatal("Could not load TLS blocklist: %v", err)
 				}
+				blockList.Replace(fingerprints)
+				logrus.Infof("Loaded TLS blocklist with %d entries", blockList.Len())
 				mtlsauth.Blocklist = blockList
 				opts = append(opts, principal.WithBlocklist(blockList))
 
 				logrus.Infof("Using mTLS authentication (source: %s, pattern: %s)", source, regexStr)
-				err = authMethods.RegisterMethod("mtls", mtlsauth)
+				err = authMethods.RegisterMethod(auth.MethodMTLS, mtlsauth)
 				if err != nil {
 					cmdutil.Fatal("Could not register mtls auth method: %v", err)
 				}
@@ -337,17 +336,17 @@ func NewPrincipalRunCommand() *cobra.Command {
 				if !requireClientCerts {
 					opts = append(opts, principal.WithRequireClientCerts(true))
 				}
-			case "userpass":
+			case auth.MethodUserPass:
 				userauth := userpass.NewUserPassAuthentication(authConfig)
 				err = userauth.LoadAuthDataFromFile(authConfig)
 				if err != nil {
 					cmdutil.Fatal("Could not load user database: %v", err)
 				}
-				err = authMethods.RegisterMethod("userpass", userauth)
+				err = authMethods.RegisterMethod(auth.MethodUserPass, userauth)
 				if err != nil {
 					cmdutil.Fatal("Could not register userpass auth method: %v", err)
 				}
-			case "header":
+			case auth.MethodHeader:
 				// Generic header-based authentication extracts agent ID from any HTTP header
 				// Format: header:<header-name>:<extraction-regex>
 				headerName, extractionRegex, err := parseHeaderAuth(authConfig)
@@ -358,7 +357,7 @@ func NewPrincipalRunCommand() *cobra.Command {
 				if err := headerAuth.Init(); err != nil {
 					cmdutil.Fatal("Error initializing header auth: %v", err)
 				}
-				err = authMethods.RegisterMethod("header", headerAuth)
+				err = authMethods.RegisterMethod(auth.MethodHeader, headerAuth)
 				if err != nil {
 					cmdutil.Fatal("Could not register header auth method: %v", err)
 				}
@@ -786,12 +785,12 @@ func parseAuth(authStr string) (string, string, error) {
 		return "", "", fmt.Errorf("invalid auth string")
 	}
 	switch p[0] {
-	case "userpass":
-		return "userpass", p[1], nil
-	case "mtls":
-		return "mtls", p[1], nil
-	case "header":
-		return "header", p[1], nil
+	case auth.MethodUserPass:
+		return auth.MethodUserPass, p[1], nil
+	case auth.MethodMTLS:
+		return auth.MethodMTLS, p[1], nil
+	case auth.MethodHeader:
+		return auth.MethodHeader, p[1], nil
 	default:
 		return "", "", fmt.Errorf("unknown auth method: %s", p[0])
 	}
@@ -849,7 +848,7 @@ func parseMTLSConfig(config string) (mtls.IdentitySource, string) {
 //   - mtls auth requires TLS mode (needs client certificates)
 func validateAuthTLSPairing(authMethod string, insecurePlaintext bool) error {
 	switch authMethod {
-	case "header":
+	case auth.MethodHeader:
 		if !insecurePlaintext {
 			return fmt.Errorf("invalid configuration: header-based authentication requires --insecure-plaintext=true\n" +
 				"  Header authentication is designed for service mesh environments (e.g., Istio) where\n" +
@@ -858,7 +857,7 @@ func validateAuthTLSPairing(authMethod string, insecurePlaintext bool) error {
 				"  - Add --insecure-plaintext flag when using header auth behind a service mesh\n" +
 				"  - Use --auth=mtls:<regex> for direct TLS connections without a service mesh")
 		}
-	case "mtls":
+	case auth.MethodMTLS:
 		if insecurePlaintext {
 			return fmt.Errorf("invalid configuration: mtls authentication cannot be used with --insecure-plaintext\n" +
 				"  mTLS authentication requires TLS to be enabled to receive client certificates.\n" +
--- a/cmd/ctl/agent.go
+++ b/cmd/ctl/agent.go
@@ -341,6 +341,7 @@ func NewAgentInspectCommand() *cobra.Command {
 				Name           string    `yaml:"name" json:"name" text:"Server name"`
 				NotValidBefore time.Time `yaml:"notValidBefore" json:"notValidBefore" text:"Not valid before"`
 				NotValidAfter  time.Time `yaml:"notValidAfter" json:"notValidAfter" text:"Not valid after"`
+				Fingerprint    string    `yaml:"fingerprint" json:"fingerprint" text:"Fingerprint (SHA-256)"`
 			}
 			agentName := args[0]
 			argoCluster, err := loadClusterSecret(agentName)
@@ -354,11 +355,18 @@ func NewAgentInspectCommand() *cobra.Command {
 			if err != nil {
 				cmdutil.Fatal("Not a valid certificate: %v", err)
 			}
+
+			certObj, parseErr := x509.ParseCertificate(cert.Certificate[0])
+			if parseErr != nil {
+				cmdutil.Fatal("Could not parse certificate: %v", parseErr)
+			}
+
 			cluster := &clusterOut{
 				ServerAddr:     argoCluster.Server,
 				Name:           argoCluster.Name,
 				NotValidAfter:  cert.Leaf.NotAfter,
 				NotValidBefore: cert.Leaf.NotBefore,
+				Fingerprint:    tlsutil.CertificateFingerprint(certObj),
 			}
 			var out []byte
 			switch strings.ToLower(outputFormat) {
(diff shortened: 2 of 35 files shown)


cmd/argocd-agent/principal.go
```diff
@@ -313,8 +314,20 @@ func NewPrincipalRunCommand() *cobra.Command {
...
+				} else {
+					blockList.Replace(fingerprints)
+					logrus.Infof("Loaded TLS blocklist with %d entries", blockList.Len())
+				}
+				mtlsauth.Blocklist = blockList
+				opts = append(opts, principal.WithBlocklist(blockList))
+
 				logrus.Infof("Using mTLS authentication (source: %s, pattern: %s)", source, regexStr)
-				err := authMethods.RegisterMethod("mtls", mtlsauth)
+				err = authMethods.RegisterMethod("mtls", mtlsauth)
```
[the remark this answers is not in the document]
jparsai: I dont think this is valid, The `LoadFromConfigMap` function already handles it.

internal/blocklist/blocklist.go
```diff
@@ -0,0 +1,94 @@
...
+func ParseFingerprints(data string) ([]string, error) {
+	if data == "" {
+		return []string{}, nil
+	}
+	var fingerprints []string
+	if err := json.Unmarshal([]byte(data), &fingerprints); err != nil {
+		return nil, err
+	}
+	return fingerprints, nil
+}
```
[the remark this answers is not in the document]
jparsai: This seems not valid, agent and principal use same `CertificateFingerprint` and `ParseFingerprints` functions so output will be same.

internal/blocklist/configmap.go
```diff
@@ -0,0 +1,84 @@
...
+
+	if cm.Data == nil {
+		cm.Data = make(map[string]string)
+	}
+	cm.Data[config.ConfigMapKeyBlocklistChecksums] = string(data)
+	if _, err := client.CoreV1().ConfigMaps(namespace).Update(ctx, cm, metav1.UpdateOptions{}); err != nil {
+		return fmt.Errorf("failed to update blocklist ConfigMap: %w", err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
jparsai: This is not valid. `SaveToConfigMap` is only called from the CLI command and possibility of conflicts is low and admin will never run multiple `add` commands at the same time.

jannfis (commented):
Instead of having a single key in the ConfigMap encoding a JSON document, I think we should have any fingerprint being it's own key (without the colons added for human readability, for obvious reasons).

IMO, this would make operations much faster (you operate on map keys, instead of having to encode/decode JSON all of the time).

cmd/ctl/blocklist.go
```diff
@@ -0,0 +1,162 @@
...
+
+func NewBlocklistCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "blocklist",
+		Short: "Manage the TLS certificate blocklist",
+	}
+
+	cmd.AddCommand(NewBlocklistAddCommand())
+	cmd.AddCommand(NewBlocklistRemoveCommand())
+	cmd.AddCommand(NewBlocklistListCommand())
```
jannfis: To leave room for other types of potential blocklists in the future without having to have weird UX or a breaking change, IMO the name of the commands to add TLS certificate fingerprints should be `blocklist tls add`, `blocklist tls remove` etc. instead of just `blocklist add <fingerprint>`, etc

jparsai:
> Instead of having a single key in the ConfigMap encoding a JSON document, I think we should have any fingerprint being it's own key (without the colons added for human readability, for obvious reasons).
> 
> IMO, this would make operations much faster (you operate on map keys, instead of having to encode/decode JSON all of the time).

Thanks Jann, It is updated, PATL. 
Now I am saving data as map instead of array. 
Though it is reading ConfigMap data only once in informer when CM is created/updated. Informer saves blocked fingerprints into a in-memory map (in FingerPrint -> AgentName pair) and this map is used everytime in authentication.

cmd/argocd-agent/principal.go
```diff
@@ -314,8 +315,20 @@ func NewPrincipalRunCommand() *cobra.Command {
 					}
 				}
 				mtlsauth := mtls.NewMTLSAuthentication(regex, source)
+
+				blockList := blocklist.New()
+				fingerprints, err := blocklist.LoadFromConfigMap(ctx, kubeConfig.Clientset, namespace)
+				if err != nil {
+					logrus.Warnf("Could not load TLS blocklist: %v", err)
```
jgwest: Should fail if blocklist can't be loaded, as in exception error cases this might allow previously blocked clients to connect.

cmd/ctl/agent.go
```diff
@@ -354,11 +355,18 @@ func NewAgentInspectCommand() *cobra.Command {
 			if err != nil {
 				cmdutil.Fatal("Not a valid certificate: %v", err)
 			}
+
+			certObj, parseErr := x509.ParseCertificate(cert.Certificate[0])
```
jgwest, line 365: Should we report an error if len(cert.Certificate) != 1?

internal/tlsutil/tlsutil.go
```diff
@@ -130,6 +136,32 @@ func KeyDataToPEM(k crypto.PrivateKey) (string, error) {
...
+func FingerprintFromContext(ctx context.Context) string {
+	p, ok := peer.FromContext(ctx)
+	if !ok {
+		return ""
+	}
+	tlsInfo, ok := p.AuthInfo.(credentials.TLSInfo)
+	if !ok || len(tlsInfo.State.VerifiedChains) < 1 {
+		return ""
+	}
+	return CertificateFingerprint(tlsInfo.State.VerifiedChains[0][0])
```
jgwest, line 162: I see us accessing via `[0][0]` on VerifiedChains, should we return an error if this variable does not have the expected shape?

principal/apis/auth/auth.go
```diff
@@ -151,7 +153,15 @@ func (s *Server) Authenticate(ctx context.Context, ar *authapi.AuthRequest) (*au
 		return nil, status.Errorf(codes.FailedPrecondition, "version mismatch")
 	}
 
-	logCtx.WithField("client", clientID).WithField("agent_version", agentVersion).Info("client authentication successful")
+	logFields := logrus.Fields{"client": clientID, "agent_version": agentVersion}
+	var fingerPrint string
+	if ar.Method == "mtls" {
```
jgwest: Is there a constant we can use instead of `mtls`?

principal/server.go
```diff
@@ -863,6 +893,17 @@ func (s *Server) Start(ctx context.Context, errch chan error) error {
 		}
 	}
 
+	// Start the blocklist informer
+	if s.blocklistInformer != nil {
+		go func() {
+			if err := s.blocklistInformer.Start(s.ctx); err != nil {
+				log().WithError(err).Error("Blocklist informer has exited non-successfully")
```
jgwest: If we can't start the blocklist informer that sounds like it should be fatal, as we won't be informed of any blocklist changes.

principal/apis/auth/auth.go
```diff
@@ -151,7 +153,15 @@ func (s *Server) Authenticate(ctx context.Context, ar *authapi.AuthRequest) (*au
 		return nil, status.Errorf(codes.FailedPrecondition, "version mismatch")
 	}
 
-	logCtx.WithField("client", clientID).WithField("agent_version", agentVersion).Info("client authentication successful")
+	logFields := logrus.Fields{"client": clientID, "agent_version": agentVersion}
+	var fingerPrint string
```
jgwest: `fingerPrint` -> `fingerprint`

principal/server.go
```diff
@@ -92,6 +93,11 @@ type Server struct {
 	server      *http.Server
 	grpcServer  *grpc.Server
 	authMethods *auth.Methods
+	blocklist   *blocklist.Blocklist
+	// blocklistInformer watches the TLS blocklist ConfigMap for changes
+	blocklistInformer *informer.Informer[*corev1.ConfigMap]
```
jgwest: This suggestion is entirely optional, entirely up to you if you think it's worth bothering:
- At the moment, in the code (over time) we've added a ton of new fields to `Server` struct in server.go; by my count currently `60` total
- It can thus be helpful to consolidate any new fields we add, to prevent the problem from getting worse
- My suggestion in this case would be to move `blocklistInformer` and `fingerprintToAgent` into `blocklist`'s struct, since all 3 fields are related to blocklist functionality.

principal/callbacks.go
```diff
@@ -1146,3 +1147,71 @@ func (c *concurrentMap[K, V]) Delete(key K) {
...
+// that were not present in the previous blocklist. It resolves the agent name
+// from the in-memory fingerprint-to-agent mapping.
+func (s *Server) disconnectNewlyBlocklisted(fingerprints []string, oldFingerprints map[string]bool) {
+	if s.eventStreamSrv == nil {
+		return
+	}
+	for _, fp := range fingerprints {
+		if oldFingerprints[fp] {
+			continue
+		}
```
jgwest: Rather than skipping oldFingerprints, would it make sense to just disconnect all of fingerprints each time?

Is this done for performance reasons? I wouldn't expect this area of the code to be a performance bottleneck.

principal/apis/auth/auth.go
jparsai: There is no existing constant, I created new for all 3 auth methods.

cmd/ctl/agent.go
```diff
@@ -354,11 +355,21 @@ func NewAgentInspectCommand() *cobra.Command {
 			if err != nil {
 				cmdutil.Fatal("Not a valid certificate: %v", err)
 			}
+
+			if len(cert.Certificate) == 0 {
```
jgwest, line 359: We're currently checking if len == 0, but we only use the first entry in the certificate below, so I assume we instead want to check if len == 1

internal/tlsutil/tlsutil.go
```diff
@@ -130,6 +136,32 @@ func KeyDataToPEM(k crypto.PrivateKey) (string, error) {
...
+// FingerprintFromContext extracts the client certificate from a gRPC TLS
+// context and returns its SHA-256 fingerprint. Returns empty string if the
+// certificate cannot be extracted.
+func FingerprintFromContext(ctx context.Context) string {
+	p, ok := peer.FromContext(ctx)
+	if !ok {
+		return ""
+	}
+	tlsInfo, ok := p.AuthInfo.(credentials.TLSInfo)
+	if !ok || len(tlsInfo.State.VerifiedChains) < 1 || len(tlsInfo.State.VerifiedChains[0]) < 1 {
```
jgwest, line 159: Since we are using [0][0], I presume we would want to verify that len == 1 for both slices, rather than the logic here

principal/server.go
```diff
@@ -867,6 +891,18 @@ func (s *Server) Start(ctx context.Context, errch chan error) error {
 		}
 	}
 
+	// Start the blocklist informer
+	if s.blocklist != nil && s.blocklist.Informer != nil {
+		go func() {
+			if err := s.blocklist.Informer.Start(s.ctx); err != nil {
+				log().WithError(err).Error("Blocklist informer has exited non-successfully")
+				errch <- fmt.Errorf("blocklist informer failed: %w", err)
```
jgwest: Should we instead make this fatal? (Or is it fatal already via errch?)

principal/callbacks.go
```diff
@@ -1144,3 +1145,59 @@ func (c *concurrentMap[K, V]) DeleteByValue(value V, eq func(a, b V) bool) {
...
+// addBlocklistCallback is called by the informer when the blocklist ConfigMap
+// is first added.
+func (s *Server) addBlocklistCallback(cm *corev1.ConfigMap) {
+	s.updateBlocklistCallback(nil, cm)
+}
+
+// updateBlocklistCallback is called by the informer when the blocklist
+// ConfigMap is updated. It replaces the in-memory blocklist with
+// the entries from the ConfigMap and disconnects all blocklisted agents.
+func (s *Server) updateBlocklistCallback(oldCM, newCM *corev1.ConfigMap) {
```
jgwest: `oldCM` is unused, can either remove the param or replace with `_`

internal/tlsutil/tlsutil_test.go
```diff
@@ -424,3 +430,98 @@ func Test_SetTLSConfigFromFlags(t *testing.T) {
 		assert.Equal(t, []uint16{cs.ID}, cfg.CipherSuites, "cipherSuites should be preserved")
 	})
 }
+
+func Test_CertificateFingerprint(t *testing.T) {
+	t.Run("returns colon-separated uppercase SHA-256 hex", func(t *testing.T) {
```
jgwest: Flagged by Claude: 
> Test name says 'returns colon-separated uppercase SHA-256 hex' but CertificateFingerprint returns hex WITHOUT colons (e.g.,'AABB...' not 'AA:BB:...').

jgwest (changes requested):

cmd/ctl/agent.go
jparsai: Yes that is also correct, but it doesn't avoid the condition for 0 check. I am checking if length is zero, then it throws error and returns before itself before parsing. In current implementation secret can not have more than 1 entries (if created by CLI). If user manually updated secret and added new entry, we will use first entry only. I can check for len>=1, but still I need to add else block to see if it is 0 then trow error:

```
if len>=1 {
    certObj, parseErr := x509.ParseCertificate(cert.Certificate[0])
} else {
    cmdutil.Fatal("Certificate chain is empty")
}
```
vs
```
if len==0 {
    cmdutil.Fatal("Certificate chain is empty")
}
certObj, parseErr := x509.ParseCertificate(cert.Certificate[0])
```
If you still want me to change it, I can update.

internal/tlsutil/tlsutil.go
jparsai: Same as above. I need to return empty string if length is 0. So I can have condition len == 1 but I still need to add else block. Here is comparison of both conditions:

```
if len == 1 {
    return CertificateFingerprint(tlsInfo.State.VerifiedChains[0][0])
} else {
    return ""
}

```
vs 

```
if len == 0 {
    return ""
}
return CertificateFingerprint(tlsInfo.State.VerifiedChains[0][0])

```

principal/server.go
jparsai: Yes it is fatal. Fetal is not used because that would simply exit the goroutine, but it will not stop the server. This approach will shutdown the server as well. Same logic used by the gRPC server errors.

cmd/ctl/agent.go
jgwest: > If user manually updated secret and added new entry, we will use first entry only.

IMHO this is the bug. If  'user manually updated secret and added new entry', we should not silently ignore that new entry (because that is likely not what the user expected would happen. They likely expected we would use it for something). 

We should instead fail with an error reporting that there are too many entries, and that we only use the first. At that point it's up to the user to correct their mistake.

OTOH if you are saying there is a valid use case for having multiple entries, then we should USE those multiple entries in our logic, rather than just defaulting to the first.

internal/tlsutil/tlsutil.go
jgwest: Likewise same as above, why are we silently ignoring any other entries in verifiedChains after the first?

Or, said another way, if the cert does not have the _exact_ format we are currently expect, we should reject it, rather than assuming that any other entries in verifiedChains are not useful (or assuming that the _first_ entry is the one that is useful)

principal/server.go
jgwest: Fatal calls `os.Exit()`, which will necessarily terminate the process, not just the goroutine (at least for all the fatal implementations I can find in use in the code)

> This approach will shutdown the server as well. Same logic used by the gRPC server errors.

I'm not seeing it. It seems like `errch` is not actually used for anything. AFAICT if an error occurs it will only shutdown blocklist informer.

cmd/ctl/agent.go
jparsai: I updated it and now throwing error if user manually added more than one elements.

internal/tlsutil/tlsutil.go
jparsai: This can actually have multiple elements based on the configurations. We need first one which is `PeerCertificates`.
https://pkg.go.dev/crypto/tls#ConnectionState

principal/server.go
jparsai: Yes this makes sense. I followed pattern as given in listen.go https://github.com/argoproj-labs/argocd-agent/blob/main/principal/listen.go#L237, but updated to just use Fetal now.

jannfis (commented):
I have another minor comment about UX

cmd/ctl/blocklist.go
jparsai: It is updated. PTAL

jgwest (approved):
LGTM, thanks @jparsai!

jannfis (approved):
LGTM, thanks!

PR #1037, opened 2026-07-15
feat: allow agents to periodically resync with the principal

### LINKED ISSUE #102: Agent should periodically resync apps
An agent should periodically resync apps with the principal in case an event got lost somewhere.

The period duration should be configurable per agent, and the principal should enforce a global minimum period duration, for example, each agent can request their own resync period. If that requested period is lower than the global minimum, the principal should refuse the request.

enbiyagoral:

Resync currently happens only once, when the agent connects for the first time after startup (`resyncOnStart` is guarded by a flag that is never reset). If an event is lost afterwards — for example while the agent is disconnected, or dropped anywhere in transit — the agent and the principal stay out of sync until the agent pod is restarted.

This PR adds an optional periodic resync, per #102:

- **Agent**: a new `--resync-interval` flag (`ARGOCD_AGENT_RESYNC_INTERVAL`) re-runs the mode-appropriate resync mechanism (the same one used on startup) at the given interval, while the agent is connected. Disabled by default (`0`), so existing behavior is unchanged unless opted in.
- **Principal**: a new `--resync-min-interval` flag (`ARGOCD_PRINCIPAL_RESYNC_MIN_INTERVAL`) enforces a global minimum interval between resync rounds requested by any single agent (`SyncedResourceList` and `RequestResourceResync`). Rounds arriving more often are refused with a warning and dropped. Also disabled by default.

A design note: #102 suggests agents could negotiate their resync period with the principal, with the principal refusing periods below the global minimum. To avoid extending the wire protocol, this PR implements the enforcement at request time instead: the principal refuses resync rounds that arrive more often than the configured minimum, which provides the same protection. Happy to rework towards handshake-based negotiation if that's preferred.

Since the resync mechanism is checksum-based, a periodic round on an in-sync system only exchanges request messages and does not re-transmit resource contents.

Fixes #102

1. Start the principal with `--resync-min-interval 1m` and an agent with `--resync-interval 2m`.
2. Delete or modify a resource on the receiving side while the agent is disconnected, then reconnect: within the next resync round, the resource converges back without restarting the agent.
3. Set the agent's interval below the principal's minimum (e.g. `--resync-interval 10s`): the principal logs "Refusing resync request" and drops the extra rounds.

Unit tests cover the new agent option validation, the per-agent rate limiting logic, and refusal of too-frequent rounds for both managed (`SyncedResourceList`) and autonomous (`RequestResourceResync`) request types.

**Checklist**

## Summary by CodeRabbit

* **New Features**
  * Added optional periodic, checksum-based resource resync for agents via `--resync-interval` / `ARGOCD_AGENT_RESYNC_INTERVAL` (disabled by default with `0`).
  * Added principal-side throttling via `--resync-min-interval` / `ARGOCD_PRINCIPAL_RESYNC_MIN_INTERVAL` to refuse overly frequent resync requests.
* **Bug Fixes**
  * Prevented overlapping resync rounds and only mark startup resync as completed after a successful request.
* **Metrics**
  * Added `argocd_principal_resync_requests_total` metric for resync requests (accepted vs refused).
* **Documentation**
  * Updated agent and principal configuration references with new flags, defaults, and guidance.
* **Tests**
  * Added/extended unit and end-to-end coverage for periodic resync behavior and throttling.


principal/server.go
```diff
@@ -170,6 +170,16 @@ type Server struct {
...
+
+	// resyncMinInterval is the global minimum interval the principal enforces
+	// between resync rounds requested by any single agent. Resync requests
+	// arriving more often are refused. A zero value disables the enforcement.
+	resyncMinInterval time.Duration
+	// lastResyncRound tracks when each agent's last accepted resync round
+	// happened (key: agent name). lastResyncRoundMu must be held when
+	// accessing it.
+	lastResyncRound   map[string]time.Time
+	lastResyncRoundMu sync.Mutex
```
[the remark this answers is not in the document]
enbiyagoral: Good catch, this was real — a reconnect within the minimum interval would have its startup resync silently refused, with no recovery path when periodic resync is disabled.

Fixed in f01818d: the principal now clears the agent's recorded resync round when the agent (re)connects. I deliberately did not put the reset into `handleResyncOnConnect`: the on-connect handlers run asynchronously (via the `notifyOnConnect` channel consumed by `RunHandlersOnConnect`), so the agent's startup resync request could race ahead of the reset. Instead, the event stream server now supports a synchronous `WithOnConnect` callback which is invoked in `Subscribe` before the receive goroutine for that connection starts, so the reset provably happens before any event from the new connection is processed. The minimum interval therefore only throttles rounds within the lifetime of a single connection, which is the behavior the flag intends.

enbiyagoral:
Addressed the review feedback in f01818d:

- **Startup resync dropped after quick reconnect (critical)**: fixed — the principal resets the agent's resync allowance on (re)connect via a synchronous on-connect callback in the event stream server, so startup resyncs are always accepted and the minimum interval only throttles rounds within a single connection. See the inline reply for why the reset is not done in `handleResyncOnConnect`.
- **Overlapping resync rounds (nitpick)**: fixed — the agent now holds an in-flight flag for the whole round (including the managed-mode `SendRequestUpdates` goroutine), so ticks that fire while a round is still running are skipped.
- **Bulk `RequestUpdate` events bypassing the rate limit in managed mode**: acknowledged, and I'd propose handling it in a follow-up. Fully closing it requires the agent to wait for the principal to accept the round before generating bulk updates, i.e. a request/response handshake — a wire protocol addition this PR deliberately avoids. Two mitigations are in place meanwhile: the new in-flight guard prevents a well-behaved agent from stacking rounds, and on an in-sync system each `RequestUpdate` is a checksum comparison with no payload retransmission. Happy to file the follow-up issue for the handshake if the approach sounds right.

docs/configuration/reference/agent.md
```diff
@@ -288,6 +288,24 @@ Increase this value on large clusters or clusters under high API server load whe
...
+| **Environment Variable** | `ARGOCD_AGENT_RESYNC_INTERVAL` |
+| **ConfigMap Entry** | N/A |
+| **Type** | Duration |
+| **Default** | `0` (disabled) |
+
+Interval at which the agent periodically resyncs its resources with the principal, in addition to the resync that happens once on startup. Periodic resync recovers state that would otherwise stay out of sync when an event is lost, for example while the agent was disconnected.
+
+The resync mechanism is checksum-based: when both sides are already in sync, a resync round only exchanges request messages and does not re-transmit resource contents.
+
+Note that the principal may refuse resync rounds that arrive more often than its configured `--resync-min-interval`, so this interval should be set to a higher value than the principal's minimum.
```
jparsai, line 305: It would be better to handle it in implementation. If someone skips doc and configure wrong interval then, principal will silently refuse and agent will have no idea about that. In this case agent will assume resync completed successfully and users will never notice that they provided wrong configurations. Maybe you can send and ack back to agent.

jparsai:
> * **Bulk `RequestUpdate` events bypassing the rate limit in managed mode**: acknowledged, and I'd propose handling it in a follow-up. Fully closing it requires the agent to wait for the principal to accept the round before generating bulk updates, i.e. a request/response handshake — a wire protocol addition this PR deliberately avoids. Two mitigations are in place meanwhile: the new in-flight guard prevents a well-behaved agent from stacking rounds, and on an in-sync system each `RequestUpdate` is a checksum comparison with no payload retransmission. Happy to file the follow-up issue for the handshake if the approach sounds right.

If you want to fix in follow-up PR, then it would be better to create new or update same issue and provide the approach / implementation logic you will follow to fix it. This way we know what is coming next after this PR is merged and if that fix in good enough.

jparsai (commented):
- There should be metrics for refused/accepted requests
- It would be nice to have e2e tests to cover this. There are existing resync tests available for your reference.

docs/configuration/reference/agent.md
enbiyagoral: Agreed — a misconfigured interval should not fail silently. Two parts to the answer:

**Now (in 8200efb):** the refusal is observable and alertable on the principal: it logs a warning and increments the new `argocd_principal_resync_requests_total{result="refused"}` counter (also added to the metrics reference docs).

**Agent-side ack:** making the refusal visible to the *agent* needs a response event for the resync request — and that is exactly the handshake that also fixes the bulk `RequestUpdate` bypass CodeRabbit flagged. Rather than adding a one-off ack signal here that the handshake would immediately replace, I've specced the full approach (event pair, agent flow with bounded wait and retry-after, version-skew fallbacks) in #1046 as the follow-up. Happy to adjust the approach there, or pull it into this PR if you'd rather not split it.

enbiyagoral:
@jparsai thanks for the review — addressed in 8200efb:

- **Metrics**: added `argocd_principal_resync_requests_total`, a counter partitioned by `agent_name` and `result` (`accepted`/`refused`), recorded on every resync round decision. Documented in `docs/operations/metrics.md`.
- **E2e test**: added `Test_PeriodicResyncRounds` to the existing resync suite. It restarts the principal and the managed agent with `ARGOCD_AGENT_RESYNC_INTERVAL=5s` / `ARGOCD_PRINCIPAL_RESYNC_MIN_INTERVAL=30s` via the e2e env file, then asserts through the principal's metrics endpoint that (a) periodic rounds arriving within the minimum interval are refused, (b) a round is accepted again once the interval elapses without any reconnect, and (c) a managed app stays intact throughout. Configuration is restored after the test.
- **Follow-up issue**: filed #1046 with the approach and implementation logic for the resync round handshake, which covers both the bulk `RequestUpdate` bypass and agent-side visibility of refusals (your inline comment).

PR #1043, opened 2026-07-20
feat(helm/agent): add deploymentAnnotations value for Deployment metadata annotations

runderwoodcr14:
The agent chart exposes `podAnnotations` (pod template) but no way to set annotations on the Deployment object itself. Tools like [Stakater Reloader](https://github.com/stakater/Reloader) read their trigger annotations (e.g. `secret.reloader.stakater.com/reload` for TLS secret rotation restarts) from the **workload's** `metadata.annotations` only — so cert-rotation-driven agent restarts can't currently be wired via the chart.

This adds an optional `deploymentAnnotations` value rendered into the agent Deployment's `metadata.annotations` (empty by default — no behavior change), mirroring the existing `podAnnotations` pattern. Use case: mTLS client-cert rotation with cert-manager + Reloader bouncing the agent onto the renewed cert.

Note: this re-submits #1025, which was auto-closed when the source fork was accidentally deleted. All review feedback from #1025 is incorporated (thanks @jannfis), including the helm-docs README regeneration that cleared the last failing check.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring custom annotations on the agent Deployment through Helm values.
  * Documented the new `deploymentAnnotations` setting, which defaults to an empty object.

* **Documentation**
  * Updated the Helm chart version from 0.2.5 to 0.2.6.
  * Updated the chart README to include the new configuration option.


jannfis (approved):
LGTM, thanks

PR #1044, opened 2026-07-20
fix(principal): fall back to full update when granular status patch is rejected

runderwoodcr14:

The principal's `UpdateStatus` builds a granular JSON patch (via jsondiff) to update an application's status. That patch can be rejected by the API server with 400 or 422: when the diff source carries zero-valued non-pointer structs that the sparse live status doesn't have (Go's `encoding/json` ignores `omitempty` on non-pointer structs, e.g. `SyncStatus.ComparedTo`), the generated patch operations target parents that don't exist on the live object.

When this happens there is currently no recovery path — the same patch is rebuilt and rejected on every retry, and status updates for the affected application stop flowing until manual intervention. We hit this in production, with one application's status wedged for about 90 minutes.

This PR adds a fallback: when the patch is rejected with 422 (`IsInvalid`) or 400 (`IsBadRequest`), the update is retried once through the existing full-update path (`patchFn = nil`), which replaces status/annotations/labels/operation wholesale and involves no JSON diffing. All other error classes keep their existing behavior, and the happy path is unchanged.

No existing issue — happy to file one first if that's preferred.

Three new unit tests in `internal/manager/application` cover the behavior:
- a 422 (`IsInvalid`) patch rejection falls back to the full update
- a 400 (`IsBadRequest`) patch rejection falls back to the full update
- non-4xx errors (e.g. an internal error) do not trigger the fallback

The two anonymous update/patch callbacks in `UpdateStatus` are extracted into named variables (`updateFn`, `patchFn`) so the fallback can reuse them — no functional change to either.

**Checklist**

Note: this re-submits #1042, which was auto-closed when the source fork was accidentally deleted; content is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved application status updates when granular API patches are rejected.
  * Automatically retries with a full update for client-side validation or bad-request errors, helping status changes complete successfully.
  * Preserves existing error behavior for unrelated server or internal failures.

### DIFF
--- a/internal/manager/application/application.go
+++ b/internal/manager/application/application.go
@@ -589,12 +589,13 @@ func (m *ApplicationManager) UpdateStatus(ctx context.Context, namespace string,
 		return nil, fmt.Errorf("UpdateStatus should only be called on principal")
 	}
 
-	updated, err = m.update(ctx, false, incoming, func(existing, incoming *v1alpha1.Application) {
+	updateFn := func(existing, incoming *v1alpha1.Application) {
 		existing.Annotations = incoming.Annotations
 		existing.Labels = incoming.Labels
 		existing.Status = *incoming.Status.DeepCopy()
 		existing.Operation = incoming.Operation
-	}, func(existing, incoming *v1alpha1.Application) (jsondiff.Patch, error) {
+	}
+	patchFn := func(existing, incoming *v1alpha1.Application) (jsondiff.Patch, error) {
 		refresh, incomingRefresh := incoming.Annotations["argocd.argoproj.io/refresh"]
 		_, existingRefresh := existing.Annotations["argocd.argoproj.io/refresh"]
 		target := &v1alpha1.Application{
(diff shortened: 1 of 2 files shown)


jannfis:
Hm, I'm not quite convinced whether dynamically switching from patch to update on error would be the right thing to do to be honest.

An update replaces the whole resource eventually, and at a high rate of updates to other resources of the same kind, might fail (i.e. if the same kind of resource is updated or patched between the retrieval, change and update of any given resource).

Is there any chance your problem could be resolved by producing a better patch?

runderwoodcr14:
Thanks for taking a look, @jannfis — fair questions, let me try to address both.

> at a high rate of updates to other resources of the same kind, might fail (i.e. if the same kind of resource is updated or patched between the retrieval, change and update of any given resource)

The fallback reuses the existing full-update path in `update()`, which already retrieves a fresh copy from the API server on every attempt (the `ForUpdateContextKey` GET bypasses the informer cache) and wraps the whole read-modify-write in `retry.RetryOnConflict`. A concurrent write to the same app between the read and the write yields a 409 and the attempt restarts from a fresh read, so nothing is silently overwritten. It's also a one-shot recovery rather than a mode switch: the granular patch stays the primary path, and the fallback only fires on a 400/422 rejection — which today wedges the app permanently (the same patch is rebuilt and rejected on every retry; in our production case status updates for the affected app stopped flowing for ~90 minutes until manual intervention).

> Is there any chance your problem could be resolved by producing a better patch?

I dug into that first, and I don't believe a reliably-valid RFC 6902 patch can be produced from the typed objects. The problem isn't staleness — the diff base is already a live GET — it's that decoding the stored JSON into the Go types erases the difference between "absent" and "zero-valued" for non-pointer structs (`encoding/json` ignores `omitempty` on them, e.g. `SyncStatus.ComparedTo`). Re-marshaling for the diff then manufactures parents that don't exist in the stored object, so jsondiff can emit ops like `add /status/sync/comparedTo/...` whose parent is missing server-side — and RFC 6902 `add`/`replace` hard-fail on that, with no "create intermediate parents" semantics. Since the client can't reconstruct the stored document's actual shape, any granular 6902 patch it builds can be invalid for some live-object shapes. The existing zero-status special case in `UpdateStatus` (prepending `replace /status` when the existing status is empty) already band-aids one instance of this class; the wedge we hit is a partial-status instance it doesn't catch.

What *would* fix patch generation structurally is building this update as a JSON **merge** patch (RFC 7386) instead — e.g. `jsonpatch.CreateMergePatch` over the same source/target pair. That stays minimal, creates missing parents instead of erroring, treats removal of an already-absent key as a no-op (which also covers the refresh-annotation `remove` op racing an external annotation change), and replaces changed subtrees wholesale — which matches the intent here anyway, since `UpdateStatus` is by definition making status/operation match the agent's version. The reason I didn't do that in this PR: the backend `Patch` is pinned to `types.JSONPatchType`, so it needs a patch-type parameter or a new method through the backend interface (+ mocks), and it changes the wire format of every status update — a noticeably bigger change than the 4-line fallback.

Happy to go whichever way you prefer: keep the one-shot fallback as-is, rework this PR to build the status update as a merge patch, or land the fallback now and I'll file an issue + follow-up PR for the merge-patch rework. Let me know and I'll adjust.

PR #962, opened 2026-05-28
fix: UI navigation to child apps fails in autonomous app-of-apps

jparsai:
This PR is to fix https://github.com/argoproj-labs/argocd-agent/issues/949

Issue is caused because in app-of-apps scenario in autonomous agents, parent application `status.resources` has the agent instance namespace (argocd-autonomous) for child application entries and due to this the link generated by UI is not correct as principal does not have that namespace. In principal, resources are stored in namespace same as agent name (agent-autonomous).

Assisted by: Cursor

## Release Notes

* **Bug Fixes**
  * Fixed UI navigation for child applications in autonomous agent mode. Namespace references are now correctly rewritten to use the principal-side namespace for both create and update operations.

* **Tests**
  * Added test coverage validating namespace handling in autonomous-mode application processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/argoproj-labs/argocd-agent/pull/962?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

Issue #949, opened 2026-05-13
UI navigates to child apps with agent-side namespace instead of principal-side namespace (app-of-apps + autonomous mode)


SamuelMolling:
## Describe the bug

When using ArgoCD Agent in **autonomous mode** with the **app-of-apps pattern**, clicking on a child Application in the ArgoCD UI on the principal cluster results in `"permission denied"` because the UI constructs the request using the **agent-side namespace** (`argocd`) instead of the **principal-side namespace** (`argocd-agent-<agent-name>`).

The resource tree of the parent app loads correctly, but navigation to any child app fails.

## Setup

- **ArgoCD version:** v3.3.8 (Helm chart 9.5.4)
- **ArgoCD Agent version:** v0.8.1
- **Agent mode:** autonomous
- **`destination_based_mapping`:** enabled
- **`--application-namespaces=argocd-agent-*`** configured on `argocd-server`
- Principal cluster: CI/CD (GKE)
- Agent cluster: staging (GKE)

## Steps to reproduce

1. Deploy ArgoCD Agent in autonomous mode on a spoke cluster
2. On the agent cluster, create an app-of-apps Application in the `argocd` namespace that generates child Applications (e.g. via ApplicationSet or Kustomize)
3. The agent syncs the parent and child apps to the principal, where they land in the `argocd-agent-staging` namespace
4. Open the parent app in the ArgoCD UI on the principal — the resource tree loads correctly
5. Click on any child Application in the resource tree (e.g. `staging-admin-dashboard`)
6. The UI requests the child app with `appNamespace=argocd` (the agent-side namespace)
7. Result: `{"error":"permission denied","code":7,"message":"permission denied"}`

## Expected behavior

When clicking on a child Application in the resource tree, the UI should translate the agent-side namespace (`argocd`) to the principal-side namespace (`argocd-agent-staging`) before making the API request. The child app should open correctly.

## Actual behavior

The UI uses the namespace from the resource tree verbatim, which reflects the agent-side namespace (`argocd`). Since the app doesn't exist in the `argocd` namespace on the principal, ArgoCD returns "permission denied" (it masks "not found" as "permission denied" to prevent enumeration).

**Server logs:**
```
level=warning msg="application does not exist" application=staging-admin-dashboard namespace=argocd user=admin
[...]

PR #1048, opened 2026-07-22
fix: rewrite ApplicationSet/Application namespace in cached resource tree for agent-owned apps

### LINKED ISSUE #1040: Hub-local parent app-of-apps: child app links use wrong appNamespace (404) when children are managed by an autonomous agent
## Description

When running `argocd-agent` in **hybrid mode** with a **hub-local parent app-of-apps** whose children are managed as **autonomous agent** applications on a spoke cluster, child application links generated from the parent's resource tree in the ArgoCD UI are broken. Clicking a child app from the parent's resource tree builds a URL with the wrong `appNamespace` query parameter, resulting in a 404.

This looks related to #949 and #962, but appears to be a gap those fixes don't cover: both address namespace rewriting for **principal-managed** parent apps, not parents that are **hub-local** (created directly on the hub, never reconciled by the principal).

## Setup

- Hub cluster runs the principal and a hub-local ArgoCD instance.
- A parent "app-of-apps" Application lives directly in the `argocd` namespace on the hub cluster. It is **not** managed by the principal/agent - it's a plain hub-local Application.
- A spoke cluster runs an agent in **autonomous** mode. Its managed child Applications live in a dedicated namespace on the hub, e.g. `argocd-agent-staging`.
- `destination_based_mapping=true` is configured on the principal.
- The parent app's `spec.source` (or `generators`, if using an ApplicationSet-style app-of-apps) targets the child Applications, which live in the spoke's namespace.

## Symptom

In the ArgoCD UI, opening the parent app's resource tree shows the child Applications as expected. However, clicking a child Application node constructs a link like:

```
/applications/argocd/<child-app-name>?appNamespace=argocd
```

instead of the correct:

```
/applications/argocd-agent-staging/<child-app-name>?appNamespace=argocd-agent-staging
```

This 404s, since the child Application does not actually exist in the `argocd` namespace - it exists in the agent-specific namespace (`argocd-agent-staging` in this example).

**Workaround:** navigate to the child Application directly via URL using the correct `appNamespace`, bypassing the parent's resource tree links.

## Root Cause (as far as I can tell)

The namespace rewrite that maps a child resource's `appNamespace` to its correct agent-owned namespace only happens as part of `processApplicationEvent` on the principal. That rewrite path is triggered when an Application is reconciled through the principal.

Because the parent app-of-apps is **hub-local** - created directly on the hub and never touched by the principal/agent reconciliation - `processApplicationEvent` is never invoked for it. As a result, the parent's `Status.Resources` (which includes the child Application entries shown in the resource tree, along with their namespace hints) never gets rewritten to point at the child's actual namespace on the hub. The UI falls back to using the parent's own namespace (`argocd`) for the child links.

#949 and #962 both fix namespace rewriting for the case where the parent Application itself is managed by the principal (i.e., reconciled through `processApplicationEvent`). Neither appears to handle the hub-local-parent case, since the fix depends on that same event path running for the parent - which it never does here.

## Expected behavior

Child Application links generated from a hub-local parent's resource tree should resolve to the child's actual namespace on the hub, the same as they would if the parent were principal-managed.

## Environment

- argocd-agent: hybrid mode (hub-local Applications + autonomous agent-managed Applications)
- `destination_based_mapping=true`
- Related: #949, #962

m3rryqold:

For hybrid setups with an autonomous agent, the resource tree cache for an agent-owned Application is built by the agent using its own local namespace. That's correct on the agent's cluster, but not where the mirrored Application lives on the principal, so links built from this data in the UI point at the wrong namespace and 404/permission-deny for every child in the tree, direct children and ApplicationSet-generated grandchildren alike.

`Application.status.resources` already gets this rewrite in `principal/event.go` (#949, #962), but that only covers the flat status field, not the interactive resource tree served via `ResourceTree`/`WatchResourceTree` and cached in Redis, which is what the UI actually uses to render the graph and build links.

This adds the same rewrite for `Application`- and `ApplicationSet`-kind nodes in that cached tree, applied in `principal/redisproxy` where the cached bytes are forwarded back to the principal's Argo CD. Both kinds need it: the UI only draws a connecting edge between a parent and child node when their `Namespace` fields are equal, so rewriting only `Application` nodes left ApplicationSet-generated children with a mismatched namespace against their (unrewritten) ApplicationSet parent, visually disconnecting them from the tree even though their links were already correct.

Fixes #1040

Verified against a live hub/autonomous-spoke setup: built this fix into a test image, deployed it to the principal, and confirmed both the previously-broken child links now resolve to the correct namespace and ApplicationSet-generated children are visually connected to their parent in the resource tree again.

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected cached resource-tree responses so `Application` and `ApplicationSet` namespaces are aligned with the target agent.
  * Maintained support for both gzipped and uncompressed cached payloads when applying the namespace update.
  * Confirmed that non-`argoproj.io` namespaces (including orphaned nodes) are left unchanged, and malformed cached payloads fail safely.

### DIFF
--- a/principal/redisproxy/redisproxy.go
+++ b/principal/redisproxy/redisproxy.go
@@ -575,12 +575,6 @@ func (rp *RedisProxy) handleAgentGet(connState *connectionState, key string, arg
 
 		responseBytes := response.Get.Bytes
 
-		// The resource tree cache is built by the agent using its own (local) namespace
-		// for every Application-kind node, since that's genuinely where those objects
-		// live on the agent's cluster. The UI uses this data verbatim to build links in
-		// the resource tree, so for agent-owned Application nodes we rewrite Namespace to
-		// the agent's namespace on the principal, mirroring the same rewrite that's
-		// already applied to Application.status.resources in event.go.
 		if strings.HasPrefix(key, "app|resources-tree|") {
 			if rewritten, err := rewriteResourcesTreeNamespace(responseBytes, agentName); err != nil {
 				logCtx.WithError(err).Warn("unable to rewrite namespace in cached resource tree, forwarding as-is")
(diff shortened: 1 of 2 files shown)


jannfis (approved):
Hey @m3rryqold, sorry that it took me so long to get back to this PR. I apologize!

Thank you for this fix. LGTM.

PR #1049, opened 2026-07-23
feat: add SPIFFE/SPIRE integration for automated mTLS authentication

jparsai:
This PR is to add support for SPIFFE/SPIRE-based workload identity for authentication between principal and agent. 
This eliminates need to manually create certificates and secrets for each agent. Users can choose the authentication method via --spire-auth-method flag, supported values for are jwt (separate SPIRE Server per cluster, separate trust domains) and mtls (single SPIRE Server, shared trust domain). It is recommended to use this feature with self agent registration enabled, then it will be completely automated process from agent registration to authentication.

Few blogs/docs for reference
- https://www.redhat.com/en/topics/security/spiffe-and-spire#overview
- https://spiffe.io/docs/latest/planning/scaling_spire/
- https://github.com/spiffe/spire
- https://www.redhat.com/en/technologies/cloud-computing/openshift/zero-trust-workload-identity-manager

Assisted by: Cursor

## Summary by CodeRabbit

* **New Features**
  * Added optional SPIRE-based authentication and automatic TLS credential provisioning through mTLS or SPIFFE JWT.
  * Added Helm configuration for SPIRE sockets, authentication methods, and socket mounting.
  * Added shared client certificate issuance for principal self-registration.
  * Added SPIFFE JWT identity and audience validation.
* **Bug Fixes**
  * Improved TLS CA fallback when client secrets omit a CA certificate.
* **Documentation**
  * Added SPIRE setup, authentication, certificate management, and deployment guidance.
* **Tests**
  * Added end-to-end coverage for SPIRE authentication, connectivity, and terminal streaming.

### DIFF
--- a/cmd/argocd-agent/agent.go
+++ b/cmd/argocd-agent/agent.go
@@ -210,6 +210,15 @@ func NewAgentRunCommand() *cobra.Command {
 			if spireAuthMethod != "" && spireAuthMethod != "jwt" && spireAuthMethod != "mtls" {
 				cmdutil.Fatal("--spire-auth-method must be 'jwt' or 'mtls', got %q", spireAuthMethod)
 			}
+			if spireAgentSocket != "" && insecurePlaintext {
+				cmdutil.Fatal("--spire-agent-socket cannot be used with --insecure-plaintext; SPIRE provides TLS credentials")
+			}
+			if spireAgentSocket != "" && insecure {
+				cmdutil.Fatal("--spire-agent-socket cannot be used with --insecure-tls; SPIRE handles certificate verification")
+			}
+			if spireAgentSocket != "" && creds != "" {
+				cmdutil.Fatal("--creds cannot be used with --spire-agent-socket; SPIRE derives credentials from --spire-auth-method")
+			}
 
 			// Configure TLS: SPIRE, plaintext, or static certs
 			if spireAgentSocket != "" {
--- a/cmd/argocd-agent/agent.go
+++ b/cmd/argocd-agent/agent.go
@@ -216,9 +216,6 @@ func NewAgentRunCommand() *cobra.Command {
 			if spireAgentSocket != "" && insecure {
 				cmdutil.Fatal("--spire-agent-socket cannot be used with --insecure-tls; SPIRE handles certificate verification")
 			}
-			if spireAgentSocket != "" && creds != "" {
-				cmdutil.Fatal("--creds cannot be used with --spire-agent-socket; SPIRE derives credentials from --spire-auth-method")
-			}
 
 			// Configure TLS: SPIRE, plaintext, or static certs
 			if spireAgentSocket != "" {
--- a/cmd/argocd-agent/agent.go
+++ b/cmd/argocd-agent/agent.go
@@ -34,6 +34,7 @@ import (
 	"github.com/argoproj-labs/argocd-agent/internal/env"
 	"github.com/argoproj-labs/argocd-agent/internal/grpcutil"
 	"github.com/argoproj-labs/argocd-agent/internal/metrics"
+	"github.com/argoproj-labs/argocd-agent/internal/spire"
 	"github.com/argoproj-labs/argocd-agent/internal/tracing"
 	"github.com/argoproj-labs/argocd-agent/pkg/client"
 	"github.com/argoproj-labs/argocd-agent/pkg/types"
@@ -118,6 +119,10 @@ func NewAgentRunCommand() *cobra.Command {
 
 		// Adoption options
 		adoptionPolicy string
+
+		// SPIRE integration
+		spireAgentSocket string
+		spireAuthMethod  string
 	)
 	command := &cobra.Command{
 		Use:   "agent",
@@ -195,8 +200,40 @@ func NewAgentRunCommand() *cobra.Command {
 			}
 			var remote *client.Remote
 
-			// Configure TLS or plaintext mode
-			if insecurePlaintext {
+			// Validate SPIRE flags
+			if spireAgentSocket != "" && spireAuthMethod == "" {
+				cmdutil.Fatal("--spire-auth-method is required when --spire-agent-socket is set (use 'jwt' or 'mtls')")
+			}
+			if spireAuthMethod != "" && spireAgentSocket == "" {
+				cmdutil.Fatal("--spire-auth-method requires --spire-agent-socket to be set")
+			}
+			if spireAuthMethod != "" && spireAuthMethod != "jwt" && spireAuthMethod != "mtls" {
+				cmdutil.Fatal("--spire-auth-method must be 'jwt' or 'mtls', got %q", spireAuthMethod)
+			}
+
+			// Configure TLS: SPIRE, plaintext, or static certs
+			if spireAgentSocket != "" {
+				logrus.Infof("Using SPIRE for TLS credentials (socket: %s)", spireAgentSocket)
+				spireSource, err := spire.New(ctx, spireAgentSocket)
+				if err != nil {
+					cmdutil.Fatal("Failed to connect to SPIRE Agent: %v", err)
+				}
+				defer spireSource.Close()
+				remoteOpts = append(remoteOpts, client.WithSPIRE(spireSource, spireAuthMethod == "mtls"))
+				if creds == "" {
+					switch spireAuthMethod {
+					case "mtls":
+						creds = "mtls:"
+					case "jwt":
+						creds = "spiffe-jwt:"
+					}
+					authMethod, authCreds, err := parseCreds(creds)
+					if err != nil {
+						cmdutil.Fatal("Error setting up creds: %v", err)
+					}
+					remoteOpts = append(remoteOpts, client.WithAuth(authMethod, authCreds))
+				}
+			} else if insecurePlaintext {
 				// Plaintext mode - skip all TLS configuration (e.g., when behind Istio)
 				logrus.Warn("INSECURE: Connecting without TLS - ensure Istio or similar service mesh provides mTLS")
 				remoteOpts = append(remoteOpts, client.WithInsecurePlaintext())
@@ -503,6 +540,14 @@ func NewAgentRunCommand() *cobra.Command {
 		env.StringWithDefault("ARGOCD_AGENT_ADOPTION_POLICY", nil, "always"),
 		"Set the adoption policy for applications that already exist on a managed agent (always or never)")
 
+	command.Flags().StringVar(&spireAgentSocket, "spire-agent-socket",
+		env.StringWithDefault("ARGOCD_AGENT_SPIRE_AGENT_SOCKET", nil, ""),
+		"SPIRE Agent socket URI (e.g., unix:///run/spire/sockets/agent.sock). When set, TLS credentials are obtained from SPIRE instead of static certs")
+
+	command.Flags().StringVar(&spireAuthMethod, "spire-auth-method",
+		env.StringWithDefault("ARGOCD_AGENT_SPIRE_AUTH_METHOD", nil, ""),
+		"SPIFFE authentication method (required when --spire-agent-socket is set): 'jwt' uses JWT-SVIDs (for federated SPIRE), 'mtls' uses X.509-SVIDs (for centralized SPIRE)")
+
 	command.Flags().StringVar(&kubeConfig, "kubeconfig", "", "Path to a kubeconfig file to use")
 	command.Flags().StringVar(&kubeContext, "kubecontext", "", "Override the default kube context")
 	return command
@@ -524,6 +569,8 @@ func parseCreds(credStr string) (string, auth.Credentials, error) {
 		return auth.MethodUserPass, creds, nil
 	case auth.MethodMTLS:
 		return auth.MethodMTLS, auth.Credentials{}, nil
+	case auth.MethodSPIFFEJWT:
+		return auth.MethodSPIFFEJWT, auth.Credentials{}, nil
 	case auth.MethodHeader:
 		// Header-based auth doesn't require credentials - the sidecar/proxy injects the header
 		return auth.MethodHeader, auth.Credentials{}, nil
--- a/cmd/argocd-agent/agent.go
+++ b/cmd/argocd-agent/agent.go
@@ -34,6 +34,7 @@ import (
 	"github.com/argoproj-labs/argocd-agent/internal/env"
 	"github.com/argoproj-labs/argocd-agent/internal/grpcutil"
 	"github.com/argoproj-labs/argocd-agent/internal/metrics"
+	"github.com/argoproj-labs/argocd-agent/internal/spire"
 	"github.com/argoproj-labs/argocd-agent/internal/tracing"
 	"github.com/argoproj-labs/argocd-agent/pkg/client"
 	"github.com/argoproj-labs/argocd-agent/pkg/types"
@@ -118,6 +119,10 @@ func NewAgentRunCommand() *cobra.Command {
 
 		// Adoption options
 		adoptionPolicy string
+
+		// SPIRE integration
+		spireAgentSocket string
+		spireAuthMethod  string
 	)
 	command := &cobra.Command{
 		Use:   "agent",
@@ -195,8 +200,40 @@ func NewAgentRunCommand() *cobra.Command {
 			}
 			var remote *client.Remote
 
-			// Configure TLS or plaintext mode
-			if insecurePlaintext {
+			// Validate SPIRE flags
+			if spireAgentSocket != "" && spireAuthMethod == "" {
+				cmdutil.Fatal("--spire-auth-method is required when --spire-agent-socket is set (use 'jwt' or 'mtls')")
+			}
+			if spireAuthMethod != "" && spireAgentSocket == "" {
+				cmdutil.Fatal("--spire-auth-method requires --spire-agent-socket to be set")
+			}
+			if spireAuthMethod != "" && spireAuthMethod != "jwt" && spireAuthMethod != "mtls" {
+				cmdutil.Fatal("--spire-auth-method must be 'jwt' or 'mtls', got %q", spireAuthMethod)
+			}
+
+			// Configure TLS: SPIRE, plaintext, or static certs
+			if spireAgentSocket != "" {
+				logrus.Infof("Using SPIRE for TLS credentials (socket: %s)", spireAgentSocket)
+				spireSource, err := spire.New(ctx, spireAgentSocket)
+				if err != nil {
+					cmdutil.Fatal("Failed to connect to SPIRE Agent: %v", err)
+				}
+				defer spireSource.Close()
+				remoteOpts = append(remoteOpts, client.WithSPIRE(spireSource, spireAuthMethod == "mtls"))
+				if creds == "" {
+					switch spireAuthMethod {
+					case "mtls":
+						creds = "mtls:"
+					case "jwt":
+						creds = "spiffe-jwt:"
+					}
+					authMethod, authCreds, err := parseCreds(creds)
+					if err != nil {
+						cmdutil.Fatal("Error setting up creds: %v", err)
+					}
+					remoteOpts = append(remoteOpts, client.WithAuth(authMethod, authCreds))
+				}
+			} else if insecurePlaintext {
 				// Plaintext mode - skip all TLS configuration (e.g., when behind Istio)
 				logrus.Warn("INSECURE: Connecting without TLS - ensure Istio or similar service mesh provides mTLS")
 				remoteOpts = append(remoteOpts, client.WithInsecurePlaintext())
@@ -503,6 +540,14 @@ func NewAgentRunCommand() *cobra.Command {
 		env.StringWithDefault("ARGOCD_AGENT_ADOPTION_POLICY", nil, "always"),
 		"Set the adoption policy for applications that already exist on a managed agent (always or never)")
 
+	command.Flags().StringVar(&spireAgentSocket, "spire-agent-socket",
+		env.StringWithDefault("ARGOCD_AGENT_SPIRE_AGENT_SOCKET", nil, ""),
+		"SPIRE Agent socket URI (e.g., unix:///run/spire/sockets/agent.sock). When set, TLS credentials are obtained from SPIRE instead of static certs")
+
+	command.Flags().StringVar(&spireAuthMethod, "spire-auth-method",
+		env.StringWithDefault("ARGOCD_AGENT_SPIRE_AUTH_METHOD", nil, ""),
+		"SPIFFE authentication method (required when --spire-agent-socket is set): 'jwt' uses JWT-SVIDs (for federated SPIRE), 'mtls' uses X.509-SVIDs (for centralized SPIRE)")
+
 	command.Flags().StringVar(&kubeConfig, "kubeconfig", "", "Path to a kubeconfig file to use")
 	command.Flags().StringVar(&kubeContext, "kubecontext", "", "Override the default kube context")
 	return command
@@ -524,6 +569,8 @@ func parseCreds(credStr string) (string, auth.Credentials, error) {
 		return "userpass", creds, nil
 	case "mtls":
 		return "mtls", auth.Credentials{}, nil
+	case "spiffe-jwt":
+		return "spiffe-jwt", auth.Credentials{}, nil
 	case "header":
 		// Header-based auth doesn't require credentials - the sidecar/proxy injects the header
 		return "header", auth.Credentials{}, nil
--- a/cmd/argocd-agent/principal.go
+++ b/cmd/argocd-agent/principal.go
@@ -29,6 +29,7 @@ import (
 	"github.com/argoproj-labs/argocd-agent/internal/auth"
 	"github.com/argoproj-labs/argocd-agent/internal/auth/header"
 	"github.com/argoproj-labs/argocd-agent/internal/auth/mtls"
+	"github.com/argoproj-labs/argocd-agent/internal/auth/spiffejwt"
 	"github.com/argoproj-labs/argocd-agent/internal/auth/userpass"
 	"github.com/argoproj-labs/argocd-agent/internal/blocklist"
 	"github.com/argoproj-labs/argocd-agent/internal/config"
@@ -36,6 +37,7 @@ import (
 	"github.com/argoproj-labs/argocd-agent/internal/grpcutil"
 	"github.com/argoproj-labs/argocd-agent/internal/kube"
 	"github.com/argoproj-labs/argocd-agent/internal/labels"
+	"github.com/argoproj-labs/argocd-agent/internal/spire"
 	"github.com/argoproj-labs/argocd-agent/internal/tlsutil"
 	"github.com/argoproj-labs/argocd-agent/internal/tracing"
 	"github.com/argoproj-labs/argocd-agent/pkg/ha"
@@ -44,6 +46,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/spiffe/go-spiffe/v2/bundle/jwtbundle"
 )
 
 // NewPrincipalRunCommand returns a new principal run command.
@@ -131,6 +134,10 @@ func NewPrincipalRunCommand() *cobra.Command {
 		haAdminPort                    int
 		haAllowedReplClients           []string
 		haReplicationInitialAckTimeout time.Duration
+
+		// SPIRE integration
+		spireAgentSocket string
+		spireAuthMethod  string
 	)
 	command := &cobra.Command{
 		Use:   "principal",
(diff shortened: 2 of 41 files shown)


cmd/ctl/ca.go
```diff
@@ -520,6 +521,40 @@ func NewPKIIssueAgentClientCert() *cobra.Command {
...
+			if principalCfg.KubeContext == "" {
+				cmdutil.Fatal("Principal kubecontext must be set.")
+			}
+			keyOpts := parseKeyGenFlags(keyAlgorithm, keySize)
+			issueAndSaveSecret(principalCfg.KubeContext, config.SecretNameSharedClientCert, principalCfg.Namespace, upsert, days, keyOpts, func(c *x509.Certificate, pk crypto.PrivateKey) (string, string, error) {
+				return tlsutil.GenerateClientCertificate("argocd-agent-shared-client", c, pk, days, keyOpts)
+			})
+		},
+	}
+	command.Flags().BoolVarP(&upsert, "upsert", "u", false, "Whether to update an existing certificate if it exists")
```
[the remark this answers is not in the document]
jparsai: This is existing pattern used by all CLI commands. `issueAndSaveSecret()` is an existing function and not added by this PR.

test/e2e/fixture/spire.go
```diff
@@ -0,0 +1,272 @@
...
+	autonomous, err := fakespire.New(spireAutonomousAgentID, caCert, caKey, jwtKey)
+	if err != nil {
+		return nil, fmt.Errorf("create autonomous agent fake SPIRE server: %w", err)
+	}
+
+	return &FakeSpireEnvironment{
+		Principal:       principal,
+		ManagedAgent:    managed,
+		AutonomousAgent: autonomous,
+	}, nil
```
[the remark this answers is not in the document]
jparsai: This is fine for now. It is a fake environment for e2e test. Different trust domains will add complexity.

install/helm-repo/argocd-agent-principal/values.yaml
```diff
@@ -279,8 +279,24 @@ principal:
...
+    # "jwt" uses JWT-SVIDs (for federated SPIRE with multiple trust domains).
+    # "mtls" uses X.509-SVIDs (for centralized SPIRE with a single trust domain).
+    authMethod: ""
+    # -- SPIRE Agent socket URI passed to --spire-agent-socket.
+    socketPath: "unix:///run/spire/agent-sockets/spire-agent.sock"
+    # -- How to mount the SPIRE Agent socket into the pod.
+    # "hostPath" mounts the socket directory from the host.
+    mountMethod: "hostPath"
+    # -- Path to the SPIRE Agent socket directory. Used as the container mountPath and host path.
+    hostSocketDir: "/run/spire/agent-sockets"
```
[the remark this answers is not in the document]
jparsai: It is fixed. Check again.

test/e2e/fixture/spire.go
```diff
@@ -0,0 +1,272 @@
...
+		*spec.backup = secret
+	}
+
+	for _, spec := range specs {
+		if err := deleteSecret(ctx, spec.client, spec.name, spec.namespace); err != nil {
+			return nil, fmt.Errorf("delete secret %s/%s: %w", spec.namespace, spec.name, err)
+		}
+	}
+
+	return backups, nil
```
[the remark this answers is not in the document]
jparsai: Is fix not good enough yet?

install/helm-repo/argocd-agent-principal/values.yaml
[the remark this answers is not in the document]
jparsai: review again

test/e2e/application_test.go
```diff
@@ -70,11 +70,11 @@ func (s *ApplicationTestSuite) Test_ApplicationManagementAPI() {
 	requires.Eventually(func() bool {
 		createdApp, createErr = s.argoClient.CreateApplication(app)
 		if createErr != nil && strings.Contains(createErr.Error(), "project does not exist") {
-			s.T().Logf("Project does not exist in informer cache, retrying...")
+			s.T().Logf("Project does not exist in informer cache, retrying... (error: %v)", createErr)
 			return false
 		}
 		return true
-	}, 60*time.Second, 2*time.Second)
+	}, 120*time.Second, 2*time.Second)
```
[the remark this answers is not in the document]
jparsai: This is an existing test and not something added by this PR.

cmd/argocd-agent/agent.go
```diff
@@ -195,8 +200,40 @@ func NewAgentRunCommand() *cobra.Command {
...
+			}
+
+			// Configure TLS: SPIRE, plaintext, or static certs
+			if spireAgentSocket != "" {
+				logrus.Infof("Using SPIRE for TLS credentials (socket: %s)", spireAgentSocket)
+				spireSource, err := spire.New(ctx, spireAgentSocket)
+				if err != nil {
+					cmdutil.Fatal("Failed to connect to SPIRE Agent: %v", err)
+				}
+				defer spireSource.Close()
```
jgwest: JGW: check

cmd/argocd-agent/agent.go
```diff
@@ -503,6 +540,14 @@ func NewAgentRunCommand() *cobra.Command {
 		env.StringWithDefault("ARGOCD_AGENT_ADOPTION_POLICY", nil, "always"),
 		"Set the adoption policy for applications that already exist on a managed agent (always or never)")
 
+	command.Flags().StringVar(&spireAgentSocket, "spire-agent-socket",
```
jgwest: JGW: since this is running in container, how would this be set

cmd/ctl/ca.go
```diff
@@ -520,6 +521,40 @@ func NewPKIIssueAgentClientCert() *cobra.Command {
...
+func NewPKIIssueSharedClientCert() *cobra.Command {
+	var (
+		upsert       bool
+		days         int
+		keyAlgorithm string
+		keySize      int
+	)
+	command := &cobra.Command{
+		Use:   "shared-client",
+		Short: "Issue a shared client certificate for self-registration",
```
jgwest: Should we have NON-PROD warnings similar to other commands in this file?

docs/configuration/authentication.md
```diff
@@ -229,9 +230,74 @@ argocd-agent agent --creds="header:"
...
+```yaml
+# ConfigMap (argocd-agent-params)
+agent.spire.socket-path: "unix:///run/spire/agent-sockets/spire-agent.sock"
+agent.spire.auth-method: "mtls"
+# Creds defaults to mtls:
+```
+
+### Self-registration (required with SPIRE)
+
+With static certificates, agent registration happens when you run `argocd-agentctl pki issue agent <name>` — this creates the agent's cluster secret on the principal. With SPIRE, there are no per-agent certificate steps, so **self-registration must be enabled** on the principal. When an agent connects for the first time, the principal automatically creates the cluster secret.
```
jgwest: JGW: what is the secret that is created on principal?

docs/configuration/tls-certificates.md
```diff
@@ -46,6 +46,208 @@ graph TB
     ProxyCert --> ArgoCD[Argo CD Server]
 ```
 
+## Using SPIRE (Automated Certificate Management)
+
+[SPIRE](https://spiffe.io/docs/latest/spire-about/) can replace static certificate management for agent-to-principal communication. When enabled, TLS credentials are obtained automatically from the SPIRE Workload API — no per-agent certificate generation, no secret copying between clusters.
```
jgwest: `When enabled, TLS credentials are obtained automatically from the SPIRE Workload API — no per-agent certificate generation, no secret copying between clusters.`
->
`When enabled, TLS credentials are obtained automatically from the SPIRE Workload API. There is no per-agent certificate generation, and no secret copying between clusters. `

docs/configuration/tls-certificates.md
```diff
@@ -46,6 +46,208 @@ graph TB
     ProxyCert --> ArgoCD[Argo CD Server]
 ```
 
+## Using SPIRE (Automated Certificate Management)
```
jgwest: JGW: move spire into it's own MD?

docs/configuration/authentication.md
```diff
@@ -229,9 +230,74 @@ argocd-agent agent --creds="header:"
...
+
+```yaml
+# Environment variables on the principal
+ARGOCD_PRINCIPAL_ENABLE_SELF_CLUSTER_REGISTRATION: "true"
+ARGOCD_PRINCIPAL_SELF_REGISTRATION_CLIENT_CERT_SECRET: "argocd-agent-shared-client-tls"
+```
+
+The shared client certificate must be created beforehand using `argocd-agentctl pki issue shared-client`. See [TLS & Certificates](tls-certificates.md#using-spire-automated-certificate-management) for the complete setup steps.
+
+No static certificate secrets are needed on the agent cluster — SPIRE provides all credentials automatically.
```
jgwest: `No static certificate secrets are needed on the agent cluster. SPIRE provides all credentials automatically.`

docs/configuration/tls-certificates.md
```diff
@@ -46,6 +46,208 @@ graph TB
...
+| **JWT** | `jwt` | Agent fetches a JWT-SVID from its local SPIRE Agent and sends it as a bearer token. Principal validates the JWT using its SPIRE JWT bundle source. | Each cluster has its own SPIRE Server (separate trust domains). Both sides need JWT-SVID support. This is the most common setup with federated SPIRE. |
+| **mTLS** | `mtls` | Agent presents its X.509-SVID as a TLS client certificate. Principal verifies the cert chain against the SPIRE trust bundle and extracts identity from the SPIFFE URI in the certificate SAN. | All clusters share a single SPIRE Server (same trust domain). Uses only X.509-SVIDs, which are the core SPIFFE standard — works with any SPIFFE-compatible provider, even those that don't support JWT-SVIDs. |
+
+!!! tip "Which method should I choose?"
+    - If your clusters each have their own SPIRE Server and trust domain → use **JWT**
+    - If all clusters share a single SPIRE Server → use **mTLS**
+
+### Step 2: Create one-time secrets on the principal
+
+These static secrets are still required because the resource proxy and JWT signing use static TLS, and self-registration needs a shared client certificate:
```
jgwest: Should we add a message about the non-production use of these certificates?

docs/configuration/tls-certificates.md
```diff
@@ -46,6 +46,208 @@ graph TB
...
+| Environment Variable | Value | Description |
+|---|---|---|
+| `ARGOCD_PRINCIPAL_ENABLE_SELF_CLUSTER_REGISTRATION` | `true` | Enables automatic cluster secret creation when a new agent connects |
+| `ARGOCD_PRINCIPAL_SELF_REGISTRATION_CLIENT_CERT_SECRET` | `argocd-agent-shared-client-tls` | Name of the shared client cert secret (created in Step 2). The principal copies this into the agent's cluster secret so the ArgoCD application controller can authenticate to the resource proxy. |
+
+### Step 4: Configure the principal
+
+```yaml
+# ConfigMap (argocd-agent-params)
+principal.spire.socket-path: "unix:///run/spire/agent-sockets/spire-agent.sock"
```
jgwest: JGW: Update operator too?

docs/configuration/tls-certificates.md
```diff
@@ -46,6 +46,208 @@ graph TB
...
+principal.spire.auth-method: "jwt"   # or "mtls"
+```
+
+When SPIRE is enabled, the principal automatically:
+
+- Uses SPIRE X.509-SVIDs for its server TLS certificate (replaces `argocd-agent-principal-tls`)
+- Skips loading the static principal TLS secret and root CA
```
jgwest: JGW: But they are still required to exist?

docs/configuration/tls-certificates.md
```diff
@@ -46,6 +46,208 @@ graph TB
...
+
+```bash
+kubectl logs deploy/argocd-agent-agent -n argocd | grep -iE "spire|auth|jwt|svid"
+```
+
+### Configuration Reference
+
+| Parameter | Environment Variable | Description |
+|-----------|---------------------|-------------|
+| `principal.spire.socket-path` | `ARGOCD_PRINCIPAL_SPIRE_AGENT_SOCKET` | SPIRE Workload API socket URI on the principal |
```
jgwest: JGW: how many other places do we mention these?

install/helm-repo/argocd-agent-agent/templates/agent-deployment.yaml
```diff
@@ -281,6 +281,18 @@ spec:
                 name: {{ include "argocd-agent-agent.paramsConfigMapName" . }}
                 key: agent.label-selector
                 optional: true
+          - name: ARGOCD_AGENT_SPIRE_AGENT_SOCKET
+            valueFrom:
+              configMapKeyRef:
+                name: {{ include "argocd-agent-agent.paramsConfigMapName" . }}
```
jgwest: JGW-TODO: double check that these values match up with each other

install/helm-repo/argocd-agent-agent/Chart.yaml
```diff
@@ -5,7 +5,7 @@ description: Argo CD Agent for connecting managed clusters to a Principal
 type: application
 
 # chart version
-version: 0.2.6
+version: 0.2.7
 
 # application version, ArgoCD Agent version
 appVersion: v0.8.1
```
jgwest: JGW: hmm

install/helm-repo/argocd-agent-principal/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: argocd-agent-principal
 description: Argo CD Agent Principal component for multi-cluster application management
 type: application
-version: 0.3.2
+version: 0.3.3
 appVersion: "v0.8.1"
```
jgwest: `appVersion`

internal/argocd/cluster/cluster.go
```diff
@@ -365,8 +367,15 @@ func readClientCertFromSecret(ctx context.Context, kubeclient kubernetes.Interfa
 	}
 
 	caBytes, ok := secret.Data["ca.crt"]
-	if !ok {
-		return "", "", "", fmt.Errorf("secret %s/%s missing ca.crt", namespace, secretName)
+	if !ok || len(caBytes) == 0 {
```
jgwest: JGW-TODO: Huh `Falls back to principal CA when ca.crt is missing`?

internal/auth/spiffejwt/spiffejwt.go
```diff
@@ -0,0 +1,103 @@
...
+	var svid *jwtsvid.SVID
+	var err error
+
+	audiences := []string{a.Audience}
+
+	if a.BundleSource != nil {
+		svid, err = jwtsvid.ParseAndValidate(token, a.BundleSource, audiences)
+	} else {
+		log.Warn("No SPIFFE JWT bundle source configured, using insecure JWT parsing")
+		svid, err = jwtsvid.ParseInsecure(token, audiences)
```
jgwest: I would expect we should fail and return an error if the signature is invalid, rather than just logging a warning

internal/auth/spiffejwt/spiffejwt.go
```diff
@@ -0,0 +1,103 @@
...
+	if a.AgentIDRegex == nil {
+		return spiffeID, nil
+	}
+
+	matches := a.AgentIDRegex.FindStringSubmatch(spiffeID)
+	if len(matches) < 2 {
+		return "", fmt.Errorf("SPIFFE ID %q did not match agent ID regex %q", spiffeID, a.AgentIDRegex.String())
+	}
+
+	agentID := matches[1]
```
jgwest: JGW: [1]

internal/spire/source.go
```diff
@@ -0,0 +1,125 @@
...
+	"crypto/x509"
+	"fmt"
+
+	"github.com/argoproj-labs/argocd-agent/internal/logging"
+	"github.com/spiffe/go-spiffe/v2/spiffetls/tlsconfig"
+	"github.com/spiffe/go-spiffe/v2/svid/jwtsvid"
+	"github.com/spiffe/go-spiffe/v2/workloadapi"
+)
+
+var log = logging.GetDefaultLogger().ComponentLogger("spire")
```
jgwest: JGW: same as other one?

internal/spire/source.go
```diff
@@ -0,0 +1,125 @@
...
+	clientOpts := workloadapi.WithClientOptions(workloadapi.WithAddr(socketPath))
+
+	x509Source, err := workloadapi.NewX509Source(ctx, clientOpts)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create SPIRE X509Source: %w", err)
+	}
+
+	jwtSource, err := workloadapi.NewJWTSource(ctx, clientOpts)
+	if err != nil {
+		x509Source.Close()
```
jgwest: Can log error from Close

pkg/client/remote.go
```diff
@@ -193,6 +199,23 @@ func WithTLSClientCertFromSecret(kube kubernetes.Interface, namespace, name stri
...
 }
 
+// WithSPIRE configures the Remote to use SPIRE for TLS and authentication.
+// When useMTLS is true, the agent presents its X.509-SVID as a client certificate
+// for mTLS authentication. When false, only server verification is configured
+// and JWT-SVIDs are used for authentication instead.
+func WithSPIRE(source *spire.Source, useMTLS bool) RemoteOption {
+	return func(r *Remote) error {
+		r.spireSource = source
+		spiffeTLS := tlsconfig.TLSClientConfig(source.X509Source(), tlsconfig.AuthorizeAny())
```
jgwest: JGW: AuthorizeAny?

pkg/client/remote.go
```diff
@@ -193,6 +199,23 @@ func WithTLSClientCertFromSecret(kube kubernetes.Interface, namespace, name stri
...
+		r.tlsConfig.InsecureSkipVerify = spiffeTLS.InsecureSkipVerify
```
jgwest: JGW: check this

pkg/client/remote.go
```diff
@@ -651,9 +674,21 @@ func (r *Remote) Connect(ctx context.Context, forceReauth bool) error {
 			}
 			authC := authapi.NewAuthenticationClient(conn)
 
+			creds := r.creds
+			// When SPIRE is configured with JWT authentication,
+			// fetch the JWT-SVID and include it in the credentials.
+			if r.spireSource != nil && r.authMethod == "spiffe-jwt" {
```
jgwest: `"spiffe-Jwt"` to e.g. `MethodSPIFFEJWT`?

principal/server.go
```diff
@@ -1236,6 +1236,29 @@ func (s *Server) loadTLSConfig() (*tls.Config, error) {
 		return nil, nil
 	}
 
+	// When SPIRE is configured, use SPIRE SVID instead of TLS certificate and key
+	if s.options.spireSource != nil {
+		log().Infof("Using SPIRE for server TLS credentials")
+		tlsConfig := &tls.Config{
+			GetCertificate: s.options.spireSource.GetCertificate(),
+			ClientAuth:     tls.NoClientCert,
```
jgwest: JGW: check

principal/server.go
```diff
@@ -1236,6 +1236,29 @@ func (s *Server) loadTLSConfig() (*tls.Config, error) {
...
+			MinVersion:     s.options.tlsMinVersion,
+			MaxVersion:     s.options.tlsMaxVersion,
+			CipherSuites:   s.options.tlsCiphers,
+		}
+		// When mTLS is used with SPIRE, require client certs verified against the SPIRE trust bundle
+		if s.options.requireClientCerts {
```
jgwest: It seems like here there may be some cases where we don't validate clientCerts (e.g. if this var is false).

Does this make sense for the SPIRE case? Would there be a valid use case where a user is using SPIRE, but DOESN'T want to validate the client certs?

test/e2e/fixture/goreman.go
```diff
@@ -59,3 +64,22 @@ func IsProcessRunning(processName string, t *testing.T) bool {
 	// process not found
 	return false
 }
+
+// VerifyProcessLogs checks if expected messages are present in process logs.
+func VerifyProcessLogs(t testing.TB, processName string, requiredMessages []string) {
```
jgwest: `testing.TB?`

internal/auth/spiffejwt/spiffejwt.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	if err != nil {
+		return "", fmt.Errorf("SPIFFE JWT-SVID validation failed: %w", err)
+	}
+
+	spiffeID := svid.ID.String()
+	log.Debugf("Validated SPIFFE JWT-SVID, SPIFFE ID: %s", spiffeID)
+
+	if a.AgentIDRegex == nil {
+		return spiffeID, nil
```
jgwest: If agentIDRegex is nil, is it likely that spiffeID is the correct value we want to return here? Or should we instead return an error?

internal/auth/spiffejwt/spiffejwt.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	matches := a.AgentIDRegex.FindStringSubmatch(spiffeID)
```
jgwest, line 88: The agentIDRegex field is customizable via the user AFAICT, but it seems like we are assuming a particular shape for the regex? How can we be sure the agentID will be at matches[1]?

pkg/client/remote.go
```diff
@@ -193,6 +199,23 @@ func WithTLSClientCertFromSecret(kube kubernetes.Interface, namespace, name stri
...
+// WithSPIRE configures the Remote to use SPIRE for TLS and authentication.
+// When useMTLS is true, the agent presents its X.509-SVID as a client certificate
+// for mTLS authentication. When false, only server verification is configured
+// and JWT-SVIDs are used for authentication instead.
+func WithSPIRE(source *spire.Source, useMTLS bool) RemoteOption {
+	return func(r *Remote) error {
+		r.spireSource = source
+		spiffeTLS := tlsconfig.TLSClientConfig(source.X509Source(), tlsconfig.AuthorizeAny())
+		r.tlsConfig.InsecureSkipVerify = spiffeTLS.InsecureSkipVerify
+		r.tlsConfig.VerifyPeerCertificate = spiffeTLS.VerifyPeerCertificate
```
jgwest, line 211: Is there a scenario where these two would be false? Could we just set them to `true` regardless of spiffeTLS?

pkg/client/remote.go
jgwest, line 209: `AuthorizeAny` is used here, Claude flagged this as:
```
`tlsconfig.TLSClientConfig(source.X509Source(), tlsconfig.AuthorizeAny())` accepts any valid SVID from the trust domain as the server's identity. Any
                           compromised workload in the same SPIRE trust domain can impersonate the principal to an agent.
```

pkg/client/remote.go
```diff
@@ -651,9 +674,21 @@ func (r *Remote) Connect(ctx context.Context, forceReauth bool) error {
 			}
 			authC := authapi.NewAuthenticationClient(conn)
 
+			creds := r.creds
+			// When SPIRE is configured with JWT authentication,
+			// fetch the JWT-SVID and include it in the credentials.
+			if r.spireSource != nil && r.authMethod == "spiffe-jwt" {
+				jwtToken, jwtErr := r.spireSource.FetchJWTSVID(ctx, config.SPIREJWTAudience)
+				if jwtErr != nil {
+					logrus.Warnf("Failed to fetch JWT-SVID: %v (retrying in %v)", jwtErr, cBackoff.Step())
```
jgwest: Maybe `Warnf` -> `Errorf`? It's true we're retrying, but it still will block connection, which sounds error-like to me.

cmd/argocd-agent/principal.go
```diff
@@ -791,6 +852,8 @@ func parseAuth(authStr string) (string, string, error) {
 		return auth.MethodMTLS, p[1], nil
 	case auth.MethodHeader:
 		return auth.MethodHeader, p[1], nil
+	case auth.MethodSPIFFEJWT:
```
jgwest, line 858: Flagged by Claude (adding it here, but it's actually refering to `validateAuthTLSPairing`:
```
     cmd/argocd-agent/principal.go
       ● 912 [security]    `validateAuthTLSPairing()` validates `header` and `mtls` auth against `insecurePlaintext`, but has no case for `spiffe-jwt`. Running `spiffe-jwt` auth with
                           `--insecure-plaintext` sends JWT-SVIDs over an unencrypted channel.
```

test/e2e/fixture/goreman.go
jgwest, line 69: Per claude:
```
     test/e2e/fixture/goreman.go
       ●  78 [correctness] `VerifyProcessLogs` reads a shared log file and checks if `requiredMessages` appear anywhere in it, ignoring the `processName` parameter. If both principal
                           and agent log the same message, the test passes even if only one process actually logged it.
```

We can likely parse the goreman field that indicates which process the log came from.

cmd/argocd-agent/agent.go
```diff
@@ -195,8 +200,40 @@ func NewAgentRunCommand() *cobra.Command {
 			}
 			var remote *client.Remote
 
-			// Configure TLS or plaintext mode
-			if insecurePlaintext {
+			// Validate SPIRE flags
```
jgwest, line 203: Per claude:
```
     cmd/argocd-agent/agent.go
       ● 203 [correctness] When `--spire-agent-socket` and `--insecure-plaintext` (or `--insecure-tls`) are both set, the SPIRE branch silently wins due to if/else ordering. No error or
                           warning is emitted, so the operator believes they have plaintext when they actually have SPIRE TLS.
       ● 215 [correctness] When both `--creds userpass:/path` and `--spire-agent-socket` are set, `WithAuth` is called twice — once for `--creds` and once for SPIRE auth. The SPIRE one
                           silently overwrites the user's explicit credential configuration with no warning.
```

install/helm-repo/argocd-agent-agent/templates/agent-deployment.yaml
```diff
@@ -318,6 +330,11 @@ spec:
             - name: userpass-passwd
               mountPath: /app/config/creds
               readOnly: true
+            {{- if .Values.spire.enabled }}
+            - name: spire-agent-socket
+              mountPath: {{ .Values.spire.hostSocketDir }}
```
jgwest, line 335: Per Claude:
```
     install/helm-repo/argocd-agent-agent/templates/agent-deployment.yaml
       ● 335 [correctness] The SPIRE socket hostPath volume uses `type: DirectoryOrCreate`, which auto-creates an empty directory if SPIRE isn't installed on the node. This masks
                           SPIRE's absence, leading to a confusing runtime error instead of a clear pod scheduling failure.
```

jgwest (changes requested):

internal/auth/spiffejwt/spiffejwt.go
chetan-rns: Should we instead return an error if the regex is not configured?

internal/argocd/cluster/cluster.go
```diff
@@ -365,8 +367,15 @@ func readClientCertFromSecret(ctx context.Context, kubeclient kubernetes.Interfa
...
+		caSecret, err := kubeclient.CoreV1().Secrets(namespace).Get(ctx, config.SecretNamePrincipalCA, metav1.GetOptions{})
```
chetan-rns, line 371: NIT: Should we handle the fallback in a different function? The function `readClientCertFromSecret` is doing more than reading from one secret.

cmd/argocd-agent/agent.go
```diff
@@ -195,8 +200,40 @@ func NewAgentRunCommand() *cobra.Command {
...
+			if spireAgentSocket != "" && spireAuthMethod == "" {
```
chetan-rns, line 204: The spire flag validation logic is duplicated between the agent and the principal. Should we move it to a common helper function?

internal/auth/spiffejwt/spiffejwt.go
jparsai: Actually the `regexp` module returns like this only. Existing mTLS auth also follows same logic https://github.com/argoproj-labs/argocd-agent/blob/main/internal/auth/mtls/mtls.go#L91C1-L100C4

pkg/client/remote.go
jparsai: Here `VerifyPeerCertificate` is a function and `InsecureSkipVerify` is a boolean, built-in objects of crypto. In case of SPIFFE, `InsecureSkipVerify` is always `true` which tells crypto to skip the regular CA verification and use function provided by `VerifyPeerCertificate` instead.

pkg/client/remote.go
jparsai: Yes, initially I tried `AuthorizeID()` but I had to keep `AuthorizeAny()`. `AuthorizeID()` which allows connection for agents having SPIFFE IDs which are configured while principal startup. But the reason to implement SPIFFE feature is to externalizing the agent's identity, where principal doen't need to know anything about agent. It should allow any agent having SPIFFE ID from control-plan's trusted domain to connect with principal.

internal/argocd/cluster/cluster.go
jparsai: Yes that is correct, but fallback is an existing logic, so I didn't modify anything in it.

cmd/argocd-agent/agent.go
jparsai: Yes at the time of your review this was valid, but now after addressing Jonathan's comments, there are few more checks added in both files and now we have only 3 common checks and rest are different in agent and principal, hence I think it wont look clean to validate first 3 condition via common function and then validate rest individually.

jparsai:
Thanks @jgwest @chetan-rns, I addressed your review comments, PTAL

chetan-rns (approved):
Thanks. LGTM!

jgwest (approved):
LGTM, thanks @jparsai!

PR #1059, opened 2026-08-05
fix: preserve notifications state on the principal across agent updates

### LINKED ISSUE #1058: Notifications state on the principal is deleted on every autonomous agent update
**Describe the bug**

Running the Argo CD notifications controller on the principal's control plane (watching the mirrored `Application` resources in the agent namespaces) results in duplicate notifications: every trigger whose condition currently holds is re-sent roughly every time the agent syncs the application's status.

The notifications engine records what it has already delivered in the `notified.notifications.argoproj.io` annotation on the Application. `UpdateAutonomousApp` inherits the agent's annotation set verbatim (preserving only `argocd.argoproj.io/source-uid`), and the agent's copy of the application never carries the notifications annotation — so the delivery state is deleted from the principal's copy on every update, and the controller re-evaluates the trigger as never-notified. The non-patch path of `UpdateStatus` has the same shape for managed agents.

1. Connect an agent in autonomous mode to a principal.
2. Run argocd-notifications on the control plane with any trigger and subscription matching the mirrored application (e.g. `on-health-degraded` with a webhook or Slack service).
3. Put the application into a state where the trigger condition holds (e.g. Degraded).
4. Watch the annotation on the principal's copy: it appears after the notification is sent, then disappears on the next status update from the agent, and the notification is sent again.

**Expected behavior**

The notification is delivered once per trigger transition, as it is with a plain Argo CD installation. Annotations written by controllers on the principal about principal-side behavior should survive updates mirrored from the agent, the same way `argocd.argoproj.io/source-uid` already does.

**Additional context**

Observed with agents on v0.9.0 and reproduced against current main with a unit test (PR to follow): after `UpdateAutonomousApp`, an existing `notified.notifications.argoproj.io` annotation on the principal's copy is gone while `source-uid` survives.

benelser:

When a notifications controller runs on the principal, it records what it has already delivered in the `notified.notifications.argoproj.io` annotation on the principal's copy of the Application. The agent's copy never carries this annotation, so `UpdateAutonomousApp` inheriting the agent's annotation set verbatim deletes the delivery state on every update — the controller then re-sends every notification whose trigger condition holds, on every agent sync. The non-patch path of `UpdateStatus` has the same shape for managed agents.

This PR carries principal-owned annotations (the existing `argocd.argoproj.io/source-uid` mapping, plus the notifications state) from the existing resource over to the incoming one before the incoming annotation set is applied, via a shared helper used by both the update and the jsondiff patch paths. Annotations the agent's side owns keep inheriting the incoming value; only state the agent cannot know about is retained.

Fixes #1058

Unit test included: `Test_ManagerUpdateAutonomous/Retain_principal-owned_annotations_across_an_agent_update` sets up an existing principal-side application carrying `source-uid`, the notifications annotation, and a stale agent-owned annotation, then applies an incoming agent update. Against current main the test fails with the notifications annotation coming back empty (while `source-uid` survives, since it is already special-cased); with this change both principal-owned annotations survive and the agent-owned annotation inherits the incoming value.

Manual reproduction: principal + autonomous agent + argocd-notifications on the control plane with any trigger matching the mirrored app; watch the `notified.notifications.argoproj.io` annotation get written after delivery and wiped by the next status update from the agent.

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Application updates now preserve system-managed source and notification metadata instead of unintentionally overwriting it.
  * Incoming application changes continue to update agent-managed metadata as expected.
  * Standard status updates retain notification delivery information.

* **Improvements**
  * Notification delivery state is tracked consistently across application updates, patches, and status changes.
  * Application metadata remains reliable when changes originate from multiple management processes.

### DIFF
--- a/internal/manager/application/application.go
+++ b/internal/manager/application/application.go
@@ -474,6 +474,23 @@ func (m *ApplicationManager) CompareSourceUID(ctx context.Context, incoming *v1a
 	return result.Exists, result.SourceUIDMatch, nil
 }
 
+// preservePrincipalAnnotations carries principal-owned annotations from the
+// existing application over to the incoming one before the incoming
+// annotation set is applied. State written on the principal's copy of the
+// application (the source UID mapping, the notifications engine's delivery
+// tracking) is unknown to the agent, so an update inherited verbatim from
+// the agent would otherwise delete it.
+func preservePrincipalAnnotations(existing, incoming *v1alpha1.Application) {
+	for _, key := range []string{manager.SourceUIDAnnotation, manager.NotifiedAnnotation} {
+		if v, ok := existing.Annotations[key]; ok {
+			if incoming.Annotations == nil {
+				incoming.Annotations = make(map[string]string)
+			}
+			incoming.Annotations[key] = v
+		}
+	}
+}
+
 // UpdateAutonomousApp updates the Application resource on the control plane side
 // when the agent is in autonomous mode. It will update changes to .spec and
 // .status fields along with syncing labels and annotations.
@@ -497,12 +514,7 @@ func (m *ApplicationManager) UpdateAutonomousApp(ctx context.Context, namespace 
 	}
 
 	updated, err = m.update(ctx, true, incoming, func(existing, incoming *v1alpha1.Application) {
-		if v, ok := existing.Annotations[manager.SourceUIDAnnotation]; ok {
-			if incoming.Annotations == nil {
-				incoming.Annotations = make(map[string]string)
-			}
-			incoming.Annotations[manager.SourceUIDAnnotation] = v
-		}
+		preservePrincipalAnnotations(existing, incoming)
 
 		existing.Annotations = incoming.Annotations
 		existing.Labels = incoming.Labels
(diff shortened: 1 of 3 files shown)


internal/manager/application/application.go
```diff
@@ -474,6 +474,23 @@ func (m *ApplicationManager) CompareSourceUID(ctx context.Context, incoming *v1a
...
 }
 
+// preservePrincipalAnnotations carries principal-owned annotations from the
+// existing application over to the incoming one before the incoming
+// annotation set is applied. State written on the principal's copy of the
+// application (the source UID mapping, the notifications engine's delivery
+// tracking) is unknown to the agent, so an update inherited verbatim from
+// the agent would otherwise delete it.
+func preservePrincipalAnnotations(existing, incoming *v1alpha1.Application) {
+	for _, key := range []string{manager.SourceUIDAnnotation, manager.NotifiedAnnotation} {
```
jannfis: Can we please outsource the list of annotations to a package-local variable? It would make it easier to find, update and test.

jannfis (commented):
Great catch, and thanks for this contribution @benelser.

I just have a minor comment, PTAL.

internal/manager/application/application.go
benelser: Done, the list is a package-local variable now. Thanks for the feedback.

jannfis (approved):
LGTM

PR #1061, opened 2026-08-05
fix: skip mapping an app to an agent that isn't actually connected

### LINKED ISSUE #1060: Changing an app's destination to a non-agent cluster breaks its resource tree in the UI
**Describe the bug**
With destination-based mapping enabled, changing an Application's spec.destination.name to a cluster secret that isn't an actual connected agent (e.g. an in-cluster secret used in a hybrid setup) still gets tracked internally as if it belonged to that "agent". This poisons the Redis proxy's routing for that app, so any later resource-tree or managed-resources lookup gets forwarded to a nonexistent agent instead of falling back to the principal's own Redis.

1. Run the principal with destination-based mapping enabled, with the hub's own Argo CD application controller also active (hybrid mode), and a connected agent.
2. Create an in-cluster secret (no agent mapping) alongside the agent's cluster secret.
3. Create an Application with spec.destination.name set to the connected agent, let it sync, and confirm the resource tree loads fine in the UI.
4. Patch the Application's spec.destination.name to point at the plain, non-agent cluster secret instead.
Open the Application's resource tree / live manifest view in the UI again.

**Expected behavior**
The resource tree should keep loading normally, served from the principal's own Redis, since the app is no longer targeting a real agent.

Instead, the resource tree fails with `error getting cached app resource tree: EOF`, and the principal logs show `no response from agent` / `unexpected nil in get response` for that app's Redis keys. 

**Additional context**
Found this while running a hybrid setup where an ApplicationSet targets both a local-cluster destination and a real spoke cluster by name (OCM-io's argocd-agent integration uses a clusterDecisionResource generator for this, but it reproduces the same way with a plain list generator or hand-created Applications). The local-cluster entry uses a destination name that isn't a connected agent, since it's meant to be reconciled by the hub's own app-controller.

mikeshng:

handleAppAgentChange tracked an app to its new destination agent unconditionally on every destination change, even if that agent was never connected. This poisoned the Redis proxy's routing table and broke resource-tree/manifest reads for that app. Now it checks HasMapping first, same as the rest of this file already does.

Assisted by: Cursor

Fixes #1060 

See #1060 for more details

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved application routing when moving to a new destination agent.
  * Applications moving to agents without a valid cluster mapping are removed without creating invalid routing or resource mappings.
  * Applications moving to properly mapped agents continue to be tracked normally.

* **Tests**
  * Added coverage for mapped and unmapped destination-agent scenarios.

### DIFF
--- a/principal/callbacks.go
+++ b/principal/callbacks.go
@@ -1073,9 +1073,12 @@ func (s *Server) handleAppAgentChange(ctx context.Context, old, new *v1alpha1.Ap
 	}
 
 	if newAgentName != "" {
-		// Add mapping for the new agent
-		s.trackAppToAgent(new, newAgentName)
-		s.resources.Add(newAgentName, resources.NewResourceKeyFromApp(new))
+		// Only map to the new agent if it's actually a connected agent, otherwise we'd
+		// poison the Redis proxy routing for an app that isn't agent-managed.
+		if s.clusterMgr == nil || s.clusterMgr.HasMapping(newAgentName) {
+			s.trackAppToAgent(new, newAgentName)
+			s.resources.Add(newAgentName, resources.NewResourceKeyFromApp(new))
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


mikeshng:
@jgwest @chetan-rns @jparsai could you PTAL? I understand it takes a while for upstream changes to flow down so I would like this fix for OCM-io/ACM sooner in upstream. Thanks!

chetan-rns (approved):
LGTM. Thanks!

PR #1078, opened 2026-08-19
fix: prevent unbounded growth in workqueue Add

### LINKED ISSUE #1053: queue: boundedQueue.Add bound check is not concurrency-safe, queue can grow far beyond maxSize (observed 31k items on a 1000-item queue)
## Summary

`boundedQueue.Add` enforces the queue size limit with an equality check and no synchronization (`internal/queue/queue.go`, present in v0.9.0 and current main):

```go
func (bq *boundedQueue) Add(item *event.Event) {
	// We pop the oldest item if the size is going to exceed maxSize.
	if bq.Len() == bq.maxSize {
		old, _ := bq.Get()
		bq.Done(old)
	}
	bq.TypedRateLimitingInterface.Add(item)
```

The check-then-act is not atomic. If two goroutines call `Add` on the same queue concurrently, interleavings can push `Len()` to `maxSize + 1`. From that point on, `Len() == bq.maxSize` is never true again, so the bound is permanently disabled and the queue grows without limit.

## How we hit it

Our agent Deployments used the default `RollingUpdate` strategy. During a rolling restart, the old and the new agent pod are briefly connected at the same time, using the same agent name. The principal then serves two Subscribe streams for that agent, and both stream handlers call `Add` on the same recv queue concurrently.

## Observed

- Principal and agents on v0.9.0, Argo CD v3.4.4, agent with 150 Applications.
- After a rolling restart of that agent, its recv queue (nominal max 1000, `ARGOCD_AGENT_RECV_QUEUE_SIZE` unset) grew to 31,824 items.
- Principal memory grew from ~0.7Gi to ~1.65Gi while the backlog was held.
- We stopped the growth by scaling the agent Deployment to 0 and letting the queue drain, then switched our agent Deployments to `Recreate` to avoid overlapping pods.

## Suggested fix

- Use `>=` instead of `==`, and make the check-and-evict atomic (a small mutex around evict+add, or `for bq.Len() >= bq.maxSize`).
- Independently, it may be worth rejecting or closing a second Subscribe stream for an agent name that is already connected, since duplicate connections also duplicate event delivery. This would also help the resync flow discussed in #1046.

## Environment

- argocd-agent v0.9.0 (principal and agents)
- Argo CD v3.4.4
- Related context: #1051 (the unbounded backlog makes the saturation described there much worse)

chetan-rns:

Currently, the queue length check and the Get() calls in the bounded queue Add() method are not atomic. So, there is a tiny window where the queue size can grow beyond the max size. Since Add() uses a hard equals `==`, the checks can fail, and the queue can grow without bound. In this PR, we use `>=` so eviction activates again once the queue is at or above the limit, making the bound self-healing

Fixes https://github.com/argoproj-labs/argocd-agent/issues/1053

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Bounded queues now consistently evict the oldest items when adding entries after exceeding their configured capacity.
  * Queue length remains controlled during subsequent additions, even if the capacity was previously overshot.

* **Tests**
  * Added coverage to verify queue behavior after capacity overshoot.

### DIFF
--- a/internal/queue/queue.go
+++ b/internal/queue/queue.go
@@ -68,7 +68,7 @@ func newBoundedQueue(maxSize int, name string) *boundedQueue {
 
 func (bq *boundedQueue) Add(item *event.Event) {
 	// We pop the oldest item if the size is going to exceed maxSize.
-	if bq.Len() == bq.maxSize {
+	if bq.Len() >= bq.maxSize {
 		old, _ := bq.Get()
 		bq.Done(old)
 	}
--- a/internal/queue/queue_test.go
+++ b/internal/queue/queue_test.go
@@ -81,6 +81,31 @@ func Test_Queue(t *testing.T) {
 		assert.Equal(t, "2", front.ID())
 	})
 
+	t.Run("Bound self-heals when the queue exceeds maxSize", func(t *testing.T) {
+		q := NewSendRecvQueues()
+		err := q.Create("agent1")
+		assert.NoError(t, err)
+		bq, ok := q.RecvQ("agent1").(*boundedQueue)
+		assert.True(t, ok)
+
+		const overshoot = 5
+		for i := 0; i < defaultMaxQueueSize+overshoot; i++ {
+			ev := event.New()
+			ev.SetID(strconv.Itoa(i))
+			bq.TypedRateLimitingInterface.Add(&ev)
+		}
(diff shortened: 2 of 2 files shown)


jannfis (approved):
LGTM, thanks!

PR #1081, opened 2026-08-19
chore: Fix the source object on which update actions are diffed, when full detail logging

jgwest:
- At present, when full detail logging is enabled (via e.g. `ARGOCD_AGENT_FULL_DETAIL`), the object patch/update diff is not being logged correctly.
    - This is because the diff (generated via `cmp.Diff`) is being generated from  _the incoming object from peer agent_, rather than from the existing object on the same cluster.
    - This was leading to strange log messages where, for example, the resource uid was changing (impossible) and/or resource version was jumping around wildly (also impossible). (Because we were comparing with the wrong object: the UID and resource version are from the object on the other cluster)
- Fix is just to make a copy of the local cluster object (which is being patched/updated), BEFORE we modify it, and then pass that to log statement.
- This PR only affects logging of update actions; it doesn't make any functional changes.
- (I also removed `CompareSourceUID` which is dead code, while I was at it)

**Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update and operation logging to accurately reflect application state before changes.
  * Ensured operation-state clearing uses the latest available application state before applying updates.
  * Improved application deletion and update handling for more reliable state transitions.
  * Preserved operation details when updating application status.

* **Tests**
  * Expanded coverage for retrieving current application state and preserving operation details during updates.

### DIFF
--- a/internal/manager/application/application_test.go
+++ b/internal/manager/application/application_test.go
@@ -438,6 +438,8 @@ func Test_ManagerUpdateStatus(t *testing.T) {
 		require.Contains(t, updated.Labels, "bar")
 		require.Contains(t, updated.Labels, "some")
 		require.Equal(t, updated.Spec.Source.Path, ".")
+		require.NotNil(t, updated.Operation)
+		require.Equal(t, incoming.Operation, updated.Operation)
 	})
 
 	t.Run("Retain principal-owned annotations across an agent status update", func(t *testing.T) {
--- a/internal/manager/application/application_test.go
+++ b/internal/manager/application/application_test.go
@@ -438,8 +438,6 @@ func Test_ManagerUpdateStatus(t *testing.T) {
 		require.Contains(t, updated.Labels, "bar")
 		require.Contains(t, updated.Labels, "some")
 		require.Equal(t, updated.Spec.Source.Path, ".")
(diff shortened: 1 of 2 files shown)


internal/manager/application/application_test.go
```diff
@@ -438,8 +438,6 @@ func Test_ManagerUpdateStatus(t *testing.T) {
 		require.Contains(t, updated.Labels, "bar")
 		require.Contains(t, updated.Labels, "some")
 		require.Equal(t, updated.Spec.Source.Path, ".")
-		require.NotNil(t, updated.Operation)
-		require.Equal(t, incoming.Operation, updated.Operation)
```
chetan-rns, line 442: Just double-checking. Are these asserts removed intentionally?

chetan-rns (commented):
Added a minor question. Everything else looks good to me.

internal/manager/application/application_test.go
jgwest: Nope, good catch! These are from a sibling branch  (which was  where the code changes in this PR were first written)

I've added them back.

chetan-rns (approved):
LGTM. Thanks!

PR #1095, opened 2026-09-01
chore: add sync-consistency-util to dependabot ignore list

jgwest:
- Dependabot is flagging `hack/sync-consistency-util/go.mod` for vulnerabilities, despite the fact it is only a standalone test utility, and thus not exploitable under any circumstances without our threat model.
- Thus lets add it to `.gitattributes` with `linguist-vendored`, which will, quote:
    - _linguist-vendored is a custom attribute used in .gitattributes files to instruct GitHub's language detection library ([Linguist](https://www.google.com/search?q=https://github.com/github/linguist)) to treat designated files or directories as third-party dependency code rather than primary project code._
    - (and which is apparently likewise used by Dependabot for this purpose)

**Checklist**

## Summary by CodeRabbit

* **Chores**
  * Marked consistency utility files as vendored for repository language and code statistics.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+hack/sync-consistency-util/* linguist-vendored


jgwest:
Nevermind, apparently not possible: 
https://github.com/dependabot/dependabot-core/issues/14408

PR #1101, opened 2026-09-09
fix: process Delete event for app already deleted on the agent

### LINKED ISSUE #1056: Race condition when Application is deleted simultaneously on both principal/managed agent: creates undeletable Application on managed-agent
**Describe the bug**

There is a race condition that can occur when an Application with same name is deleted concurrently on both principal/managed agent.

When this occurs, the symptom is an undeletable Application on managed-agent, where that Application no longer on principal. Any attempts to delete Application on managed-agent will cause it be recreated.

This bug was found with Argo CD Agent eventual consistency checker util, which I will be posting shortly.

#### Race Condition:
0) App 'A' exists on both principal/managed-agent, and is sync.
1) t(ime)=0: User deletes App 'A' on principal
2) t=0: User deletes App 'A' on managed-agent 
	- This is 'bad' user behaviour; they shouldn't do this.
	- But we DO support this case, in the sense that if the user deletes something they shouldn't, we will recreate it.
3) t=1: principal sends 'Application A deleted' event to managed-agent
4) t=2: managed-agent receives 'delete' event from 3), and tries to process it
	- BUT, managed-agent can't process that event: Application 'A' no longer exists on the workload cluster any more, due to 2)
	- Since the Application 'A' no longer exists on agent, agent is not able to figure out source uid for the Application, and thus can't remove the Application from the source cache
	- The 'delete' event from principal is thus ignored.
	- Logic is here: https://github.com/argoproj-labs/argocd-agent/blob/0554ad8a94174c38fe353998bc14c712f4201a58/agent/inbound.go#L862
5) t=3: agent recreates App 'A' on managed-agent, since 'App A' was not removed from the source cache in 4) 

But 5) is incorrect, as the Application was deleted on the source of truth at 1), and thus should not exist on peer.

When this race condition occurs, it creates an Application on managed-agent side that can never be deleted: any attempt to delete it will cause it be recreated. And, since it is out of sync from principal, nothing can remove it from the source cache.

The workaround when this occurs is to restart the agent process; it will re-sync and delete the Application. 

Reproduced from main branch commit `cc598fcd200689689de97e5f8bc3ca0e1c1ad84d` on vcluster dev-env, but likely reproducible anywhere.

KR-Ravindra:

## Problem

When an Application is deleted on the principal and, at about the same time, by a user on the managed agent, the Delete event from the principal reaches the agent after the Application is already gone from the workload cluster. The agent fails to process that event, so it never learns that the deletion was legitimate. The concurrent local deletion is then treated as an unauthorized deletion and the Application is recreated; every further attempt to delete it is reverted the same way. The Application no longer exists on the principal, so nothing can ever remove it again until the agent is restarted.

## Root cause

`agent/inbound.go:875` (`deleteApplication`): the lookup of the existing Application via `a.appManager.Get` returns the NotFound error straight to the caller. Everything that tells the agent the deletion is expected runs only after that lookup succeeds: `a.deletions.MarkExpected(...)` and `a.sourceCache.Application.Delete(...)`. As a result the source UID of the deleted Application is never marked as an expected deletion and the stale source cache entry is kept, and `addAppDeletionToQueue` in `agent/outbound.go` recreates the Application through `RevertUserInitiatedDeletion`.

The NotFound case is already handled when `Delete` itself reports NotFound a few lines further down; only the `Get` path was missing it.

## Fix

The Delete event path now goes through a new `deleteApplicationFromEvent`. It calls `deleteApplication` and, if that returns NotFound from `Get`, marks the deletion as expected for the source UID carried by the incoming event (`sourceUIDForApp(app)`, which is the principal side UID and therefore the value of the `source-uid` annotation on the agent side copy), drops the stale source cache entry in managed mode, unmanages the app, logs at debug level and returns nil. Errors other than NotFound are still returned unchanged.

`deleteApplication` itself keeps returning NotFound from `Get`. It is also called on a source UID mismatch, where the incoming app is the new application; marking its source UID as expected there would let a later user-initiated deletion of the recreated app go unreverted (pointed out in review). That path therefore behaves as on main. Per review, the app is now also unmanaged in both NotFound cases.

Fixes #1056

## How tested

Added the unit test `Delete: App already deleted on the agent must not be recreated` to `Test_ProcessIncomingAppWithUIDMismatch` in `agent/inbound_test.go`. It seeds the source cache with the incoming UID, makes the backend `Get` return NotFound, sends a Delete event through `processIncomingApplication`, and asserts that no error is returned, that no `Delete`/`Create` backend call is made, that the source cache entry is gone and that the deletion is marked as expected.

Added `Create: Old app gone before delete on UID mismatch must not mark the new UID as expected` to the same test: the identity check sees the old app, the second `Get` returns NotFound, and it asserts the error is surfaced, no `Delete`/`Create` is made and neither the new nor the old source UID is marked as an expected deletion.

Before the fix:

```
--- FAIL: Test_ProcessIncomingAppWithUIDMismatch (0.00s)
    --- FAIL: Test_ProcessIncomingAppWithUIDMismatch/Delete:_App_already_deleted_on_the_agent_must_not_be_recreated (0.00s)
        inbound_test.go:401:
            	Error:      	Received unexpected error:
            	            	application.argoproj.io "test" not found
FAIL	github.com/argoproj-labs/argocd-agent/agent	0.059s
```

After the fix:

```
--- PASS: Test_ProcessIncomingAppWithUIDMismatch (0.01s)
    --- PASS: Test_ProcessIncomingAppWithUIDMismatch/Delete:_App_already_deleted_on_the_agent_must_not_be_recreated (0.00s)
ok  	github.com/argoproj-labs/argocd-agent/agent	0.055s
```

`go test ./agent/`, `go vet ./agent/` and `go build ./...` pass.

## Links

- https://github.com/argoproj-labs/argocd-agent/issues/1056

**Checklist**

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of application deletion events when the application has already been removed.
  * Prevented unnecessary delete or recreate operations in this scenario.
  * Removed stale application state and correctly recorded the deletion as expected.
  * Improved handling of create events when an existing application disappears during processing, ensuring the operation reports an error without incorrectly recording a deletion.

### DIFF
--- a/agent/inbound.go
+++ b/agent/inbound.go
@@ -872,10 +872,24 @@ func (a *Agent) deleteApplication(app *v1alpha1.Application) error {
 	}
 
 	// Fetch the source UID of the existing app to mark it as expected deletion.
-	app, err := a.appManager.Get(a.context, app.Name, app.Namespace)
+	existing, err := a.appManager.Get(a.context, app.Name, app.Namespace)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// The app is already gone, e.g. because it was deleted on the agent
+			// at the same time it was deleted on the principal. Still mark the
+			// deletion as expected and drop the source cache entry, otherwise
+			// the agent recreates the app from stale state and the recreated
+			// app can never be deleted again.
+			logCtx.Debug("application is not found, perhaps it is already deleted")
+			a.deletions.MarkExpected(sourceUIDForApp(app))
+			if a.mode == types.AgentModeManaged {
+				a.sourceCache.Application.Delete(sourceUIDForApp(app))
+			}
+			return nil
+		}
 		return err
 	}
+	app = existing
 
 	sourceUID := app.Annotations[manager.SourceUIDAnnotation]
 	a.deletions.MarkExpected(ktypes.UID(sourceUID))
--- a/agent/inbound_test.go
+++ b/agent/inbound_test.go
@@ -379,6 +379,41 @@ func Test_ProcessIncomingAppWithUIDMismatch(t *testing.T) {
 		require.False(t, a.appManager.IsManaged(incomingApp.QualifiedName()))
 	})
 
+	t.Run("Delete: App already deleted on the agent must not be recreated", func(t *testing.T) {
+		configureManager(t, manager.ManagerModeManaged)
+		defer unsetMocks(t)
+		a.appManager.Manage(oldApp.QualifiedName())
+		defer a.appManager.ClearManaged()
+
+		// The app was deleted on the agent at the same time it was deleted on
+		// the principal, so it is gone from the cluster before the Delete event
+		// from the principal is processed.
+		getMock.Unset()
+		notFoundError := kerrors.NewNotFound(schema.GroupResource{
+			Group: "argoproj.io", Resource: "application",
+		}, incomingApp.Name)
+		getMock = be.On("Get", mock.Anything, mock.Anything, mock.Anything).Return(nil, notFoundError)
+
+		a.sourceCache.Application.Set(incomingApp.UID, incomingApp.Spec)
+
+		ev := event.New(evs.ApplicationEvent(event.Delete, incomingApp), targets.Application)
+		err := a.processIncomingApplication(ev)
+		require.NoError(t, err)
+
+		// Identity check and lookup of the existing app, but no Delete or Create
+		expectedCalls := []string{"Get", "Get"}
+		gotCalls := []string{}
+		for _, call := range be.Calls {
+			gotCalls = append(gotCalls, call.Method)
+		}
+		require.Equal(t, expectedCalls, gotCalls)
+
+		// The stale source cache entry must be dropped and the deletion marked
+		// as expected, so the agent does not recreate the app.
+		require.False(t, a.sourceCache.Application.Contains(incomingApp.UID))
+		require.True(t, a.deletions.RemoveExpected(incomingApp.UID))
+	})
+
 	t.Run("SpecUpdate: upsert policy skips delete and updates in-place", func(t *testing.T) {
 		configureManager(t, manager.ManagerModeManaged)
 		defer unsetMocks(t)


KR-Ravindra:
Round 1 self-review.

1. Checked against `agent/inbound.go` on main: `deleteApplication` returns the `Get` error (L875-878) before `MarkExpected` and the source-cache delete run, while a NotFound from `Delete` (L886-891) is already tolerated. The new branch mirrors that existing path, keyed by the incoming event's source UID, which is the value the agent-side copy carries in its `source-uid` annotation.
2. `go test ./agent/` and `go vet ./agent/` pass on the branch. With main's `inbound.go` swapped in, `Test_ProcessIncomingAppWithUIDMismatch/Delete:_App_already_deleted_on_the_agent_must_not_be_recreated` fails with `application.argoproj.io "test" not found`, matching the description.
3. No other PR references #1056. DCO and title lint pass; the ci/CodeQL/image workflows are waiting for a maintainer to approve the run.

Ready for review.

agent/inbound.go
```diff
@@ -872,10 +872,24 @@ func (a *Agent) deleteApplication(app *v1alpha1.Application) error {
...
+			// deletion as expected and drop the source cache entry, otherwise
+			// the agent recreates the app from stale state and the recreated
+			// app can never be deleted again.
+			logCtx.Debug("application is not found, perhaps it is already deleted")
+			a.deletions.MarkExpected(sourceUIDForApp(app))
+			if a.mode == types.AgentModeManaged {
+				a.sourceCache.Application.Delete(sourceUIDForApp(app))
+			}
+			return nil
+		}
```
jgwest, line 892: Can you also add `a.appManager.Unmanage()` into this code block as well? The returned `err` can simply be logged.

Can you also do me a favour and add `a.appManager.Unmanage()` into the `logCtx.Debug("application is not found, perhaps it is already deleted")` block below?

Since:
- A) `a.appManager.IsManaged(app.QualifiedName())` is necessarily true at this point in the code
- B) The Application necessarily doesn't exist anymore.
Then, I don't see why it should be managed in either case.

KR-Ravindra:
@jgwest @jannfis gentle ping on this one when you have a moment: fix for #1056, `deleteApplication` returned on the `Get` NotFound before marking the deletion expected and dropping the source-cache entry, so an app deleted on the agent at the same time as on the principal was recreated from stale state; the new branch mirrors the existing NotFound-from-`Delete` path, with a unit test. CI is green.

jgwest (changes requested):

agent/inbound.go
KR-Ravindra: @jgwest done in 7bd60e0: `Unmanage()` is now called in both NotFound blocks in `agent/inbound.go` (after `Get` and after `Delete`), with the error logged at warn level, and the new test asserts the app is no longer managed afterwards. Could you take another look?

agent/inbound.go
```diff
@@ -872,10 +872,27 @@ func (a *Agent) deleteApplication(app *v1alpha1.Application) error {
...
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// The app is already gone, e.g. because it was deleted on the agent
+			// at the same time it was deleted on the principal. Still mark the
+			// deletion as expected and drop the source cache entry, otherwise
+			// the agent recreates the app from stale state and the recreated
+			// app can never be deleted again.
+			logCtx.Debug("application is not found, perhaps it is already deleted")
+			a.deletions.MarkExpected(sourceUIDForApp(app))
+			if a.mode == types.AgentModeManaged {
```
chetan-rns: There's one scenario where this can allow unauthorized deletion. The function `deleteApplication` is also invoked when there is a source UID mismatch between the incoming app and the existing app. The agent will then delete the existing app using [deleteApplication](https://github.com/argoproj-labs/argocd-agent/blob/8e916bf75504e15f8e5ce850587773776281c9da/agent/inbound.go#L324) and create the incoming app. 

 We are marking the deletion as expected using the source UID of the new app, thereby marking the deletion for the new app as legitimate instead of the old app. This allows the user to delete the app, and the agent will not recreate it because it was marked as expected.

KR-Ravindra: @chetan-rns you're right, thanks. In the mismatch path `app` is the incoming (new) application, so marking its source UID would have let a later user deletion of the recreated app go unreverted.

Done in 0c9e69b: `deleteApplication` returns NotFound from `Get` unchanged again, so the mismatch path behaves as on main (error, retry). The Delete event path now goes through `deleteApplicationFromEvent`, which handles NotFound by marking the event's source UID as expected, dropping the source cache entry and unmanaging. New test covers the mismatch case (`Get` NotFound after the identity check) and asserts the error is surfaced and the new UID is not marked expected. Could you take another look?

This comment was drafted with AI assistance and checked against the code.

jgwest (approved):

jgwest:
Thanks @KR-Ravindra and thanks @chetan-rns!

PR #1108, opened 2026-09-11
fix(redisproxy): send RESP null bulk string on cache miss

### LINKED ISSUE #1107: Redis proxy sends RESP simple error instead of null bulk string on cache miss, breaking Argo CD cache-miss recovery
## Summary

On a Redis cache miss, the principal's resource proxy replies to Argo CD with `"-\r\n"`, intending a "redis null". In RESP, `-` is the **simple error** prefix, not a null — so this is an *error reply with an empty message*, not a cache miss.

The result is that `go-redis` hands Argo CD a non-nil error that is neither `redis.Nil` nor `ErrCacheMiss`, so Argo CD's cache-miss recovery path never runs and the UI surfaces an error with an empty cause:

```
level=error msg="finished call" grpc.code=Unknown grpc.component=server \
  grpc.error="error getting cached app managed resources: " \
  grpc.method=ManagedResources grpc.service=application.ApplicationService
```

Note the empty string after the colon — that is the empty RESP error message being propagated verbatim.

## Location

https://github.com/argoproj-labs/argocd-agent/blob/main/principal/redisproxy/redisproxy.go#L559

```go
} else if !response.Get.CacheHit {

    // Cache miss: no data for that key in agent redis

    response := "-\r\n" // Cache is miss is a simple redis null
```

Present in `v0.9.0` (line 537) and still on `main` (line 559).

Per the [RESP spec](https://redis.io/docs/latest/develop/reference/protocol-spec/):

- `-` — simple error, format `-Error message\r\n`
- `$-1\r\n` — null bulk string (RESP2), the correct reply for `GET` on a missing key
- `_\r\n` — null (RESP3 only)

## Reproduction

A minimal fake Redis server that returns each payload for a `GET`, queried with `github.com/redis/go-redis/v9`:

```
argocd-agent cache miss      payload="-\r\n"    err=true  Error()=""           isRedisNil=false
correct RESP2 null bulk      payload="$-1\r\n"  err=true  Error()="redis: nil"  isRedisNil=true
```

Full reproducer:

<details>
<summary>main.go</summary>

```go
package main

import (
	"bufio"
	"context"
	"errors"
	"fmt"
	"net"
	"strings"

	"github.com/redis/go-redis/v9"
)

// fakeRedis replies to any command with the given raw RESP payload.
func fakeRedis(payload string) net.Listener {
	l, _ := net.Listen("tcp", "127.0.0.1:0")
	go func() {
		for {
			c, err := l.Accept()
			if err != nil {
				return
			}
			go func(c net.Conn) {
				defer c.Close()
				r := bufio.NewReader(c)
				for {
					line, err := r.ReadString('\n')
					if err != nil {
						return
					}
					if !strings.HasPrefix(line, "*") {
						continue
					}
					n := 0
					fmt.Sscanf(line, "*%d", &n)
					for i := 0; i < n*2; i++ {
						if _, err := r.ReadString('\n'); err != nil {
							return
						}
					}
					c.Write([]byte(payload))
				}
			}(c)
		}
	}()
	return l
}

func probe(name, payload string) {
	l := fakeRedis(payload)
	defer l.Close()
	c := redis.NewClient(&redis.Options{Addr: l.Addr().String()})
	_, err := c.Get(context.Background(), "app|managed-resources|alloy|1.8.3.gz").Result()
	txt := "<nil>"
	if err != nil {
		txt = err.Error()
	}
	fmt.Printf("%-28s payload=%-10q err=%-6v Error()=%q isRedisNil=%v\n",
		name, payload, err != nil, txt, errors.Is(err, redis.Nil))
}

func main() {
	probe("argocd-agent cache miss", "-\r\n")   // redisproxy.go:559
	probe("correct RESP2 null bulk", "$-1\r\n") // expected
}
```

</details>

## Why this breaks the UI

Argo CD's `getCachedAppState` ([server/application/application.go](https://github.com/argoproj/argo-cd/blob/v3.4.1/server/application/application.go#L1404-L1425)) recovers from a cache miss by triggering a refresh and retrying:

```go
err := getFromCache()
if err != nil && errors.Is(err, servercache.ErrCacheMiss) {
    // ... trigger app refresh, then retry getFromCache()
}
return err
```

`ErrCacheMiss` is derived from `redis.Nil`. Because `"-\r\n"` produces neither, the branch is skipped and the empty error is returned to the client. So instead of transparently repopulating the cache, the UI shows a failure with no message.

Observed on the **Logs** and **Diff / Live Manifest** tabs for an app on an agent-managed cluster. It is intermittent: it only occurs in the window where a given key (e.g. `app|managed-resources|<app>`) has expired on the agent while related keys are still cached. Reloading usually succeeds, which is consistent with the key being repopulated by other means in between.

Agent-side trace confirms the proxy path itself is healthy — cache *hits* are served correctly:

```
Redis GET 'app|managed-resources|alloy|1.8.3.gz'  -> 8355 bytes / CacheHit=true
Redis GET 'app|resources-tree|alloy|1.8.3.gz'     -> 2155 bytes / CacheHit=true
```

Only the miss path is affected.

## Suggested fix

```diff
-		response := "-\r\n" // Cache is miss is a simple redis null
+		response := "$-1\r\n" // Cache miss is a RESP2 null bulk string
```

This makes `go-redis` return `redis.Nil`, which Argo CD maps to `ErrCacheMiss`, restoring the refresh-and-retry behaviour so the tab self-heals rather than erroring.

Happy to open a PR with this change if useful.

## Environment

- argocd-agent `v0.9.0` (also verified present on `main`)
- Argo CD `v3.4.1`
- Principal and agent on separate GKE clusters, agent in `autonomous` mode, resource proxy enabled

lucasmellos:
Fixes #1107

## What

On a Redis cache miss, the principal's resource proxy replied to Argo CD with `"-\r\n"`, intended as a "redis null". In RESP, `-` is the **simple error** prefix (`-Error message\r\n`), so this is an *error reply with an empty message*, not a null. The RESP2 null bulk string is `$-1\r\n`.

## Why it matters

Redis clients surface `"-\r\n"` as a non-nil error whose message is the empty string — neither `redis.Nil` nor Argo CD's `ErrCacheMiss`. Argo CD's `getCachedAppState` only recovers (trigger refresh, retry) when the error `errors.Is` an `ErrCacheMiss`, so the miss is never recognised and the error is returned to the client verbatim:

```
level=error msg="finished call" grpc.code=Unknown grpc.component=server \
  grpc.error="error getting cached app managed resources: " \
  grpc.method=ManagedResources grpc.service=application.ApplicationService
```

The empty string after the colon is the empty RESP error message. In the UI this shows up as the **Logs** and **Diff / Live Manifest** tabs failing for apps on agent-managed clusters, intermittently — only in the window where that key has expired on the agent — with no error message to go on.

Verified against `go-redis/v9` with a fake Redis server returning each payload for a `GET`:

```
argocd-agent cache miss      payload="-\r\n"    err=true  Error()=""            isRedisNil=false
correct RESP2 null bulk      payload="$-1\r\n"  err=true  Error()="redis: nil"   isRedisNil=true
```

The full reproducer is in #1107.

## Change

```diff
-		response := "-\r\n" // Cache miss is a simple redis null
+		// A cache miss must be reported as a RESP null bulk string. The '-' prefix
+		// is the RESP simple error type, which Redis clients surface as an error
+		// (with an empty message) rather than as a missing key, preventing Argo CD
+		// from recognizing the miss and refreshing the cache.
+		response := "$-1\r\n"
```

The existing test `cache miss returns 'CacheHit: false', with empty error string` asserted the old byte sequence, so its expectation is updated to `$-1\r\n` in the same commit.

## Testing

- `go build ./principal/...` — clean
- `go test ./principal/redisproxy/...` — `ok github.com/argoproj-labs/argocd-agent/principal/redisproxy`
- `gofmt -l principal/redisproxy/` — clean

Only the miss path is affected; cache hits already wrote a correct bulk string (`$<len>\r\n<bytes>\r\n`) and are untouched.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected cache-miss handling so Redis clients and Argo CD recognize missing keys properly instead of treating them as errors.

### DIFF
--- a/principal/redisproxy/redisproxy.go
+++ b/principal/redisproxy/redisproxy.go
@@ -556,7 +556,11 @@ func (rp *RedisProxy) handleAgentGet(connState *connectionState, key string, arg
 
 		// Cache miss: no data for that key in agent redis
 
-		response := "-\r\n" // Cache miss is a simple redis null
+		// A cache miss must be reported as a RESP null bulk string. The '-' prefix
+		// is the RESP simple error type, which Redis clients surface as an error
+		// (with an empty message) rather than as a missing key, preventing Argo CD
+		// from recognizing the miss and refreshing the cache.
+		response := "$-1\r\n"
 
 		logCtx.Tracef("Wrote GET response to argo cd: '%s'", response)
 
--- a/principal/redisproxy/redisproxy_test.go
+++ b/principal/redisproxy/redisproxy_test.go
@@ -214,6 +214,9 @@ func (m *mockSendSynchronousMessageToAgentFunc) call(agentName string, connectio
 	return m.responseToReturn
 }
 
+// Test_handleAgentGet verifies that handleAgentGet translates each agent
+// response into the correct RESP reply to Argo CD: a bulk string on a cache
+// hit, a null bulk string on a cache miss, and an error on a failed get.
 func Test_handleAgentGet(t *testing.T) {
 	t.Run("successful get with cache hit", func(t *testing.T) {
 		// Setup mocks
@@ -352,7 +355,8 @@ func Test_handleAgentGet(t *testing.T) {
 		require.NoError(t, err)
 		require.Equal(t, 1, mockSendFunc.callCount)
 		require.True(t, mockWriter.writeToArgoCDRedisSocketCalled)
-		expectedResponse := "-\r\n"
+		// RESP null bulk string: signals a missing key, rather than an error
+		expectedResponse := "$-1\r\n"
 		require.Equal(t, expectedResponse, string(mockWriter.writtenBytes))
 	})
 


jannfis:
A RESP null value is actually `_\r\n`, not `-\r\n` (refer to https://redis.io/docs/latest/develop/reference/protocol-spec/#nulls).

Can you test whether returning a true null value instead of a null bulk string works properly, too?

PR #1111, opened 2026-09-16
feat: toggle debug settings during runtime

chetan-rns:

Currently, the agent/principal only supports setting the log level and pprof at startup. In order to configure a debug setting, the principal will have to restart and reconnect to all the agent clusters. This PR provides a way to configure the debug settings during runtime. 

- Both the agent & principal will watch and reconcile the existing argocd-agent-params configmap.
- Users can dynamically apply log level, full detail, and enable/disable pprof from this params configmap.
- The ConfigMap is the source of truth. The deployment will enable the environment variables from the ConfigMap fields, and the components will reconcile settings from the same configmap
- Settings will fall back to values configured during startup if the configmap keys are removed.

Assisted-by: Cursor

Fixes #?

**Checklist**

## Summary by CodeRabbit

- **New Features**
  - Added runtime configuration for log levels, detailed logging, and profiling through the parameters ConfigMap.
  - Added configurable profiling controls for agent and principal services, including localhost listeners and runtime enablement.
  - Added Helm and Kubernetes configuration options for these settings.

- **Bug Fixes**
  - Health endpoints now use isolated routing, preventing conflicts with unrelated HTTP handlers.
  - Invalid logging configuration is rejected without partially applying changes.

- **Documentation**
  - Updated configuration, profiling, and deployment guidance to describe runtime settings, defaults, and access behavior.


cmd/argocd-agent/agent.go
```diff
@@ -486,8 +504,11 @@ func NewAgentRunCommand() *cobra.Command {
 		env.BoolWithDefault("ARGOCD_AGENT_ENABLE_COMPRESSION", false),
 		"Use compression while sending data between Principal and Agent using gRPC")
 	command.Flags().IntVar(&pprofPort, "pprof-port",
-		env.NumWithDefault("ARGOCD_AGENT_PPROF_PORT", cmdutil.ValidPort, 0),
-		"Port the pprof server will listen on")
+		env.NumWithDefault("ARGOCD_AGENT_PPROF_PORT", cmdutil.ValidPort, 6161),
+		"Port the pprof server will listen on, bound to localhost. Set to 0 to not listen at all.")
+	command.Flags().BoolVar(&pprofEnabled, "pprof-enabled",
+		env.BoolWithDefault("ARGOCD_AGENT_PPROF_ENABLED", false),
+		"Enable or disable profiling for the agent")
```
chetan-rns, line 511: The alternative approach is to bind pprof to the metrics server, similar to what Argo CD does. https://github.com/argoproj/argo-cd/blob/c48f8f8ac84310c2ceded2b3ccb3e4b184d53bb9/cmd/argocd-repo-server/commands/argocd_repo_server.go#L211

There will be no `--pprof-port` flag. `--pprof-enable` will be the only way to enable/disable the profiling.

cmd/argocd-agent/principal.go
```diff
@@ -557,6 +570,9 @@ func NewPrincipalRunCommand() *cobra.Command {
 	command.Flags().StringSliceVar(&fullDetailCategories, "full-detail",
 		env.StringSliceWithDefault("ARGOCD_PRINCIPAL_FULL_DETAIL", nil, nil),
 		"Enable full detail logging for specified categories. Comma-separated list of: "+cmdutil.AvailableFullDetailCategories)
+	command.Flags().StringVar(&paramsConfigMap, "params-configmap",
+		env.StringWithDefault("ARGOCD_PRINCIPAL_PARAMS_CONFIGMAP", nil, settings.DefaultConfigMapName),
```
jparsai: One question. Is this CM created by default? (atleast I dont see it in my local setup)
If not created by default, then should it be created?, so users dont have to do manual effort?

PR #1117, opened 2026-09-21
fix(tls): omit certificate signing usage from leaf certificates

### LINKED ISSUE #1116: Generated leaf certificates include keyCertSign and fail strict X.509 validation
**Describe the bug**

`argocd-agentctl` generates server and client leaf certificates with `CA:FALSE` in Basic Constraints but `keyCertSign` in Key Usage. Strict X.509 verification rejects these certificates, even when the issuer is trusted, the hostname matches, and the certificate is within its validity period.

Observed with OpenSSL 3.6.4 and v0.10.0 (`78a3788f5aa757707c00e563198f3d4c676fe04b`). The same assignments are present in current `main` (`8e916bf75504e15f8e5ce850587773776281c9da`), in `GenerateServerCertificate` and `GenerateClientCertificate` in `internal/tlsutil/generate.go`.

Use a disposable Kubernetes context and a fresh namespace for this example. These commands create test PKI, so do not use an existing principal namespace.

```sh
TEST_CONTEXT=your-disposable-test-context
TEST_NAMESPACE=argocd-pki-repro

kubectl --context "$TEST_CONTEXT" create namespace "$TEST_NAMESPACE"
argocd-agentctl --principal-context "$TEST_CONTEXT" \
  --principal-namespace "$TEST_NAMESPACE" pki init
argocd-agentctl --principal-context "$TEST_CONTEXT" \
  --principal-namespace "$TEST_NAMESPACE" pki issue principal \
  --dns principal.example.test

# Export only public certificates, not private keys.
kubectl --context "$TEST_CONTEXT" -n "$TEST_NAMESPACE" \
  get secret argocd-agent-ca -o jsonpath='{.data.tls\.crt}' \
  | base64 --decode > ca.pem
kubectl --context "$TEST_CONTEXT" -n "$TEST_NAMESPACE" \
  get secret argocd-agent-principal-tls -o jsonpath='{.data.tls\.crt}' \
  | base64 --decode > principal.pem

openssl x509 -in principal.pem -noout -text
openssl verify -x509_strict -purpose sslserver \
  -verify_hostname principal.example.test -CAfile ca.pem principal.pem
```

The certificate has:

```text
X509v3 Basic Constraints: critical
    CA:FALSE
X509v3 Key Usage: critical
    Digital Signature, Certificate Sign
```

Strict OpenSSL verification fails with:

```text
error 82 at 0 depth lookup: Key usage keyCertSign invalid for non-CA cert
```

The same error was encountered in a Python 3.13 TLS connection using `ssl.create_default_context(cadata=...)` and hostname verification. It is also reproducible without Kubernetes by generating a CA and server/client leaf with the `internal/tlsutil` functions and verifying the resulting public certificates with OpenSSL.

**Expected behavior**

Generated leaf certificates should pass strict verification against the generated CA. Keep `digitalSignature` and the appropriate server/client extended key usage on leaves, and retain `keyCertSign` only on the CA.

[RFC 5280 section 4.2.1.3](https://www.rfc-editor.org/rfc/rfc5280.html#section-4.2.1.3) requires the CA Basic Constraints bit when `keyCertSign` is set.

**Additional context**

Both leaf generators currently set `x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign` despite setting `IsCA: false`. Removing `KeyUsageCertSign` from those two templates makes strict verification pass for both server and client certificates; CA generation remains unchanged. I have a small fix and regression coverage to contribute.

This report concerns certificate-profile interoperability. Existing certificates need to be reissued after the generator is fixed; changing the CLI alone does not update stored certificates.

luiseterc:

Generated server and client leaf certificates currently combine `IsCA: false` with `KeyUsageCertSign`. Strict X.509 verification rejects them with `Key usage keyCertSign invalid for non-CA cert`.

Remove the certificate-signing bit from both leaf templates while preserving digital-signature usage, their existing EKUs and SANs, and CA certificate-signing usage. Add unit assertions for the CA/leaf distinction and a self-contained end-to-end test that verifies generated server and client certificates with strict OpenSSL checks (including the server hostname).

Fixes #1116

The new unit assertions and strict OpenSSL test both fail on the unmodified generators and pass with the fix. The strict verification suite needs OpenSSL but no Kubernetes cluster or existing credentials.

Validation (all passed locally):

- `go test -race ./internal/tlsutil -count=1`
- `go test -race ./test/e2e -run '^Test_VerifyGeneratedCertificatesWithStrictX509$' -count=1 -v`
- `make test` (full race-enabled unit suite)
- `make lint`
- `git diff --check`

The full vcluster-based end-to-end suites were not run locally; they remain for upstream CI. Existing stored certificates are not changed by this patch and need reissuance to gain the corrected key usage.

**Checklist**

* [x] Documentation update is required by this PR (and has been updated) OR no documentation update is required. This fixes the existing certificate profile without changing the CLI or configuration contract.


jannfis (approved):
Nice! Thanks for this fix, @luiseterc.

LGTM.

